Issue #100, opened 2014-03-26
Add multiline highlight 


pwntester:
It would be great to have multiline highlight by wrapping all the lines related with a single error in a single highlight that expands to several lines like in this picture:

http://www.sublimerge.com/Images/sublimerge-pro.png

New capture groups should be added to the linter regex to indicate the linenumber (already exists) and number of lines (default to 1)

PR #101, opened 2014-03-27
new regex attribute length

pwntester:
This pull request implements "Add multiline highlight (#100)" by basically adding a new attribute to the linter regex called length. If length is not None, then its passed to the highlight.range function (unmodified) that will higlight the a "length" number of characters from the beginning of the line and if this value is grater than the line length, it will continue marking characters in the following line, effectively creating a highlight block similar to the one shown in the issue #100


braver:
@pwntester it's been a while, but are you looking to pick this up again or should we just close it?

pwntester:
This can now be closed, thanks!

PR #104, opened 2014-04-15
[on-hold] Multi color

abathur:
Implements flexible mark types and mark-type color settings.

Global color settings may be defined under 'colors'; linter-specific colors may be defined in a 'colors' setting under a given linter. Internally the main driver is that marks are now scoped by linter (i.e., 'annotations.warning' or 'pylint.error') and then the complexity is reduced from left to right in search of the nearest reasonable setting. 

Previously the rules for mark priority were pretty simple. I've also introduced a 'precedence' setting which is a list of scopes from most to least important. If the precedence setting is insufficient for resolving a conflict, precedence is granted to the first mark made.

This changeset was larger and touched a lot more of the code than I was expecting; I don't imagine this will go through clean on the first try, so I'll be looking forward to your thoughts.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -6,7 +6,6 @@
         },
         "debug": false,
         "delay": 0.25,
-        "error_color": "D02000",
         "gutter_theme": "Packages/SublimeLinter/gutter-themes/Default/Default.gutter-theme",
         "gutter_theme_excludes": [],
         "lint_mode": "background",
@@ -17,6 +16,10 @@
             "osx": [],
             "windows": []
         },
+        "precedence": [
+            "error",
+            "warning"
+            ],
         "python_paths": {
             "linux": [],
             "osx": [],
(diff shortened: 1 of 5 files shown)


Pugio:
What's the state of this PR? Is it just waiting on the SublimeText bug?

abathur:
@Pugio Mostly. It contains a kludgy workaround for the ST3 bug which involves writing a few copies of the gutter icon so each setting can use its own. If the bug is fixed, the workaround should be removed before it is ready to go. 

I also noticed a bug of some sort when switching between color schemes for the testing I did on the other report. I'm not sure if my running copy at the moment has the code in this PR active and I don't see an open issue that appears to match up; when there is interest in advancing the PR either with or without the icon kludge, I will pull any SL updates and do some testing on that point to make sure this color-switching issue is at least not being introduced here.

LeonardoGentile:
@abathur would this solve or going in the direction of #279 ?

braver:
FYI we're now re-approaching this, a new PR is upcoming.

PR #155, opened 2014-11-14
NodeLinter subclass

andrewdeandrade:
This is the NodeLinter subclass that linter executables installed locally in `node_modules/.bin/` can subclass from.

This is my first time writing python code. Advice and recommendation of things to change welcome.

This would close https://github.com/SublimeLinter/SublimeLinter3/issues/144

### DIFF
--- a/lint/node_linter.py
+++ b/lint/node_linter.py
@@ -39,10 +39,11 @@ class NodeLinter(linter.Linter):
 
     def context_sensitive_executable_path(self, cmd):
         """
-        Attempt to locate the npm module specified in cmd, searching
-        the local node_modules/.bin folder first before looking in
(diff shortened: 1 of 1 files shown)


tschaub:
@malandrew @skj3gg - I'm unable to get a `NodeLinter` based linter working without overriding `can_lint`.  As far as I can tell, the logic in `Linter.can_lint` won't work if the linter executable is not available on the path.  I opened #179 with the changes that were required for me to get `SublimeLinter-contrib-eslint` working as a `NodeLinter` (see https://github.com/tschaub/SublimeLinter-eslint/compare/node-linter).

I'm curious if anybody else has gotten a `NodeLinter` based plugin working without the executable being installed globally.

Issue #117, opened 2014-06-07
Linter could run in a separate thread


staltz:
I'm creating a linter and also a SublimeLinter3 (v3.4.2) plugin for it, for the "Eclipse Xtend" language. It's still in alpha, and my linter takes some time (~5 sec) to process even a simple single file. When the plugin runs, Sublime Text hangs on the UI thread (I see the rainbow spinner in OSX), running my linter, and I cannot interact with the editor. Lint mode is `background` with delay 0.25.

It would be important to have the linter running in a thread parallel to the UI thread. I couldn't discover from the docs if this is already possible through some configuration.

PR #176, opened 2014-12-30
Updated util.py to support MinGW64 environment

vladipus:
Updated the util.py's search methods to support the MinGW64 environment under MS Windows.


braver:
This PR is outdated, but the original issue is still open. We'll keep tracking of MinGW64 support in #117, but I'm not keeping all the pull requests around till we have an idea of if/when/how we're going to approach it.

Issue #157, opened 2014-11-24
Don't disable linter if not installed globally


andrewdeandrade:
NodeLinter supports running local binaries installed by npm. Unfortunately, if the binary isn't also installed globally, then SublimeLinter will disable it entirely and not try looking for it again until SublimeText is restarted.

tschaub:
@malandrew see #179 for a proposed solution to this.  Thanks for getting `NodeLinter` in.  I'm happy to help get it working.

braver:
Since #179 is merged I think we can close this. If I'm wrong, let me know and we'll bring it back.

PR #179, opened 2015-01-20
Override can_lint for NodeLinter

tschaub:
Without this change, if the linter executable is not found on the path (with [`util.which(cmd)`](https://github.com/SublimeLinter/SublimeLinter3/blob/master/lint/linter.py#L1443-L1457)), the NodeLinter will be disabled.  Since `can_lint` is called with the view syntax only (and not the view), and the executable used will vary based on the view filename, the `NodeLinter`'s implementation is optimistic, determining its ability to lint based on syntax alone.

It seems to me like the only alternative would be to call `can_lint` with the `view` as well - but this would diminish the effect of caching the result.

Without `eslint` installed globally, I was unable to get the [SublimeLinter-contrib-eslint](https://github.com/roadhump/SublimeLinter-eslint) plugin working after inheriting from `NodeLinter`.  With the changes in this branch, the `eslint` plugin works with a local install (and [this change](https://github.com/tschaub/SublimeLinter-eslint/compare/node-linter) to inherit from `NodeLinter`).

### DIFF
--- a/lint/node_linter.py
+++ b/lint/node_linter.py
@@ -13,6 +13,7 @@
 import json
 import hashlib
 
+from functools import lru_cache
 from os import path, access, X_OK
 from . import linter, persist, util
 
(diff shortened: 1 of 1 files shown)


skj3gg:
I'll test this out tonight. What OS/shells has this been tried with. Thanks!

Also discussed in #157

tschaub:
Tested on OSX with bash (though this should have no impact).  As mentioned above, this is an "optimistic" way to determine if the linter can lint the given syntax - based on the linter's configured syntax alone.  This makes no change to the way the executable binary is looked up when `lint` is actually run.

Let me know if you have any suggestions on alternative ways to handle this.  I'm happy to rework this branch.

tschaub:
It's worth adding that to test this, I checked out https://github.com/tschaub/SublimeLinter3/commit/5dcefe93c7b4d7fa5b4881974f77ddf209fc7af3 and https://github.com/tschaub/SublimeLinter-eslint/tree/node-linter as packages.  I do not have `eslint` on my path, but I have it installed locally for several projects.  When opening files in a project with `eslint` installed locally, I confirmed that the linter was running as expected.

malandrew:
I will test this out tomorrow.

tschaub:
@malandrew or @skj3gg - any chance you've taken a look at this?  I'd be happy to talk over or work on alternatives.  I'm looking forward to using locally installed node linters.

skj3gg:
Thanks, sorry, had to get a Windows VM and going.

tschaub:
Thanks for merging @skj3gg.  I look forward to the next release.

skj3gg:
It actually should be out in the wild currently, so you should be able to use it.

PR #190, opened 2015-02-08
Fix an infinite loop in windows.

m-hall:
Since the root level directory in windows is not “/“, this function
would never resolve if if could not find a manifest file (package.json).


skj3gg:
Thanks

PR #221, opened 2015-03-21
demote_to_warning_match setting

brocoli:
Added the new setting "demote_to_warning_match" to each linter's preferences. It works almost exactly the same as the "ignore_match" preference, but instead of ignoring the errors it catches, it only demotes them to warnings instead.

I have been using this together with silent_warnings and lint_mode on save for some time, specifically with luacheck, demoting the "unused argument" error to a warning.


skj3gg:
Sorry this took me a month! Released in v4.4 soon.

PR #225, opened 2015-03-25
Removed invisible characters in "template" word

degorov:
Trust me or not, but there are some hidden characters in "template" word between "p" and "l" :)  I can't even see them in Sublime, but they are really there and script type='text/template' doesn't work as it should. It works only if you copy "template" word from SublimeLinter\fixed-syntaxes\HTML\HTML.tmLanguage directly. They can be seen in GitHub Editor, though. Removed them...

screenshot


skj3gg:
I'll trust you, this time... :wink:

jawshooah:
Nice catch! Thanks for the fix :+1:

PR #256, opened 2015-04-29
detect new syntax definition file format .sublime-syntax

Doingbla:
Sublime Text build 3084 introduced a new syntax definition format [.sublime-syntax](http://www.sublimetext.com/docs/3/syntax.html). Views that use 
this new format are currently not detected by sublime linter. This patch solves that issue.


skj3gg:
Thanks! Released in v4.4 soon.

PR #259, opened 2015-05-01
Only return tmpfile_suffix if the view does not have a file name

ismell:
Some syntax definitions can support multiple file extensions. If the file already has a file extension we should use it, otherwise we can try and guess.


ismell:
ping?

ismell:
ping?

groteworld:
Sorry for the wait and thank you for your patience! Everythings seems in order here. :beers:

PR #261, opened 2015-05-08
Updated `find_python_script` to support MinGW64

vladipus:
Under a MinGW64 (MSys2) environment, the scripts are wrapped in .exe files and placed in the same directory as the Python runtime. The support for this scenario was added. The Python directory path was also refactored into a separate variable.

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -1033,8 +1033,9 @@ def find_python_script(python_path, script):
                 return out
         return which(script)
     else:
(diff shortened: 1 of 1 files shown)


braver:
This PR is outdated, but the original issue is still open. We'll keep tracking of MinGW64 support in #117, but I'm not keeping all the pull requests around till we have an idea of if/when/how we're going to approach it.

PR #264, opened 2015-05-10
demote_to_warning_match only checks for the first pattern

brocoli:
demote_to_warning_match would only check for the first pattern in the list instead of all patterns in the list. Indentation on a for-if-break was wrong, making it break out of the for on the first loop.


grotewold:
Thank you for your contribution and your patience!

PR #293, opened 2015-07-10
SUBLIME_LINTER_FILE and SUBLIME_LINTER_FOLDERS env variables

kkamkou:
Hi all!

[I'm using the docker container for eslint](https://github.com/kkamkou/docker-applications/blob/master/eslint.sh). The only problem is to extract the root folder from the sublime tree. This path is required for `.eslintrc` detection.

What I did is kind of sharing the information using the ENV variables. Please, review.

p.s. tests are not working with the latest version of py.test

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -360,7 +360,7 @@ class Linter(metaclass=LinterMeta):
     errors = None
     highlight = None
     lint_settings = None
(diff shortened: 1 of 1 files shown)


kkamkou:
I'll create a new one. Closed.

PR #318, opened 2015-08-25
Basic tooltip support

Templar-von-Midgard:
Added a tooltip when the cursor is in a line where errors are present.
Using the "tooltip_style" setting a CSS file can be given to format the tooltip.
Using the "search_engine" setting you can give a url like "https://google.com/#q=", to which the warning that you clicked on will be appended to open in the default browser.


MattDMo:
Bump :+1:

ondrejbartas:
Hi any reason why this is not merged???

Arkkimaagi:
This seems quite cool and solves the issue of multiple errors in same row getting lost. For some reason I had to edit it a bit to get it working, some string joints were giving me a hard time, but changing them to be like this worked:

```python
        for err in errors:
            divContent += '<p><a href="open:'
            divContent += syntax
            divContent += ' '
            divContent += err
            divContent += '">'
            divContent += err
            divContent += '</a></p>'
```

I have no idea why, as I do not know Python, but hey, it works.

One thing that comes to mind is that maybe there should be some small delay before opening the popup so that it does not flash around as much as it does now.

Would definitely like to see this merged, maybe start so that you could enable it with a configuration flag.

buildersbrewery:
This PR should be changed to make use of `on_hover(view, point, hover_zone)` from Build 3116, see [API Reference (ST3)](https://www.sublimetext.com/docs/3/api_reference.html).

For a minimal working example compare Makopo/sublime-text-tooltip-lsl#5.

---

### However

[Build 3118 introduces Phantoms](https://forum.sublimetext.com/t/dev-build-3118/21270) and I strongly believe they fit the nature of SublimeLinter better than tooltips.

groteworld:
Sorry for the radio silence, I have been away for a while now. I have evaluated all three tooltip PRs and decided on #510 moving forward. I appreciate your work, and the discussion that it sprouted. I am interestested in how we can use phantom vs. tooltips in Tooltip Themes.

Thanks you again for your time.

Issue #150, opened 2014-11-12
Feature: Ability to set working directory for linter


luke-tsekouras-sirca:
I'm using `pylint`, and I'm having trouble trying to get it to work with a particular `pylintrc` because of some `init-hooks` which adds a few folders to the python path to include extra modules in the project.  They rely on the current working directory being the root of the project.  At the moment the working directory looks to be set to the directory of the file being linted.  

I believe the `working_directory` option was available for SublimeLinter 2, so why is it not available for SublimeLinter 3?

braver:
Seems like #325 already addresses this. If not, let me know, or it will come up by itself again. 🤷‍♂️

PR #325, opened 2015-09-06
new linter setting -- `chdir`

smanolloff:
Allows to specify a linter's working directory (#150).
In case a value is not set or is invalid, the old behavior is used (try to change to the directory of the linted file).

### DIFF
--- a/docs/linter_settings.rst
+++ b/docs/linter_settings.rst
@@ -120,3 +120,29 @@ In :ref:`debug mode <debug-mode>`, |sl| logs each occurrence of an ignore match.
 .. note::
 
    |regex|
+
+
+chdir
+-----
+This setting specifies the linter working directory.
+
+The value must be a string, corresponding to a valid directory path.
+
+.. code-block:: json
+
+    {
+        "chdir": "${project}",
+    {
+
+With the above example, the linter will get invoked from the ``${project}`` directory (see :ref:`Inline overrides <_settings-tokens>` for more info on using tokens).
+
+.. note::
+
+     If the value of ``chdir`` is unspecified (or inaccessible), then:
+
+     - If linting an unsaved file, the directory is unchanged
+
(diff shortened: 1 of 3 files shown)


zenlambda:
I setup my project's python path in a pylintrc init-hook (the pylintrc is in my projects root directory). This patch is very useful because it allows Sublimelinter-pylint to read the pylintrc and get all the right settings.

I've tried this patch at it works for me.

Thanks @smanolloff !

grotewold:
Thanks for your addition @smanolloff!

aMarCruz:
thanks, SL **not** working now.
console display various:
`SublimeLinter: chdir not set or invalid, using ...blah...`
and nothing more.

Windows 7 + SublimeLinter 3

smanolloff:
Hi @aMarCruz, thank you for your feedback.

Can you please provide the following details:
1) have you specified a value for this setting in your configuration files?
2) which linter do you observe the issue with? Is it the same with all your linters?
3) the entire console log.

aMarCruz:
Hi @smanolloff , strange case.
I have installed eslint and jscs plugins.I disable eslint months ago, so the only enabled plugins were annotations and jscs.
Yesterday I disable SL, including their plugins through Package Control, as JSCS was unusable.

Right now (today) I went back to enable SL and all its plugins and is working perfectly!
I have not configured "chdir" so SL sends messages (like `SublimeLinter: chdir not set or invalid, using C:\MinGW\msys\1.0\home\<myself>\compiler\lib`), but find the correct file directory and runs jscs with no problem.

Yesterday that I had the problem, I tried setting "chdir" with `${directory}`, later with the absolute path for the current js file, the path of my project, delete the cache, etc. ...never works.
I don't tried testing annotations, sorry.

If I notice any issue I'll tell you.
Thanks for your attention.

PR #336, opened 2015-09-22
add config_joiner setting in linter class

ikappas:
This setting can be set to either  '=' or ':'.

The default setting of  ':' makes the config file argument and the value passed as separate arguments for backwards compatibility.

When this setting is set to '=', the config file argument is joined with the config file path found by '=' and passed as a single argument.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -281,6 +281,11 @@ class Linter(metaclass=LinterMeta):
     #
     config_file = None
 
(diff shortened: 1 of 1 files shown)


ikappas:
any chance this gets merged?

groteworld:
Sorry for the wait and thank you for your patience! Everythings seems in order here. :beers:

PR #362, opened 2015-10-29
added ${sublime} to point to the sublime packages directory, fixes #361

### LINKED ISSUE #361: Add ${sublime} for sublime settings directory
It would be convenient to have this, so sublime specific linter settings can be stored in the sublime text directory instead of the home directory.

simonzack:


grotewold:
LGTM

Issue #36, opened 2013-12-19
Linter plugins not available in Package Control 


creativeaura:
After updating to ST3 Build 3059 the plugin has stopped working. I have reinstall the plugin from package control but when I try to install linter plugins, none of the plugins are available in the list.

PR #387, opened 2016-01-01
Add PythonImproved to list of python aliases

MattDMo:
This addresses [MattDMo/PythonImproved #36](https://github.com/MattDMo/PythonImproved/issues/36). Python Improved now has over 16k installations, so I figured it's about time to implement this :grinning:

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -25,6 +25,7 @@
         "show_marks_in_minimap": true,
         "syntax_map": {
             "python django": "python",
+            "pythonimproved": "python",
             "magicpython": "python",
             "html 5": "html",
             "html (django)": "html",


jawshooah:
:+1: Thanks!

MattDMo:
@jawshooah No problem. Quick question, though, since I'm not intimately familiar with all of SublimeLinter's internals. I've read through a number of `linter.py` files from various linters, just to get a feel for how they work, and they all seem to define a `syntax` variable (such as in [`SublimeLinter-pylint`](https://github.com/SublimeLinter/SublimeLinter-pylint/blob/master/linter.py#L21)). With PythonImproved now in the `python` syntax map, does this mean that *all* linters that define `syntax = "python"` will now work with PythonImproved? I would assume so, I just wanted to make sure.

Thanks!

jawshooah:
Yup, that's the long and short of it :smile:

Issue #380, opened 2015-12-04
SublimeLinter ignore_match not working when extension is specified


colin-marshall:
I'm trying to get **SublimeLinter** and **HTMLTidy** to ignore certain php warnings. Specifically the  `missing <!DOCTYPE> declaration`, `inserting implicit <body>`, and `inserting missing 'title' element`. It tells you right in the docs how to do this:

```json
{
"ignore_match": [
    "missing <!DOCTYPE> declaration",
    "inserting (?:implicit <body>|missing 'title' element)"
    ]
}
```

Adding that to my SublimeLinter user settings file worked fine. The problem is that I want it only to apply to PHP files, because using just the above will also ignore it on HTML files.

The docs have an example for applying `ignore_match` only to certain extensions:

```json
{
"ignore_match": {
    "inc": [
        "missing <!DOCTYPE> declaration",
        "inserting (?:implicit <body>|missing 'title' element)"
        ]
}
```

So I copied that and changed the `inc` to `php` and I still get those warnings on php files.

Here is what the `htmltidy` section of my SublimeLinter settings file looks like:

```json
"htmltidy": {
            "@disable": false,
            "args": [],
            "excludes": [],
            "ignore_match": {
                "php": [
                    "missing <!DOCTYPE> declaration",
                    "inserting (?:implicit <body>|missing 'title' element)"
                 ]
            }
    }
```

Any idea why `ignore_match` works as it is supposed to until I specify an extension that the `ignore_match` rule is supposed to apply to? Thanks!!!

colin-marshall:
This is a bug. See the accepted answer on my StackOverflow question:
http://stackoverflow.com/questions/34057863/sublimelinter-ignore-match-not-working-when-extension-is-specified

grotewold:
Thanks to @alkuzad, in both SO and here, this has been corrected. Thanks for your help!

colin-marshall:
Thanks @alkuzad and @grotewold!

ctf0:
@groteworld am still having this issue
```json
"htmltidy": {
    "@disable": false,
    "args": [],
    "excludes": [],
    "ignore_match": {
        "blade": [
            "(<script>|<img>|<a>) escaping malformed URI reference",
            "inserting (?:implicit <body>|missing 'title' element)",
            "missing <!DOCTYPE> declaration",
            "plain text isn't allowed in <head> elements"
        ]
    }
},
```

PR #393, opened 2016-01-18
Fixes issue !380, nested dictionaries are lost during linter settings tokenization

alkuzad:
Hi

I've found some time to look into problem from SO: http://stackoverflow.com/questions/34057863/sublimelinter-ignore-match-not-working-when-extension-is-specified/34493594

I've found it's really simple, just for dict values `recursive_replace_value` function uses `recursive_replace` instead of `recursive_replace_value` and the first one does not return any value at all. This is why we have 'None' instead of dict with replaced arrays. So I've added parameter to control if the return should be invoked and voila ... tests are passing.

Sorry for this kind of crappy tests but I usually work with pure unittest and I do not know py.test so well. So these tests are rather basic but even with them I can prove that new version works (just compare same tests with branch and master linter.py to see)

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -532,7 +532,7 @@ class Linter(metaclass=LinterMeta):
         """
         def recursive_replace_value(expressions, value):
             if isinstance(value, dict):
-                value = recursive_replace(expressions, value)
+                value = recursive_replace(expressions, value, nested=True)
             elif isinstance(value, list):
                 value = [recursive_replace_value(expressions, item) for item in value]
             elif isinstance(value, str):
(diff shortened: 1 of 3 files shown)


alkuzad:
Issue: #380

grotewold:
You had some flake8 issues in your unit tests that I fixed, but thank you for taking the time the submit this fix, and for your support in SO.

alkuzad:
@grotewold thanks for correcting this, I was planning to do the needed flake changes but 02:00 AM was too late for me :) Debugging ST plugins is very hard as sublime_api module is builtin of ST python and I could not figure out option to debug with standard pdb. Fortunatelly for this case extracting two methods was enough - this suggests that these methods could be either static or in utils module.

PR #400, opened 2016-01-27
move intrusive message to status bar

titoBouzout:


groteworld:
While I think being so blatantly explicit helps some people realize nothing is wrong or working. I agree, it disrupts flow from the user experience.

Issue #402, opened 2016-01-29
@python setting does not work when install location is not standard on Windows


aawilson:
Resubmitted from here: https://github.com/SublimeLinter/SublimeLinter3/issues/226#issuecomment-176314162

In a nutshell: If Python is installed anywhere but `c:\Python<foo>`, SublimeLinter3 can't find it with the `@python` meta, and there's no way to specify a location for that lookup.

braver:
We have been making [many changes](https://github.com/SublimeLinter/SublimeLinter/blob/master/messages/4.0.0-rc.1.txt) to SublimeLinter, some of which also should have fixed this particular problem. It would be really nice if you could [try the beta](https://github.com/SublimeLinter/SublimeLinter/tree/15bb4a2e2024a623f60bb444eadd5790764fdfb9#sublimelinter-4-beta) and let us know if that fixes the problem for you.

kaste:
Outdated. I sure fixed that. 🤞

PR #405, opened 2016-01-31
Honor user paths

alexandrul:
There is a bit of code duplication, but the alternative would be to complicate things by pairing dirs with some kind of origin indicator, since user dirs are trusted and we don't check the base folder naming conventions (my python 3.3 and 3.5 installations are named just 'python').

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -922,9 +922,16 @@ def find_windows_python(version):
         prefix_len = len(prefix)
         dirs = sorted(glob(prefix + '*'), reverse=True)
         from . import persist
+        user_dirs = persist.settings.get('paths', {}).get('windows', [])
 
         # Try the exact version first, then the major version
         for version in (stripped_version, stripped_version[0]):
+            for python_dir in user_dirs:
+                path = os.path.join(python_dir, 'python.exe')
+                if can_exec(path):
+                    persist.debug('find_windows_python: <=', path)
+                    return path
+
             for python_dir in dirs:
                 path = os.path.join(python_dir, 'python.exe')
                 python_version = python_dir[prefix_len:]


alexandrul:
Link to docs: http://www.sublimelinter.com/en/latest/troubleshooting.html#adding-to-the-paths-setting

alexandrul:
Fixes #402

mabuchner:
I think, the find_executable function should also be changed that way.

alexandrul:
@mabuchner this might be the root cause:

```
@lru_cache(maxsize=None)
def create_environment():
```

By removing the ```lru_cache``` decorator you should be able to notice that the first call returns an empty dict, and the rest of the calls are returning the actual user settings (at least this is the behavior on my PC).

Since I'm not that familiar with the code, maybe someone else could fix this issue a bit quicker.

Also, since this a complete different issue, is it possible to accept the current PR?

mabuchner:
Yes, you are right. Removing ```lru_cache``` actually "fixes" the problem.

alexandrul:
up :smile:

karolyi:
+1

alexandrul:
Actually, just checked on 3.7.1: removing `lru_cache` from `create_environment` fixes the issue, so my patch is not required for honoring the user paths.

Of course, if you choose to keep caching `create_environment`, my patch solves the initial issue.

groteworld:
While I understand removing the caching would fix the issue... I think it is better for the masses keep caching. Reason why this would loop over `path` setting over `python_path` though?

alexandrul:
@groteworld indeed, caching is better for the masses. For me, pyyaml is more important.

However, I'm not that familiar with the code to fix the root cause of this issue and since I have a workaround, I'm happy.

braver:
@groteworld @alexandrul whats the status of this thing? Is the PR anything near something we’d want to merge, or is it just a workaround?

alexandrul:
@braver fixing the root cause would be preferable and, given the insignificant number of affected users, I propose to reject this PR.

braver:
@alexandrul alrighty, sounds good to me. We still have #402 open if someone wants to look into fixing this for good.

PR #407, opened 2016-02-05
On Windows try to use py.exe to find python executable

bizoo:
Since Python 3.3 on Windows, a Python Launcher called py.exe is installed that call the correct interpreter depending of the argument or the shebang of the source file.
This Pull request try to call it to get the Python executable path.

Maybe it could be possible to always call py.exe (when available) with the correct argument instead of trying to find and call the correct Python executable. Need to look at it when I've some free times.

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -912,11 +912,10 @@ def find_windows_python(version):
 
         # Try to find python.exe using py.exe
         # Try the exact version first, then the major version
-        code = r'import sys;print(sys.executable)'
-        for version in (version, version.partition('.')[0]):
-            out = communicate(('py.exe', '-%s' % version), code)
-            if out:
-                path = out.splitlines()[0]
+        code = r'import sys;sys.stdout.write(sys.executable)'
+        for testversion in (version, version.partition('.')[0]):
+            path = communicate(('py.exe', '-%s' % testversion), code)
+            if path:
                 if ('python.exe' in path.lower()) and can_exec(path):
                     persist.debug('find_windows_python: <=', path)
                     return path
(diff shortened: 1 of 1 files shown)


groteworld:
I'll need to get a windows box setup to test this, unless @jawshooah has one and the time.

Please be patient.

bizoo:
No problem.
Notice that as Python 3.5 on Windows is now installed by default in "Program Files" folder, SL couldn't find it in the current version. With the modifications in this PR it works again.

lint/util.py
```diff
@@ -908,20 +908,30 @@ def find_windows_python(version):
...
+                    return path
+
+        # Check the root drive for 'Python*' directories, and try to match the
         # version based on the directory names. The 'Python*' directories end
         # with the <major><minor> version number, so for matching with the version
         # passed in, strip any decimal points.
         stripped_version = version.replace('.', '')
         prefix = os.path.abspath(os.path.join(
-            os.environ.get("SYSTEMROOT", "\\")[:2],
+            os.environ.get("SYSTEMDRIVE", "\\"),
```
rchl: This change has the same problem the code already had. Please see #490.
I like your change better than mine but you should adjust it to fix the problem I've fixed (append backslash).
The code could just say:
`os.path.join(os.environ.get('SYSTEMDRIVE', ''), '\\', 'Python')`

lint/util.py
```diff
@@ -908,20 +908,30 @@ def find_windows_python(version):
...
 
     if version:
-        # On Windows, there may be no separately named python/python3 binaries,
-        # so it seems the only reliable way to check for a given version is to
-        # check the root drive for 'Python*' directories, and try to match the
+        from . import persist
+
+        # Try to find python.exe using py.exe
+        # Try the exact version first, then the major version
+        code = r'import sys;print(sys.executable)'
```
rchl: Instead of using print() and then using splitlines() to cater for added newline, you could just use sys.stdout.write().

lint/util.py
```diff
@@ -908,30 +908,40 @@ def find_windows_python(version):
...
+
+        # Check the root drive for 'Python*' directories, and try to match the
         # version based on the directory names. The 'Python*' directories end
         # with the <major><minor> version number, so for matching with the version
         # passed in, strip any decimal points.
         stripped_version = version.replace('.', '')
         prefix = os.path.abspath(os.path.join(
-            os.environ.get("SYSTEMROOT", "\\")[:2],
+            os.environ.get("SYSTEMDRIVE", ""),
+            os.sep,
```
rchl: Come to think of it, you shouldn't really need the os.sep as those are arguments of os.path.join.

himalr: @rchl But it doesn't work when there's a drive letter at the beginning.

```python
>>>os.path.join("C:", "Python")
'C:Python'
```
This causes it to use "C:\Program Files\Sublime Text 3\Python" as the `prefix`.

You could use something like this,

```python
prefix = os.path.abspath(os.path.join(
    os.environ.get('SYSTEMDRIVE'),
    os.sep,
    'Python'
))
```

This will also fix some pylint related issues.

braver:
@bizoo If you could respond to the review, perhaps we can round this one up.

bizoo:
Well, I think everything must be OK in my last commit.

kaste:
Before merging: The new flake8 has a new method for finding pythons. https://github.com/SublimeLinter/SublimeLinter-flake8/blob/master/util.py#L20-L58

That tempts to be a unified approach for both POSIX and Windows, and it's very short. If the new flake8 doesn't break too much, this should be preferred over this approach here, I think.

braver:
Ok, I don't think we'll be making big changes in SL3 anymore, but SL4 should have this covered since we basically merged the new approach from flake8 (and by we I mean @kaste 😉 ).

PR #410, opened 2016-02-06
NodeLinter: check for linter binary in ancestor .bin directories

### LINKED ISSUE #229: Support nested package.json detection for npm modules
I have a project that looks like this:

```
.eslintrc
package.json
node_modules/.bin/eslint
sub-project
- package.json
- some-js-file.js
```

The general idea is that sub-projects share the parent repository and linting settings to ensure they stay consistent across the major project.

The existing logic to detect node binaries stops at the first package.json it finds. Could it be expanded to continue upwards when looking for a binary?

benmosher:
Starts at the local package's `package.json` directory and looks at successive ancestors' `node_modules/bin` folders for the linter command.

I did the minimum to fix #229. Not clear whether I followed project code style or conventions. Happy to update with feedback. 😅

### DIFF
--- a/lint/node_linter.py
+++ b/lint/node_linter.py
@@ -181,6 +181,11 @@ class NodeLinter(linter.Linter):
         if binary:
             return path.normpath(path.join(cwd, binary))
 
+        return self.find_ancestor_cmd_path(cmd, cwd)
+
+    def find_ancestor_cmd_path(self, cmd, cwd):
+        """Recursively check for command binary in ancestors' node_modules/.bin directories."""
+
         node_modules_bin = path.normpath(path.join(cwd, 'node_modules/.bin/'))
 
         binary = path.join(node_modules_bin, cmd)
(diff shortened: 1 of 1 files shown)


dozoisch:
I wish this gets merged soon :/

trekare:
+1

vvo:
This type of project structure is starting to be more common than before, react and babel being two good examples.

https://github.com/babel/babel/tree/master/packages
https://github.com/facebook/react/tree/master/packages

vvo:
I was not able to make this work (sublime text 3, sublime linter latest, sublime eslint latest)

dozoisch:
@vvo did you get any errors?I used benMosher fork directly.

Basically I added his git repo as a repository in sublime. Then installed SublimeLInter3 package.I then had to update the link in sublimer-linter-eslint to use the package SublimeLinter3 instead of SublimeLinter (since the name is different when checkouting from a repo) and it worked

vvo:
Yes I got weird python errors, maybe I messed up copy pasting/indenting :) It was about highliting smg like that.

benmosher:
Yeah, I recall it being a little bit dicey when using a fork locally. Several superstitious restarts of Sublime. @dozoisch's steps sound like what I did to develop the PR.

I would think copy-pasting the patch directly into your modules would work too though, unless Package Control did something to reset the state at startup...

vvo:
I am not at all a python dev so if this PR is solving it, deal, lets merge ;)

groteworld:
Sorry this took so long.

vvo:
Awe-some :)

vvo:
So I believe you will be releasing this soon? :D

groteworld:
3.7.0 has been released, it's still not visible in Package Control, but hopefully it gets picked up soon.

PR #418, opened 2016-02-22
Don't terminate on package.json with no node_modules/deps.

bryanhelmig:
Imagine a directory like this:

```
- package.json
- node_modules
  - .bin
    - eslint (we want linters to use this)
- app
  - package.json (empty, for managing absolute imports)
  - stuff.js
  - more
    - hello.js
- gulp.js
```

Only `gulp.js` would be linted by the proper local `node_module/.bin/eslint`. `stuff.js` and `hello.js` would revert to the "global" `/usr/local/bin/eslint` installation.

This change doesn't terminate on any `package.json` that don't have an accompanying `node_modules`.

### DIFF
--- a/lint/node_linter.py
+++ b/lint/node_linter.py
@@ -155,7 +155,9 @@ class NodeLinter(linter.Linter):
         name = 'package.json'
         manifest_path = path.normpath(path.join(cwd, name))
 
-        if path.isfile(manifest_path):
+        bin_path = path.join(cwd, 'node_modules/.bin/')
+
+        if path.isfile(manifest_path) and path.isdir(bin_path):
             return manifest_path
 
         parent = path.normpath(path.join(cwd, '../'))


bryanhelmig:
This was the source of some weird `maximum call stack size exceeded` - presumably because I didn't have the accompanying babel plugins installed globally. I get this error in Atom too but prefer SublimeText - hence the fix here. :smile:

bryanhelmig:
Might be a dupe of https://github.com/SublimeLinter/SublimeLinter3/pull/410?

bryanhelmig:
Fixes reported https://github.com/SublimeLinter/SublimeLinter3/issues/229.

e-cloud:
As I test, compareing with #410, this PR is more simple and elegant. Of course, both work.

dozoisch:
@groteworld Can this or #410 get merged? This is a very big pain point right now! Thanks. I'm currently using this fork directly in my sublime.

trekare:
+1

benmosher:
FWIW, I don't think this would solve my issue, as my nested package might have a `.bin` folder (without `eslint`).

So I think #410 is a superset of this PR.

groteworld:
Thanks for your contribution! Should be available here soon.

PR #444, opened 2016-04-15
allow linters to override how chdir is computed

trashhalo:


groteworld:
Sorry for the wait. Thanks for this addition!

Will be release with v3.7.2

Issue #51, opened 2013-12-28
Syntax and/or file specific linter settings


webuniverseio:
Hi, would it be possible to add syntax and/or file specific override settings for linters? So that in global settings for jshint I can have an option for indent: 4. But for <scripts> within html file indent can be set to something else. I'm not sure if it is supported or not yet. Thank you.

PR #452, opened 2016-05-12
Alternative Error Panel Flow

rgant:
I like the errors panel because at a glance I can tell if there are any problems somewhere in my code, but most of the time it's just full of TODOs which I am not really interested in clicking. However not clicking on an error in the Panel causes the file to jump around which means I frequently start typing somewhere I didn't want to.

So this patch makes two changes:
1. The first error is not highlighted by default. (SublimeLinter/SublimeLinter-jshint#51)
2. I've disabled the "cancellation" branch of the select_error function handling closing the quick_panel without selecting an error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: python
 
-sudo: false
-
 python:
   - "3.3"
   - "3.4"
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -78,6 +78,10 @@
         "command": "sublimelinter_choose_gutter_theme"
     },
     {
+        "caption": "SublimeLinter: Choose Tooltip Theme",
+        "command": "sublimelinter_choose_tooltip_theme"
+    },
+    {
         "caption": "SublimeLinter: Lint This View",
         "command": "sublimelinter_lint"
     },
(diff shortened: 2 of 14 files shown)


koleror:
Why has this PR been closed? It would have been helpfull for many of us!
An option to control that behaviour would be better though I guess

rgant:
@koleror The pull request seems to still be open to me. Issue #51 is the one that is closed. 

However, I don't believe that @aparajita is interested in supporting this feature request.

However if you want to add this to your SublimeLinter then you can easily edit the files in your install. Just checkout the changeset and comment out those lines.

koleror:
Oh my bad you're right. Sorry I was tired. And yes, I commented out the line by myself :)

adnanoner:
I'd love for this PR to get merged :). The show errors on save feature is great, but the jumping towards the first error just kills my flow. Is this not merged because of CI failures? Or are you not interested in this feature @aparajita ?
Thanks for your time anyway, SublimeLinter is an unmissable tool!

braver:
We're going to approach this differently, work is currently underway in #666, so this PR will not get merged.

PR #458, opened 2016-05-24
Correct Windows python path example in doc

smsrkr:
<major><minor> was missing from Windows python root folder name.


groteworld:
Thanks!

PR #468, opened 2016-06-22
Fix the text method to return the text of the view with the correct l…

Nevitones:
…ine endings


Nevitones:
I'm closing this pull request for now, since it can be affecting the error highlight position.

Any clue about it is very welcome.

Issue #476, opened 2016-07-04
JSON5 support


mathieumg:
I've been using SublimeLinter with SublimeLinter-eslint for a while and really enjoying it. I currently use the global `eslint` executable approach, but I'm trying to migrate to the local executable approach. Among the reasons for that is that there are more and more ESLint plugins getting added to my configuration and I need to install them all globally. The main reason, however, is that different projects can be using different (and possibly incompatible) versions of ESLint and ESLint plugins and that cannot work with the global approach.

It is my understanding that this can already happen automatically if SublimeLinter locates a `package.json` file in a folder, is that correct? [This line](https://github.com/SublimeLinter/SublimeLinter3/blob/3f2973811f1fbb38568677be7be7bfbe77f7f4f4/lint/node_linter.py#L155) seems to indicate so.

If so, the only obstacle preventing me from migrating to the local approach would be the fact that I use [JSON5](http://json5.org/) and a `package.json5` through [`npm-json5`](https://github.com/wiredmax/npm-json5) rather than the conventional `package.json` to get trailing commas and JSON comments. Would you be open for a PR that modifies the lookup above by also looking for a `package.json5`? If so, I imagine you would rather the `package.json5` lookup occurs after the `package.json` lookup fails to find a file, for each directory level.

Thank you!

braver:
Yeah, this is "won't fix", we need to focus on what's important to everyone. Also json5 and its npm fork seem to be pretty dead.

mathieumg:
`npm-json5` wildcard-installs npm because it just mocks the filesystem read function, but noted!

PR #477, opened 2016-07-06
Added support for `package.json5` manifest

mathieumg:
See #476.

I leveraged https://github.com/dpranke/pyjson5 and mentioned it in the files I took from it.

It got me thinking, though, is reading the `package.json` content really necessary? Wouldn't we be better off simply looking for the "binary" (`./node_modules/.bin/somename`) and then run that one?

Thank you!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
 
 before_script:
   - flake8 . --max-line-length=120 --exclude=__init__.py,linter-plugin-template,docs
-  - pydocstyle . --add-ignore=D202
+  - pydocstyle --add-ignore=D202 --match-dir='(?!json5).*?'
 
 script:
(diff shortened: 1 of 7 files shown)


braver:
@groteworld I don't know about you, but I generally don't like having to support "fixed" variants of industry standard solutions like this. It's not like there's nothing else to do. I don't believe this one has any serious usage in the wild. So.. unless you stop me I'm going to close it.

mathieumg:
My question still stands though, why is reading the package file absolutely necessary? We're looking for the "binary" to run, after all.

braver:
@mathieumg that's a fair question I guess, but has nothing to do with wether or not to support JSON5. You're welcome to open a new issue about this specifically.

PR #480, opened 2016-07-20
Include error or warning code in message.

fschulze:
This adds error or warning codes in messages. For Python this results in for example:

```
... E226 missing whitespace around arithmetic operator ...
```

This allows one to use that code to ignore certain errors in the config.


braver:
It's up to the specific linter to decide what to put in the message. If you want error codes in the messages, please take it up with the maintainers of that linter.

Issue #288, opened 2015-06-20
SublimeLinter pep8 PATH problems


HashixC:
Hi,

I'm trying to get SublimeLinter-pep8 working on SublimeText3 (I'm on windows)
I got this error from the console : SublimeLinter: WARNING: pep8 deactivated, cannot locate 'pep8@python' 

I'm looking at the troubleshooting section( at sublimelinter.readthedocs.org). And to verify my path I'm typing, in windows command prompt, "where pip", but nothing is found.
I have already added C:\Python34\Scripts to my PATH.

Can someone help me ? 
Thank you

jawshooah:
You need to install [pip](https://pip.pypa.io/en/latest/installing.html#install-pip) and [pep8](https://pypi.python.org/pypi/pep8).

HashixC:
I already installed pip.

jawshooah:
Strange. And it doesn't show up when you run `where pip`? Could you run `echo %PATH%` and post the results? And also verify that `pip.exe` is in `C:\Python34\Scripts`?

rchl:
I've investigated such problem myself before finding this issue and came up with a patch (#490) that should address this issue.

braver:
We have been making [many changes](https://github.com/SublimeLinter/SublimeLinter/blob/master/messages/4.0.0-rc.1.txt) to SublimeLinter, some of which also should have fixed this particular problem. It would be really nice if you could [try the beta](https://github.com/SublimeLinter/SublimeLinter/tree/15bb4a2e2024a623f60bb444eadd5790764fdfb9#sublimelinter-4-beta) and let us know if that fixes the problem for you.

kaste:
Almost sure this is solved.

PR #490, opened 2016-08-28
[Windows] Fix Python installations discovery for python@ type linters (issue #288)

rchl:
First two letters were taken from the SYSTEMROOT environment variable which is
incorrect as that would return 'C:' for 'C:\Windows'. Appending 'Python' to that
would end up with 'C:Python' which os.path.abspath would actually resolve to
'current working directory on C drive ' + Python which end up creating Path to
sublime directory.

Correctly take three characters so that we end up with C:\Python prefix.


rchl:
Superseded by #407

PR #491, opened 2016-08-28
Fix reference before assignment error

rchl:
'executable is not necessarily assigned at this point. Use cls.executable_path
'which is more appropriate anyway.


braver:
@groteworld I noticed some reports of this error being thrown. Don't really know how to test for this one but it seems simple enough...?

braver:
Looks to be a good solution, thanks!

PR #503, opened 2016-09-27
doc: add ubuntu .profile filename

scips:
.bash_profile is not used on ubuntu when not in terminal (see https://help.ubuntu.com/community/EnvironmentVariables#Other_files)
using .profile works perfectly


groteworld:
Sorry for your wait, thank you for your contribution!

PR #504, opened 2016-09-28
Support specifying a full path as a python version, thus enabling virtualenv support.

### LINKED ISSUE #70: virtualenv support
I work on both Python 2 and Python 3 projects and have various versions of them installed on my system.
- My system is OS X and I use homebrew to install Python 2 and 3.
- I also use virtualenvs to separate various Python 2 and 3 configurations.
- I want to have 'sublimelinter2' 'sublimelinter3' virtualenvs with the modules installed for python 2 & 3.
- I'd like to tell SublimeLinter3 to use these globally and on a per project basis.

Essentially, at a global level I'd like SublimeLinter3 to use a globally configured virtualenv for the modules as a minimum.  I don't want to mess with the system python or libraries at all.

Thanks.

nfvs:
This enables us to use a virtualenv python version by specifying a full
path in .sublimelinterrc.

Example:
{
    "@python": "/home/user/.myvirtualenv/bin/python"
}

Fixes #70.

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -38,7 +38,7 @@ STREAM_STDOUT = 1
 STREAM_STDERR = 2
 STREAM_BOTH = STREAM_STDOUT + STREAM_STDERR
 
-PYTHON_CMD_RE = re.compile(r'(?P<script>[^@]+)?@python(?P<version>[\d\.]+)?')
+PYTHON_CMD_RE = re.compile(r'(?P<script>[^@]+)?@python(?P<version>[\d\.]+|.+)?')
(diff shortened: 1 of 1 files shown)


groteworld:
Thank you for your patience, your contributions to the community have been amazing.

This will be released with #510 merging

nfvs:
Happy to help!

Triquetra:
Is there an ETA on when this very useful contribution will be released on packagecontrol.io?

SaintDako:
Looking at the commit, it looks like this would work with a relative path, too. Is this the case?

posita:
> Looking at the commit, it looks like this would work with a relative path, too. Is this the case?

@Dakkers, relative to what? The `.sublimelinterrc` file?

Issue #520, opened 2016-12-12
[feature request] Use Phantoms API to display errors/warning


flashios09:
Hi, 

Instead of displaying the error/warning in the **status bar** like this:

![image](https://cloud.githubusercontent.com/assets/1313482/21085654/b285183e-c014-11e6-856b-6dec488c7a3d.png)

Why we don't use the [Phantoms Api](http://www.sublimetext.com/docs/3/api_reference.html#sublime.Phantom) to display the **error/warning** like this:

![image](https://cloud.githubusercontent.com/assets/1313482/21085676/f2f37e92-c014-11e6-9864-42069d7d5f8a.png)

This is the blog post from the official ST website:
https://www.sublimetext.com/blog/articles/sublime-text-3-build-3124

Thanks

groteworld:
duplicate of #470

PR #521, opened 2016-12-13
Add Phantom to sublimelinter

flashios09:
Hi @groteworld, 

I added **Phantom** to sublimelinter #520 

image

### DIFF
--- a/lint/highlight.py
+++ b/lint/highlight.py
@@ -386,6 +386,7 @@ class Highlight:
 
     @staticmethod
     def hide_phantoms(view):
+        """Hide Phantoms."""
         view.erase_phantoms('sublimelinter')
 
     def draw(self, view):


lint/highlight.py
```diff
@@ -400,6 +433,28 @@ def draw(self, view):
                 region = sublime.Region(self.newlines[line], self.newlines[line])
                 gutter_regions[error_type].append(region)
 
+        def hide_phantoms(href):
+            view.erase_phantoms("sublimelinter")
+
+        for _line, _error in self.errors.items():
+            _column = _error[0][0]
```
rchl: Is it in line of current style to prefix local variables with underscore? It doesn't seem like it.

lint/highlight.py
```diff
@@ -389,6 +393,35 @@ def draw(self, view):
         """
         self.set_mark_style()
 
+        _stylesheet = '''
```
rchl: Should be a constant defined on top of the file.

lint/highlight.py
flashios09: @rchl , it's my first time with python(i'm php dev)
i used the `_` prefix to be sure that those vars will not affect the rest of the code
i think that you can optimize it ;)

rchl: I'm not the author of this plugin so I won't do anything with it.
You should just follow local style of the code.

lint/highlight.py
flashios09: @rchl i copy/paste it from the `Packages/Default/exec.py` 
ok i will update the code now

rchl:
I haven't investigated why but there is something wrong with those changes as randomly phantoms are not created. I see the squiggly line indicating the error but phantom is not shown. This happens randomly in the same file without any modifications.

flashios09:
> I haven't investigated why but there is something wrong with those changes as randomly phantoms are not created.

I test it with php and it works, i test it also with python flake8 linter and it works also.

rchl:
I assure that the code is incorrect.
Probably here:
```
        for highlight in self.all:
            all.errors = highlight.errors
            all.update(highlight)
```
Due to `highlight.errors` being non-empty on first iteration of the loop but empty on second. Then the errors will be empty and won't be drawn.

I don't know the codebase very well myself, but I think you made some design violations and shortcuts while making this change. The errors should probably be propagated through existing Highlight methods like it happens for non-phantom errors.

lint/highlight.py
```diff
@@ -400,6 +433,28 @@ def draw(self, view):
                 region = sublime.Region(self.newlines[line], self.newlines[line])
                 gutter_regions[error_type].append(region)
 
+        def hide_phantoms(href):
```
rchl: I meant to do this:
```
@staticmethod
def hide_phantoms(view):
  view.erase_phantoms('sublimelinter')
```
so that you have one method to remove phantoms. That way, if some handling of phantoms is changed, you only have to change one place.

flashios09: this will not work because `hide_phantoms` require the `href` attr(i don't know why)

flashios09: @rchl Fixed, i don't know what lamda does but it works now 👍

flashios09:
@rchl
> I haven't investigated why but there is something wrong with those changes as randomly phantoms are not created.

Please re-check now, if it persist i will have to wait for the plugin Author to fix it.

rchl:
> Please re-check now, if it persist i will have to wait for the plugin Author to fix it.

You haven't really changed any part of the code that could fix the problem.

PR #537, opened 2017-02-01
Prevent IndexError if strip_quotes returns empty string

silvasur:
If `strip_quotes` returns an empty string this

    for i, pos in enumerate((0, -1)):
        if near[pos].isalnum() or near[pos] == '_':

will fail with an IndexError:

      File "/home/silvasur/.config/sublime-text-3/Packages/SublimeLinter/lint/queue.py", line 65, in loop
        self.lint(view_id, timestamp)
      File "/home/silvasur/.config/sublime-text-3/Packages/SublimeLinter/lint/queue.py", line 111, in lint
      File "/home/silvasur/.config/sublime-text-3/Packages/SublimeLinter/sublimelinter.py", line 119, in lint
      File "/home/silvasur/.config/sublime-text-3/Packages/SublimeLinter/lint/linter.py", line 929, in lint_view
        linter.lint(hit_time)
      File "/home/silvasur/.config/sublime-text-3/Packages/SublimeLinter/lint/linter.py", line 1487, in lint
        col = self.highlight.near(line, near, error_type=error_type, word_re=self.word_re)
      File "/home/silvasur/.config/sublime-text-3/Packages/SublimeLinter/lint/highlight.py", line 334, in near
        if near[pos].isalnum() or near[pos] == '_':
    IndexError: string index out of range

This commit fixes this

### DIFF
--- a/lint/highlight.py
+++ b/lint/highlight.py
@@ -327,6 +327,9 @@ class Highlight:
         text = self.code[start:end]
         near = self.strip_quotes(near)
 
+        if near == '':
+            return 0
+
         # Add \b fences around the text if it begins/ends with a word character
         fence = ['', '']
 


braver:
Can you illustrate with a situation where this would occur?

silvasur:
(Sorry, should have documented this when I opened the issue)

For example when you have this invalid PHP file:

```php
<?php
foo("" "")
```

`php -l` will then return this error:
```
PHP Parse error:  syntax error, unexpected '""' (T_CONSTANT_ENCAPSED_STRING), expecting ',' or ')' in /tmp/foobar.php on line 2
```

SublimeTextLinter will then unquote the `""` part after the `unexpected`, which gives an empty string leading to the exception. SublimeTextLinter will then not recover, even after you fix the error in the PHP code.

Seems to always happen if you place an unexpected empty string `""` somehere, which can happen easily (forgetting a separating `,` for example)

Linter Plugin used: [Sublime​Linter-php](https://github.com/SublimeLinter/SublimeLinter-php).

PHP Version:
```
$ php -v
PHP 7.1.11-1+ubuntu14.04.1+deb.sury.org+1 (cli) (built: Oct 27 2017 14:28:45) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.11-1+ubuntu14.04.1+deb.sury.org+1, Copyright (c) 1999-2017, by Zend Technologies
```

Perhaps this is actually an issue in the PHP linter plugin? Perhaps it should have returned `'""'` and not only `""`? Not sure. If you think it's the PHP linter plugins fault, I can open an issue there.

braver:
Well it’s never a bad idea to guard against weird input. I’ll test this a bit and then merge. Thanks!

PR #550, opened 2017-03-11
Rvm performance via use_rvm_auto_ruby setting

smanolloff:
Add a `use_rvm_auto_ruby` boolean setting.

By default, SublimeLinter will try to use `rvm-auto-ruby` executable for linting. Although this guarantees that the latest RVM-activated ruby will be used, it has the downside of being too slow (it re-initializes the RVM environment on every invocation, which has a noticeable performance impact).

The rationale of this PR would be: RVM installed should not mean that `rvm-auto-ruby` is the only means of executing RVM-enabled ruby for linting. In my case, I do have RVM, but I initialize it only once, at Sublime boot, so paths are already in place and the linter will lint using the "correct" ruby without the need of any magic tricks -- it has the correct environment already set. See this gist:

http://gist.github.com/smanolloff/54a787f661a3b8b5af10541f5cc5fa2c

### DIFF
--- a/docs/defines.inc
+++ b/docs/defines.inc
@@ -16,6 +16,7 @@
 .. |cmd| replace:: Command Palette (:kbd:`cmd+shift+p` on Mac OS X, :kbd:`ctrl+shift+p` on Linux/Windows)
 .. |path| replace:: ``PATH``
 .. |regex| replace:: If you need help constructing, testing or debugging a regular expression, `regular expressions 101 <https://regex101.com/>`__ provides an easy way to do so.
+.. |customenv| replace:: For more information on how to customize Sublime Text's environment on Mac OS, see `this gist <https://gist.github.com/smanolloff/54a787f661a3b8b5af10541f5cc5fa2c>`__
 
 .. |_sl| replace:: `SublimeLinter <https://github.com/SublimeLinter/SublimeLinter3>`__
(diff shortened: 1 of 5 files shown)


smanolloff:
Closing prematurely as I imagine this causing a problem on POSIX systems where the same executable exists in multiple segments of the PATH -- in the cases where the rc file itself re-organizes the order of those segments, it wont have effect as now, instead of _replaced_, the computed PATH is _merged_ (with lower preference) to the default PATH.

I will need to find a solutions to this issue before re-submitting the PR.

smanolloff:
Pushed a new commit, safe to reopen PR -- the computed PATH is now merged with higher priority so it will not introduce any problems.

kaste:
Hi 👋 !

There are a couple of issues here, bc a lot has changed since your PR, but the general approach you show in your gist seems interesting. I think this could be integrated into SublimeLinter core, so users don't have to follow your gist to get it working. WDYT?

pykong:
@smanolloff  @braver @kaste Could we move this PR forward or close it?
It think it is bad if PRs just lay around abandoned and collect dust. 
We should avoid the [broken window](https://en.wikipedia.org/wiki/Broken_windows_theory) effect.

braver:
Let me know if we need to re-open this.

PR #510, opened 2016-10-17
Inline tooltip

iRyusa:
Hi there,

When creating our [MJML](https://mjml.io) plugin for SublimeLinter, I was really confused about the lack of a tooltip feature. I'm neither a Sublime Text User, and a Python programmer, so i'll need some help to finish this, but I have an early prototype using the Sublime Text Tooltip API : http://derniercri.d.pr/1lIFJ/4azXvKfE

TODO
- [x] Theme support
- [x] Disable/Enable Tooltip
- [x] Display tooltips
- [ ] Tests?
- [ ] User Theme?

Can you check this PR and give me some advice to make this feature available for everyone ?

PR #552, opened 2017-03-17
Inline tooltip [updated]

TheKevJames:
Copied from #510, see original discussion there.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -78,6 +78,10 @@
         "command": "sublimelinter_choose_gutter_theme"
     },
     {
+        "caption": "SublimeLinter: Choose Tooltip Theme",
+        "command": "sublimelinter_choose_tooltip_theme"
+    },
+    {
         "caption": "SublimeLinter: Lint This View",
         "command": "sublimelinter_lint"
     },
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -33,6 +33,9 @@
             "javascript (babel)": "javascript",
             "php": "html"
         },
+        "tooltip_fontsize": 16,
+        "tooltip_theme": "Packages/SublimeLinter/tooltip-themes/Default/Default.tooltip-theme",
+        "tooltip_theme_excludes": [],
         "warning_color": "DDB700",
         "wrap_find": true
     }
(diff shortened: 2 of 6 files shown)


TheKevJames:
Ah, my bad, didn't notice the codecov callback.

@groteworld it looks like we don't have much in the way of scaffolding for tests on mini-features like this one. How would you like me to proceed on building those?

groteworld:
No that's fine. It's more a work in progress to get tests in. I'll try to review this soon

outpunk:
@groteworld Any updates? Pleeeeeeease =)

pablo-coco:
there is a curse on this issue

TheKevJames:
@groteworld Any chance you have a rough estimate on when you'll find the time to look at this?

averypfeiffer:
@TheKevJames I know this may be an unpopular opinion. But I switched to VSCode from Sublime specifically because of this issue. They have a very nice implementation of tooltips over there, I haven't looked back since the switch. 

Just thought I'd mention it as a possible alternative.

TheKevJames:
You switched editors because one of the plugins for that editor has not yet merged an inline tooltip PR,? Seems like overkill to me. On a personal level, I've found VSCode (and Atom) to be noticeably slower than Sublime -- as a bunch of [benchmarks](https://blog.xinhong.me/post/sublime-text-vs-vscode-vs-atom-performance-dec-2016/) have found. I think I'll stick with Sublime.

Y'all know you can [install from source](http://www.sublimelinter.com/en/latest/installation.html#installing-from-source) literally just by checking out this branch into a folder, right?

groteworld:
Extremely sorry everyone, I know it can be frustrating to have this be so close yet so far. I've been speaking at a conference the last week so I've been extremely absent in everything but that. I'm traveling back today, and will try to look at it a little. at the very least I will review tomorrow. I Promise!

I have been having some roll out thoughts, and may create a config variable to turn this off, that will be set to false at default until a later release. Similar to a deprecation schedule, except the reverse. I feel this is the least disruptive to the current users, but informs them that this new feature is out. This will also give us some time to iron out issues before a huge release like this.

Kristinita:
@TheKevJames , @groteworld , please, look **#563**.

Thanks.

PR #579, opened 2017-04-30
Fix tooltip appearing after every re-highlight and view focus

maryqygao:
Bug: 
1. Tooltip pops up after every lint highlight update
2. Tooltip pops up on view focus

Expected:
- Linter tooltip should not appear on re-highlight or on view focus

To reproduce:
1. Highlight triggers tooltip
   - Enter code that will fail linter
   - Switch to another window before re-highlighting is finished
   - After a few seconds, linter tooltip will appear despite the original view not being in focus
2. View focus triggers tooltip
   - Select text with lint errors
   - Switch to another window and switch back
   - Tooltip appears automatically

This bug was caused by manually triggering `on_selector_modified` in the `highlight` and `on_activated` methods. I extracted the contents of the original `on_selector_modified` into a new method named `display_errors` with an optional flag `tooltip`. Within `on_selector_modified`, it calls the method with `tooltip=True`, while `highlight` and `on_activated` call it with `tooltip` defaulting to `False`.

Feedback is welcome! Currently the codebase seems to fail the lint test, [presumably due to the pydocstyle v2.0.0 update](https://github.com/SublimeLinter/SublimeLinter3/pull/577#issuecomment-298201887), so perhaps we should test with `pydocstyle . --add-ignore=D202,D401` for now. Let me know what you think.

### DIFF
--- a/sublimelinter.py
+++ b/sublimelinter.py
@@ -172,7 +172,7 @@ class SublimeLinter(sublime_plugin.EventListener):
                         window_views[wid] = other_view
 
         for view in window_views.values():
-            self.on_selection_modified_async(view)
+            self.display_errors(view)
 
     def hit(self, view):
         """Record an activity that could trigger a lint and enqueue a desire to lint."""
@@ -286,7 +286,7 @@ class SublimeLinter(sublime_plugin.EventListener):
             if persist.settings.get('lint_mode', 'background') in ('background', 'load/save'):
                 self.hit(view)
 
-        self.on_selection_modified_async(view)
+        self.display_errors(view)
 
     def on_open_settings(self, view):
         """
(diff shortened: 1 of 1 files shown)


braver:
I'm consolidating this and other related issues in #619

cmalard:
Hi @braver , just to check as your message makes me doubt: you wanted to close this PR? (it's not an issue)

braver:
Wow, thanks, I think I got confused yesterday. This is still open to be merged IMO

braver:
@maryqygao you've got two PR's open about tooltips. Is it possible to consolidate the two? Also, since you're already addressing some of the issues with tooltips, could you have a look at #619 and let me know which elements of that issue are addressed and which aren't? I'd love to get some traction on fixing the tooltips and if you're willing to help out that's beyond awesome. The efforts are missing clarity and focus right now, so that would really help us get your PR's in and improve the package.

braver:
I merged #577 and #579 and solved their merge conflicts in #636, so they're ready to be merged. Thanks for these contributions!

PR #581, opened 2017-05-04
Add ${root} setting token

capripot:
It gives the ability to `chdir` to root folder of a view when SublimeText is open in folder or project mode.

I used the var `folder` from [window variables](https://www.sublimetext.com/docs/3/api_reference.html#sublime.Window).

In the case of the Rubocop linter, `rubocop` uses the `.ruby-version` file [to set the targeted ruby version](https://rubocop.readthedocs.io/en/latest/configuration/#setting-the-target-ruby-version). Since the command is not run from the root folder of the project, and that the project file is not necessarly inside the root folder, `${root}` was needed to `chdir` in the right folder containing the `.ruby-version` file.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -66,8 +66,9 @@ After the default, user and project settings are merged, SublimeLinter iterates 
 =================== =========================================================================
 Token               Value
 =================== =========================================================================
-${sublime}          The full path to the Sublime Text packages directory
-${project}          The full path to the project’s parent directory, if available.
+${sublime}          The full path to the Sublime Text packages directory.
+${project}          The full path to the project file's parent directory, if available.
+${root}             The full path to the root folder of the current view in project or folder mode. Falls back to `${directory}` in single file mode.
 ${directory}        The full path to the parent directory of the current view’s file.
 ${home}             The full path to the current user’s home directory.
 ${env:x}            The environment variable 'x'.
@@ -77,11 +78,11 @@ Please note:
 
 - Directory paths do **not** include a trailing directory separator.
 
-- ``${project}`` and ``${directory}`` expansion are dependent on a file being open in a window, and thus may not work when running lint reports.
+- ``${project}``, ``${root}`` and ``${directory}`` expansions are dependent on a file being open in a window, and thus may not work when running lint reports.
 
 - The environment variables available to the ``${env:x}`` token are those available within the Sublime Text python context, which is a very limited subset of those available within a command line shell.
 
-Project and parent directory paths are especially useful if you want to load specific configuration files for a linter.
+Project, root and parent directory paths are especially useful if you want to load specific configuration files for a linter.
 For example, you could use the ``${project}`` and ``${home}`` tokens in your project settings:
 
 .. code-block:: json
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -482,6 +482,10 @@ class Linter(metaclass=LinterMeta):
         window = self.view.window()
         if window:
             view = window.active_view()
+            window_vars = window.extract_variables()
+            directory = (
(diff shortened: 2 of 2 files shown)


capripot:
How come `master` branch passes the tests on TravisCI? I branched out of master and touched none of these files.

maryqygao:
@capripot The 2.0.0 update for `pydocstyle` broke some tests, [so recent PRs have been failing on Travis](https://github.com/SublimeLinter/SublimeLinter3/pull/577#issuecomment-298201887).

capripot:
Made a PR for fixing the build https://github.com/SublimeLinter/SublimeLinter3/pull/583

capripot:
@braver Any way I could help to make this PR to be merged in master and get released?

braver:
someone just really needs to make time to test and review it and it's just not top priority for the little time that I have at this point.

capripot:
Rebased on last master. Could we include it in version 4 @braver ?

FichteFoll:
Maybe we should just use `sublime.expand_variables` with the window's variables here and add a few custom fields like `$sublime` and `$home`?
This functionality would then be covered by `${folder:$file_path}`.

braver:
I never used any of these settings so I find it really hard to review and decide on any of this. This PR looks sound so I don’t mind merging it. @FichteFoll usually says smart things so we can go with that, or just save that idea for a refactor later?

FichteFoll:
I don't know how that feature is used currently either. I don't even know the places where variables are replaced.

It's definitely a good target for a refactor, but not something hugely important I suppose. And even if we have to do it, it would be easy enought to re-implement the feature added in this PR to make it backwards compatible.

braver:
Aw heck, if @capripot says it's good it's good. Let's merge.

7cc

kaste:
I'm confused. Per doc `{project}` should already point to the root of the project. So `{root}` is just the combination of `{project} or {directory}`, @capripot?

There is a bug there, I can spot reading the code. But otherwise...

capripot:
The problem with `${project}` is that you need to be in a project and to have the project file in the folder containing the your project files in order to make it work, which is not always the case. `${root}` works either way. When you open a folder or when you open a project, regardless of the location of your `.sublime-project` file.

Where is the bug? Could you please comment the line you spotted?

kaste:
> Where is the bug?

It is in the original code. E.g. We have an early return https://github.com/SublimeLinter/SublimeLinter/blob/e14e93bc666c044419f0b70e7e7ea62a681d70e6/lint/linter.py#L524

> to have the project file in the folder containing the (sic!) your project files

I count this as a bug as well. Using https://github.com/SublimeLinter/SublimeLinter/blob/e14e93bc666c044419f0b70e7e7ea62a681d70e6/lint/linter.py#L535 to get a project path is problematic. As far as I know, if we have a project_file_name(), we can already use `extract_variables` which will give us the 'project_path'. Here, I don't know why you use `folder` and not `project_path`. But that's not important, it's about the original code.

Say overall, this pretends to be an improvement over `extract_variables`. But then this fishy way of getting the project path takes precedence over anything else.

https://github.com/SublimeLinter/SublimeLinter/blob/e14e93bc666c044419f0b70e7e7ea62a681d70e6/lint/linter.py#L534-L551

The one improvement is, it asks `window.project_data()` which can be populated even if `extract_variables` is not useful. (I also don't like it when the `for folder in folders` doesn't have a `break` statement, so we get probably the same token multiple times.)

So I think we should just ask `extract_variables` then `project_data()['folders']`.

FichteFoll:
The thing with extract variables is that it returned the first folder path always, iirc, but the file you are editing might be in the second folder of the project that does not have to be a subfolder of the first one.

PR #583, opened 2017-05-06
Fix version of pydocstyle for TravisCI

capripot:
Fix CI build by fixing version `pydocstyle` to 1.1.1

Required to make pass new PRs like #577 or #581

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
 
 install:
   - "pip install flake8"
-  - "pip install pydocstyle"
+  - "pip install pydocstyle==1.1.1"
   - "pip install pytest-cov"
   - "pip install codecov"
 


capripot:
Is anybody looking at these PRs? @TheKevJames @groteworld

TheKevJames:
Personally, I strongly disagree with pinning to an old version. If the maintainers do not want imperative phrasing, we should just `--ignore D401`; otherwise, we should just fix the comments.

capripot:
I agree with you @TheKevJames. It was rather a temporary patch to make the build pass, but a decision should be taken.

braver:
Yeah, pinning is not ideal. I'd much rather ignore the rule until someone feels like taking the time to address it. Not super important anyway. Closing this in preference of #588

PR #587, opened 2017-06-12
[Fix #586] Allow the linter to configure itself.

### LINKED ISSUE #586: Allow PATH option in user settings.
I know that linter executables should be in PATH in order for the linter to be enabled. My request is to remove that restriction. I have [implemented a path option](https://github.com/veelo/SublimeLinter-contrib-epcomp/blob/master/linter.py#L46) in the [user settings of my linter](https://github.com/veelo/SublimeLinter-contrib-epcomp#linter-configuration), which almost works. The problem is that SublimeLinter3 does not seem to read that file when it initially tries to enable linters. It works if I go into linter.py and make a change, then SublimeLinter3 reloads the linter, at which point it reads the user settings and the executable is found and properly executed.

So it appears the only thing missing is that SublimeLinter3 should load the user settings before it tries to find the executables when it enables linters.

veelo:
By running the supplied linter initialiser in `can_lint`, the linter can adjust its `executable_path`. This allows the executable path to be configured in the linter user settings, as in https://github.com/veelo/SublimeLinter-contrib-epcomp/blob/master/linter.py#L49.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1540,6 +1540,8 @@ class Linter(metaclass=LinterMeta):
         can = False
         syntax = syntax.lower()
 
+        cls.reinitialize()  # Allow the linter to configure itself.
+
         if cls.syntax:
             if isinstance(cls.syntax, (tuple, list)):
                 can = syntax in cls.syntax


veelo:
Managed to resolve issue #586 without patches.

PR #588, opened 2017-06-12
Fix build by ignoring imperative mood docstrings.

veelo:
Alternative to #583.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
 
 before_script:
   - flake8 . --max-line-length=120 --exclude=__init__.py,linter-plugin-template,docs
-  - pydocstyle . --add-ignore=D202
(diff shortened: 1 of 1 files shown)


braver:
@groteworld builds are failing, which isn't really inviting for contributors. This solution seems preferable over #583, and just merging this seems a lot less work than fixing all the comments. We could do that at some point, but I think the primary concern should be to keep people on board and contributing improvements.

groteworld:
These docs have been updated in favor of ignoring. Thank you for your contribution though!!

veelo:
Can you restart Travis on PRs with failing builds manually, or do we have to make a phony commit to trigger a rebuild?

groteworld:
I believe I can either restart the builds but if not i will manually run through the build. I'll look into that now. Thanks again!

PR #589, opened 2017-06-19
move executable variable initialization to stop missing variable exceptions

smarden1:
I've seen at least two of these issues already (https://github.com/SublimeLinter/SublimeLinter-jshint/issues/33 and https://github.com/SublimeLinter/SublimeLinter3/issues/88) and they have been closed, but wasn't sure why, as I had the same problem this morning.

This pr attempts to fix the problem where the `executable` variable is not defined, which seems to happen when can is set to `False`, therefore `executable` is never defined.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1539,6 +1539,7 @@ class Linter(metaclass=LinterMeta):
 
         can = False
         syntax = syntax.lower()
+        executable = ''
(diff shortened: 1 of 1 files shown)


braver:
this is a different solution, but duplicate nonetheless, of #491

PR #597, opened 2017-07-11
linter: Print debugging message if match fails

woodruffw:
This commit makes split_match print a debugging message if the match completely fails, meaning that no groups have been captured.

I ran into this problem while writing a linter plugin for SublimeLinter3 - I had a *very* trivial (and hard to pick out) mistake in my regular expression, and the lack of a debugging message telling me about a missing match made it difficult to pinpoint the failure.


groteworld:
Thanks and sorry for the delay

woodruffw:
Not a problem, thanks for merging!

PR #603, opened 2017-08-08
variable 'executable' referenced before assignment

lanfon72:
` UnboundLocalError: local variable 'executable' referenced before assignment `
error occur in L:1601
when `cls.executable_path is not None` and `status is None`


braver:
another variation of the same as #491

PR #612, opened 2017-09-19
Make events async; Seems to improve performance a lot

gerardroche:
There are some serious performance issues with SublimeLinter. I patched it about three months ago and forgot about it until I was [reminded of yesterday](https://twitter.com/appsforartists/status/909881393827561472).

The fix is to change the events to async calls. It seems to improve performance a lot. It doesn't fix the issue entirely, but it helps.

I'm not entirely sure what the implications of making these calls async is, or if there are any bc issues.

* https://twitter.com/appsforartists/status/909881393827561472
* https://forum.sublimetext.com/t/high-cpu-usage-on-mac-sierra-with-3143/31367/24

### DIFF
--- a/sublimelinter.py
+++ b/sublimelinter.py
@@ -46,7 +46,7 @@ def plugin_loaded():
     window = sublime.active_window()
 
     if window:
-        plugin.on_activated(window.active_view())
+        plugin.on_activated_async(window.active_view())
 
 
 class SublimeLinter(sublime_plugin.EventListener):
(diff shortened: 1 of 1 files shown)


gerardroche:
The tests are failing due a bunch of flake8 issues which don't really have anything to do with the PR.

braver:
I have been using/testing this for a while now and it's works perfectly as far as I can tell. All for merging this.

gerardroche:
There's no doubt a great deal more that can be done with the performance side of things, but this does certainly seem help a lot. I assume when these calls are no async it blocks, which in terms of performance is catastrophic really, because things like flake8 are slow, and SL runs after keystrokes. I think it's even worse when using something like vim emulation.

PR #617, opened 2017-09-22
tooltip line numbers are off by one

### LINKED ISSUE #560: [popups] Tooltip line numbers are off by one
The line numbers that show up in the tooltips are zero indexed, but SublimeText starts them at 1.

image

braver:
This doesn't really feel like an excellent solution, but it seems like it's just a presentation issue so why not. 

Fixes #560 

---------

Aside:
I noticed `open_tooltip` and `close_tooltip` are used only once. Wouldn't it be better to inline them so we don't need to hustle with the parameters?


braver:
Actually, come to think of it, why do we even display the line number? The popup is show *at* the line, at the error itself, so the line number is pretty obvious and even if it somehow isn't, it's not very relevant.

braver:
@groteworld are others also watching this repo or is it just you? Do you want me to request review like this, or mention you.. or I could just label things that I think are ready to go so you can go through a list of those things in your own time?

groteworld:
What linters has this been tested on. I don't use tooltips so I wasn't sure if just 1 linter isn't working or if this is consistent among many.

braver:
I don't use them much either, but for the occasion I tested with eslint, tslint, stylelint, jsonlint and flake8: all off by one.

PR #626, opened 2017-09-29
support for external status bar messages handler

mg979:
For support to my plugin

[StatusBarControls](https://github.com/mg979/StatusBarControls)

Imgur
Imgur
Imgur
Imgur

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -38,6 +38,7 @@
         "tooltip_theme_excludes": [],
         "tooltips": false,
         "warning_color": "DDB700",
-        "wrap_find": true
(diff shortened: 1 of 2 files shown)


braver:
I don't think having all status bar updates go through a single package is the best idea ever. The ST api also doesn't really allow you to do what your package attempts in a very clean way. In something like Atom perhaps, but not here. As such I don't think this is the time for SL to make changes to support your package.

mg979:
> I don't think having all status bar updates go through a single package is the best idea ever.

It does nothing like that.

> The ST api also doesn't really allow you to do what your package attempts in a very clean way.

I disagree, but I won't waste more of your time.

PR #633, opened 2017-10-06
enable jruby support

Torinson:
The ruby_linter does not work on Windows with jruby installations.
It deactivates all ruby related plugins (e.g. linter-rubocop) because it can not detect native jruby installations.

This PR fixes this issues and enables usage of SublimeLinter with natively installed jruby


Torinson:
Build fails due to issue mentioned in  PR #588 
After this issue has been merged, things should work out

PR #636, opened 2017-10-10
Make tooltips behave better

braver:
Replaces #577 and #579 and solves their merge conflicts. 

Currently our inline tooltips conflict with all other popups, usually hiding them as soon as they appear. This makes it impossible to use them in combination with e.g. GitGutter popups, and even standard auto-complete popups. They also appear at unexpected events, like when a window receives focus.

I still think our popups need a complete re-implementation, and therefore didn't really review if the code in this PR is optimal. As @maryqygao also mentioned:

> In the future, it may benefit to modify the tooltip to 1) bind to a less-frequently triggered event such as on_hover, and 2) display the tooltip for each highlighted section, not for a whole line.

This PR at least stops it from misbehaving too badly, which buys us time and peace of mind to consider real solutions.


groteworld:
Thank you both for your contributions!

PR #660, opened 2017-11-05
Tooltip content should be HTML escaped before

alin23:
I encountered `Parse Error`s on error messages containing ampersands (Rust dereference operator in my case) because the popup HTML parser treats that as a HTML entity.


braver:
LGTM, @groteworld?

Issue #544, opened 2017-02-21
Custom gutter icons: colorize warnings and errors differently


Kristinita:
### 1. Request

It would be nice, if in SublimeLinter will be possible make different colors, if user use not default gutter theme.

For example, I use [**Blueberry — round**](https://sublimelinter.readthedocs.io/en/latest/gutter_themes.html#standard-gutter-themes) gutter theme. I add in my color scheme file:

```xml
<dict>
    <key>name</key>
    <string>SublimeLinter: Gutter Mark</string>
    <key>scope</key>
    <string>sublimelinter.gutter-mark</string>
    <key>settings</key>
    <dict>
        <key>foreground</key>
        <string>green</string>
    </dict>
</dict>
```

I get result:

![Actual](http://i.imgur.com/bG8RbtY.png)

Warnings and errors now green.

### 2. Justification

1. Easy distinguish warnings and errors.
2. More customization — is good.

### 3. Desirable behavior

I add in my color scheme file:

```xml
<dict>
    <key>name</key>
    <string>SublimeLinter: Gutter Mark</string>
    <key>scope</key>
    <string>sublimelinter.gutter-mark.error</string>
    <key>settings</key>
    <dict>
        <key>foreground</key>
        <string>red</string>
    </dict>
</dict>
<dict>
    <key>name</key>
    <string>SublimeLinter: Gutter Mark</string>
    <key>scope</key>
    <string>sublimelinter.gutter-mark.warning</string>
    <key>settings</key>
    <dict>
        <key>foreground</key>
        <string>orange</string>
    </dict>
</dict>
```

I get result:

![Expected](http://i.imgur.com/xa34tWr.png)

Thanks.

braver:
We have been making [many changes](https://github.com/SublimeLinter/SublimeLinter/blob/master/messages/4.0.0-rc.1.txt) to SublimeLinter, and it includes this feature. You can [try the beta](https://github.com/SublimeLinter/SublimeLinter/tree/15bb4a2e2024a623f60bb444eadd5790764fdfb9#sublimelinter-4-beta) if you want.

Issue #299, opened 2015-07-23
Linter error priority


ezk84:
It would be useful if there was a way to assign a priority to linters. 

For example I am running both the `php` and `phpcs` linters. The first one checks syntax errors, the second one checks code style. 

Syntax errors should always 'override' code style violations: I want to see the syntax error rather than try to work out why the coding standard is being violated (because a missing parenthesis throws the whole coding style checker out of whack). Currently the `phpcs` errors shadow the `php -l` errors.

I see this working in two ways:
- for a given line a higher priority linter will display its errors before lower-priority linters.
- a higher priority linter will disable lower priority linters if it detects errors.

I prefer the first one myself.

braver:
The SL4 beta provides this feature for icons in the gutter, check it out!  https://github.com/SublimeLinter/SublimeLinter3#sublimelinter-4-beta

Priority for inline highlights would be super complicated, it doesn’t weigh up to the value it would provide.

Issue #661, opened 2017-11-06
Default settings can be edited


braver:
SublimeLinter currently prevents ST from packaging it to render the default settings etc. immutable. Therefore all package contents, including settings, can be edited by users, which can lead to unexpected problems and impossible to debug situations. 

The reasons for not packaging seems to be:

1. menus are generated (seemingly to reduce duplication)
2. the linter template copies files from the package 

1 Seems like an over-complication, it's better to have some duplication in static data files than having to deal with and maintain the templating code and magic file creation.  
2 Could also be dealt with differently. I would propose splitting the linter template off into its own repo and letting people start their new linter by forking the template. IMO this is more elegant on several levels.

Issue #652, opened 2017-10-29
Support ".hidden-tmTheme"


pykong:
In order to prevent modified `tmTheme` to pollute the list of available color schemes, they should be saved with the suffix '.hidden-tmTheme'.

See https://github.com/SublimeTextIssues/Core/issues/1663#issuecomment-294503789

Assign that work to me and it is done.

pykong:
Duplicate of #569

PR #665, opened 2017-11-12
Major Update: Scheme Generation, Custom Highlights and Cleanups

pykong:
Major update:
1) Support for the new json scheme format: `.sublime-color-scheme` ( `.hidden-color-scheme`).
#639 

2) Supporting custom highlights, in resolutions: **linters**, **warnings/errors** and even individual **error codes**.
#609 #274

3) In the wake of 1) and 2): Complete overhaul of scheme parsing.
- In the future, may allow for easy fix: #544
- For the new color scheme format, no patched color schemes are generated anymore: #292
- Gutter icon priority can be specified. Partly fixes: #299

4) Supporting `.hidden-tmTheme` file format to hide manipulated theme files from scheme selector.
(This is only relevant as long as tmTheme files are still supported)
#569 #652 https://github.com/SublimeTextIssues/Core/issues/1663

5) Removed base keys `"default"` and `"user"` to flatten settings file. Overhauled settings merge method.
#649

6) Implemented a settings converter for a seamless transition from old to new settings structure.

7) Some refactoring of the `lint` dir and its contents.
Partly addressing #650, without resolving the mutual import issue.

8) Converted name of the main file `"sublimelinter.py"` to `snake_case`. This is to conform to the correct conversion of PascalCased `SublimeLinter`.

9) Command palette no longer allows setting `"mark_style"` as this would conflict with new settings format.

10) Default values for various variables are now taken from the default settings file.
Requires #661


braver:
I may have some more feedback later. Two small things right now:

> Gutter icon priority can be specified. Partly fixes: #299

I don't think having priorities on inline regions is worth the effort if it's possible at all. I don't want to complicate things. Or do you feel addressing #299 fully would be easy?

Also, could you pull "upstream master" into your branch and resolve the conflict?

commands.py
```diff
@@ -500,19 +500,19 @@ def get_settings(self):
...
+            from .sublime_linter import SublimeLinter
             SublimeLinter.lint_all_views()
         else:
             linter.Linter.clear_all()
 
 
-@choose_setting_command('mark_style', preview=True)
-class SublimelinterChooseMarkStyleCommand(ChooseSettingCommand):
-    """A command that selects a mark style from a list."""
+# @choose_setting_command('mark_style', preview=True)
```
braver: We can no longer support commands like this. Please also remove the related menu entries, and check if there are docs about it.

lint/highlight.py
```diff
@@ -100,9 +131,12 @@ def draw(self, view):
     @staticmethod
     def clear(view):
         """Clear all marks in the given view."""
-        for error_type in (WARNING, ERROR):
-            view.erase_regions(MARK_KEY_FORMAT.format(error_type))
-            view.erase_regions(GUTTER_MARK_KEY_FORMAT.format(error_type))
+        # for error_type in (WARNING, ERROR):
```
braver: Some more commented code that can go

lint/highlight.py
```diff
@@ -249,20 +286,29 @@ def range(self, line, pos, length=-1, near=None, error_type=ERROR, word_re=None)
...
-            if mark.a == region.a and mark.b == region.b:
-                if error_type == WARNING:
-                    return
-                else:
-                    self.marks[other_type].pop(i - i_offset)
-                    i_offset += 1
+        if not style:
+            style = error_type
+
+        if style not in self.marks[error_type]:  # TODO: None handling
```
braver: Don't forget about the TODO!

lint/highlight.py
```diff
@@ -387,47 +430,82 @@ def draw(self, view):
         since each one potentially needs a different color.
 
         """
-        self.set_mark_style()
-
-        gutter_regions = {WARNING: [], ERROR: []}
-        draw_gutter_marks = persist.settings.get('gutter_theme') != 'None'
+        self.set_mark_style()  # TODO: does this still make sense?
```
braver: I think most of set_mark_style doesn't make sense anymore, although it also contains a check for show_marks_in_minimap. Needs a look anyway.

lint/highlight.py
```diff
@@ -387,47 +430,82 @@ def draw(self, view):
...
     def clear(view):
         """Clear all marks in the given view."""
-        for error_type in (WARNING, ERROR):
-            view.erase_regions(MARK_KEY_FORMAT.format(error_type))
-            view.erase_regions(GUTTER_MARK_KEY_FORMAT.format(error_type))
+        # # TODO: clear all regions
+        # for error_type in (WARNING, ERROR):
+        #     view.erase_regions(MARK_KEY_FORMAT.format(error_type))
+        #     view.erase_regions(GUTTER_MARK_KEY_FORMAT.format(error_type))
+        persist.region_store.del_regions(view)
```
braver: If `clear` only calls another function, just replace it everywhere.

lint/settings.py
```diff
@@ -0,0 +1,304 @@
...
+
+        if "gutter_theme" in self.changeset:
+            self.update_gutter_marks()
+
+        # TODO: change whether this optimal:
+        Linter.reload()  # always reload
+        from ..sublime_linter import SublimeLinter
+        SublimeLinter.lint_all_views()
+
+        # TODO: what does this do?
```
braver: I dunno man. Just remove it and we'll dig it out of the git log or whatever if something breaks.

lint/util.py
```diff
@@ -412,9 +270,9 @@ def build_submenu(caption):
     if setting == 'lint mode':
         from . import persist
         names = [mode[0].capitalize() for mode in persist.LINT_MODES]
-    elif setting == 'mark style':
-        from . import highlight
-        names = highlight.mark_style_names()
+    # elif setting == 'mark style':
```
braver: I think this can go

lint/util.py
```diff
@@ -1406,65 +1244,14 @@ def __exit__(self, etype, value, traceback):
...
 
 
 # menu command constants
 
-CHOOSERS = (
-    'Lint Mode',
-    'Mark Style'
-)
+CHOOSERS = ['Lint Mode']
+    # ,'Mark Style'
```
braver: some more comments to clean out

sublime_linter.py
```diff
@@ -7,7 +7,7 @@
 # Project: https://github.com/SublimeLinter/SublimeLinter3
 # License: MIT
 #
-
+3
```
braver: eh wut?

braver:
Also, when switching color schemes in ST 3143 I get this error:

```
reloading settings Packages/User/Preferences.sublime-settings
XmlScheme.generate_color_scheme called.
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/legacy.py", line 76, in generate_color_scheme_async
    with open(mod_scheme_path, 'w', encoding='utf8') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/User/SublimeLinter/Mariana (SL).hidden-tmTheme'
```

lint/legacy.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # Set the amended color scheme to the current color scheme
+        scheme_path_rel = self.packages_relative_path(
+            os.path.join(self.paths["usr_dir_rel"], os.path.basename(mod_scheme_path)))
+
+        # TODO: is there another way to prevent entering vicious cycle?
+        self.set_scheme_path(scheme_path_rel)
+
+    def gen_xml_nodes(self, unfound):
+        """"""
```
braver: empty comment?

braver:
On 3154 I get this when changing color schemes:

```
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/scheme.py", line 196, in generate_color_scheme_async
    if not unfound and not self.nodes:
AttributeError: 'JsonScheme' object has no attribute 'nodes'
```

lint/highlight.py
pykong: We can refactor this another time. For now, I have just removed the comments.

lint/legacy.py
pykong: 'pydocstyle' is such a bully.

pykong:
0ea1bc4 fixes issues mentioned in this code review. 
However, due to the redone status bar message, new bugs have been introduced.

lint/highlight.py
pykong: `self.set_mark_style()` is still needed to set the visibility of marks in the minimap. 
I have adapted the code respectively.

sublime_linter.py
```diff
@@ -363,124 +379,106 @@ def display_errors(self, view, tooltip=False):
...
-        """
-        If no tooltip theme has been configured, returns False.
 
-        Otherwise, returns a Template object with the template text.
+        def count_msgs(key):
+            return len(line_dict.get(key, []))
 
-        """
-        tooltip_theme = persist.settings.get('tooltip_theme')
+        status = "SublimeLinter: ⚠: {} ⮾: {}".format(
```
braver: These unicode characters aren't working: too small to be readable, and I'm getting a missing glyph. Let's just use E and W.

screen shot 2017-11-18 at 21 03 40

pykong: As I stated 'configuration of the message still needs to be implemented'.
I will implement it, of course.

braver: okidoki, just putting it on the checklist here 👍

pykong: OK. Implemented it. It's `W` for warnings and `E` for errors now.

pykong:
122c0c0 addresses #619 by adding the command `sublimelinter_line_report` which will make a popup showing all lint messages for the current line appear. As of that commit, popups can no longer appear without explicit triggering. This means a cleaner UX.

Lint messages are shown on the fly in the status bar still.

pykong:
#663 no longer relevant. The underlying issue hs been solved by cleaned up lint and settings logic.

pykong:
As of 85ada27 line report popups will now show on hovering the respective gutter icon.

line_report_popup

The popup appearance may be further enhanced.
E.g. The icons may be dynamically set according to the chosen gutter theme.
CSS and formatting work may be done.

This is, however, no job that I would want to perform, nor, do I think it is particularly pressing before release.

pykong:
The new 'goto next/previous error' method under `commands.py`can be easily modified to address enhancement request: #659 (**If we decide to do so at all.**)

However, this is not what we should address leading up to the next release.

braver:
> If we decide to do so at all 

yeah, I don't think supporting anyone else's plugin should be our priority until we get a bunch of our own features in a better place.

Love the line report popup, I think it (together with the new status bar info) should replace the inline popup we have right now.

pykong:
> Love the line report popup, I think it (together with the new status bar info) should replace the inline popup we have right now.

Inline popup? The line report popup is the new inline popup! ;-)
You may either hover over the respective gutter icon but still can produce the popup by shortcut!

Great, you love it, though!

Default.sublime-commands
```diff
@@ -1,38 +1,4 @@
 [
-    {
-        "caption": "Preferences: SublimeLinter Settings",
```
braver: Actually, it’s kind of a convention to have this and the one for keybindings. Lots of packages provide them and it makes them easy to access by keyboard.

pykong: What about keybindings?

pykong: Put both back in with 2668f20.

lint/legacy.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def generate_color_scheme_async(self):
+        """
+            Generate a modified copy of the current color scheme that contains SublimeLinter color entries.
+
+            The current color scheme is checked for SublimeLinter color entries. If any are missing,
+            the scheme is copied, the entries are added, and the color scheme is rewritten to Packages/User/SublimeLinter.
+
+            """
+        print("XmlScheme.generate_color_scheme called.")
```
braver: Can we only print these in debug mode?

pykong: 5560c90 Done. Removed all `print` statement I have introduced.

Default.sublime-commands
pykong: Now that issue has been addressed, can you close that change request, because it has become a distraction by now. Thanks.

braver:
Open issues:

- [ ] In settings: styles -> After entering a scope that isn't in the current color scheme (e.g. `bla`), it will complain for every next change even if you enter a valid scope. 
- [ ] After every change, all highlights are removed and then restored, making them "blink" under certain circumstances.

Feature requests:

- [ ] Tooltip on hover over a highlight, displaying the errors for that specific position (which may be multiple, but it shouldn't show all errors for the line)
- [ ] Replace `Report (Open Files)` and `Show All Errors` with an output panel, like the one in [LSP](https://packagecontrol.io/packages/LSP).

We could consider dropping the new line report popup in favour of these two features, but it's ok to first try and see how it feels and then decide. I think reporting per line isn't all that natural, but perhaps that depends on the linters you're using.

FichteFoll:
> I think reporting per line isn't all that natural, but perhaps that depends on the linters you're using.

Just keep it for the gutter, maybe. Sometimes a column isn't specified, but I guess in that case the entire line could be highlighted for multiple errors, in which case they'd still show up on hover.

pykong:
@FichteFoll you can indeed open the line report via the _interims_ key combination is 
<kbd>ctrl</kbd> +  <kbd>alt</kbd> +  <kbd>shift</kbd> +  <kbd>l</kbd>

pykong:
> Tooltip on hover over a highlight
This can be implemented. However, I wonder what will happen if more than one popup is attached to a region. I do not like to fight other plugins or SublimeLinter's 'reference/definition' popup.

A candidate for such conflicting region is shown in the gif below.

multi_popup

braver:
> However, I wonder what will happen if more than one popup is attached to a region.

Good point. Same goes for the line report though, if I'm running GitGutter with its hovers enabled, I'll never see the linter errors. But I don't have any hovers like goto definition enabled, so the error region hovers would work just fine for me. 
Perhaps that tells us that users should be able to turn on or off all hovers, on the gutter or elsewhere, so they're in control.

pykong:
We can and probably should allow the user to disable the line report on gutter icon hover.
I have no conflicts with GitGutter. They work in perfect harmony.
I wonder have you put ```sublime_linter.protected_regions``` in GitGutter's `protected_regions` settings?

pykong:
Also, the line report popup should respect
lr_sing_plr

 singular/plural for error warning count.

Either implement that logic (it is in the gitlog) or alter popup to:
```warning: 3```

**Comment by braver** [comment]
> I wonder have you put sublime_linter.protected_regions in GitGutter's protected_regions settings?

Yes, but I think this only protects against overwriting the icon, there are no checks against those regions here: https://github.com/jisaacks/GitGutter/blob/9e5ab88c1014ba0ed07e66b8e09178d52144548d/modules/events.py#L92

**Comment by pykong** [comment]
The much-requested feature of tooltip showing on hovering a highlight is now implemented.

on_region_hover

Maybe we should make that highlight look different from the full line report.
For example, we could put back in the line number into the full line report.

**Comment by braver** [comment]
I think we can leave out the headers with the counters for the highlight tooltip.

Also, is the highlight tooltip aligned with the cursor position? In the gif it looks like it’s at the start of the line.

Also, why are the headers indented in your screenshots and why is there so much padding on the popups? What theme and color scheme are you using?

**Comment by pykong** [comment]
I did not modify the popup template. Maybe, as you implied, it indeed has to do with my theme and color scheme.

I am using `Seti_UI` as the theme and `Seti_UX` as my color scheme.

**Comment by pykong** [comment]
Note: In 03974ac gutter themes are currently broken, as I am in the middle of refurbishing the underlying logic to set gutter themes.

The initial error when switching to gutter theme "Hands" was caused by the respective `Hands.gutter-theme` file being empty. This is not a bug I have introduced. 

Please have a look whether there is an issue open for broken gutter theme settings logic already.
Somebody must have noticed such blatant error before. 
We could then reference that issue here.

**Comment by braver** [comment]
Hmmm, perhaps I broke the gutter themes somehow...

**Comment by pykong** [comment]
> Hmmm, perhaps I broke the gutter themes somehow...
Not sure what happened. I'll fix it and make it better than it used to be before.

Now you can set a gutter themes just by providing its name, instead of giving the full path to the `.gutter-theme` file. _(Full paths will still work, for reasons of backward-compatibility.)_
image

I am right now extending this to the icons itself, so can either state the full path to an icon or just state its name like:
```
"icon": "warning"
```
This will make settings editing easier and also allow easy switching of themes without breaking paths. 
Hopefully, this will spur the development of gutter themes with more than two icons to give a more lively experience with the custom highlighting.

**Comment by pykong** [comment]
I have now implemented a *very early* version of our new error panel.

image

You may test it by hitting <kbd>ctrl</kbd> + <kbd>alt</kbd> + <kbd>shift</kbd> +<kbd>a</kbd>
(If panel appears empty close it and hit again.)

_Please do not comment on code quality or functionality, yet. This is the earliest working version of the panel which gives meaningful output and much will still need to be changed and added._

**Comment by pykong** [comment]
`force_xml_scheme` is not part of the default settings file. 

However, you can add it to the user settings to activate scheme generation via the old `.tm-Theme` file. 
I like to keep the possibility as long we need to keep that backward compatibility, i.e. as long as the new `.sublime-color-scheme` format does not hit SublimeText's stable branch.

**Comment by pykong** [comment]
Just ignore the next commits until I tell the plugin is working again.
Currently, the plugin is undergoing some refactoring in preparation for the panel implementation.

**Comment by deathaxe** [comment]
Keeps crashing on my Win64 box:

```
SublimeLinter: Traceback (most recent call last):
  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\lint\queue.py", line 57, in loop
    item = self.q.get(block=True, timeout=self.MIN_DELAY)
  File "./python3.3/queue.py", line 175, in get
queue.Empty

During handling of the above exception, another exception occurred:

  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\lint\queue.py", line 65, in loop
    self.lint(view_id, timestamp)
  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\lint\queue.py", line 111, in lint
  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\sublime_linter.py", line 266, in lint
  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\lint\linter.py", line 1000, in lint_view
    callback(cls.get_view(vid), linters, hit_time)
  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\sublime_linter.py", line 319, in highlight
    highlights.draw(other_view)
  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\lint\highlight.py", line 133, in draw
    all.draw(view)
  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\lint\highlight.py", line 456, in draw
    icon = self.style_store.get_val("icon", style, err_type)
  File "C:\Apps\Sublime\Data\Packages\SublimeLinter\lint\style.py", line 37, in wrapper
    if res != os.path.basename(res):
  File "./python3.3/ntpath.py", line 249, in basename
  File "./python3.3/ntpath.py", line 220, in split
  File "./python3.3/ntpath.py", line 162, in splitdrive
TypeError: object of type 'NoneType' has no len()
 
SublimeLinter: -------------------- 
```

deathaxe:
Done.

braver:
@deathaxe if you want to try the new features, the `next` branch is the one to watch. I try to keep that relatively stable, while @pykong is working on some breaking changes in his fork.

pykong:
@braver Good idea, we need to communicate this more clearly.

Still, regarding the bug @deathaxe reported it likely affects the "stable" branch as well, since I have not touched that part of the code for some time.

braver:
I'm going to close this PR. The change set is too big to review inside GitHub anyway.

We're not done yet though, please follow #666 to find out what's happening. 

There will be a beta release, with more opportunities for review, feedback and questions. However, if you can't wait you're welcome to try out the `next` branch, which should be stable enough for daily use (but might break any day).

If you want to see the real bleeding edge, @pykong is working on that over at https://github.com/pykong/SublimeLinter3/tree/scheme_style_major

Issue #649, opened 2017-10-28
Simplify Settings Structure


pykong:
Currently `sublimelinter.sublime-settings` contain the top level key 'default' or 'user' respectivly.
This is unnecessary as **ST** discerns default and user settings automatically based on the location of the respective settings files.

Also, **ST** automatically merges both dictionaries.

Therefore it would be best to simplify settings structure in accordance with the convention for plugin settings.

The method to handle merging of default and user settings in `lint/persist.py` will likely be rendered redundant and/or defunct by these changes. The code needs to be adapted accordingly.

*Assign to me.*

Issue #639, opened 2017-10-14
Implement color-scheme handling on basis of new `.sublime-color-scheme` format


pykong:
With the dev 3149 a new color scheme format `.sublime-color-scheme` has been introduced.
The most interesting feature is the capacity of automatically overwriting the default color scheme, as it happens with other **ST** formats.

https://forum.sublimetext.com/t/dev-build-3149/32549

This immediately implies a better way of handling color schemes in **SL**, compared to the current way of modifying the current scheme. Once this new color scheme format founds its way into the stable branch of **ST**, **SL** should switch over. 
Possible **SL** could support both old and new color scheme handling in parallel and switch depending on the current sublime version.

If this is desired by the maintainers of **SL**, I can well do the coding work required.

braver:
I think we will have to support the old style as well, unless ST does an on the fly conversion of old color schemes. However, the new approach is an excellent fit for our use case and we should try to support it ASAP and not wait for it to land in stable. That way we have a chance of testing it with early adopters and provide jps and wbond with feedback to make it the best it can be.

I’m not sure I read it correctly, but I think the new builds will also give us `—redish` and `—yellowish` variables to work with.

If you’re up for coding something, I’ll definitely help with testing and review.

pyking2:
@braver Thanks for the fast reply.

I am up for coding it. The other new feature I am currently implementing for **SL** and you were __so__ curious about deals with the same parts of the code. As I have already cleaned up the current color scheme generation, it is the most logical thing I code support for the new color theme system as well.

*I get to coding it starting tomorrow.**
I am thankful for your offer of testing and review and most likely come back to it!

I noticed the on the fly generated scopes for easier coloring from **3148**. I have still to wrap my mind around how this would profit **SL**.

Citing jps from the mentioned forum post:
> (.sublime-color-scheme files can override rules in both other .sublime-color-scheme files, and .tmTheme files)



For me this reads that the new color scheme system can entirely replace the old one.
So as long it is implemented at all (version 3149+), we do not need to care about the actual theme file type (`.tmTheme` vs `.sublime-color-scheme`)

braver:
> For me this reads that the new color scheme system can entirely replace the old one.

Oh wow, that’s pretty awesome. Looking forward to testing your contributions!

thatryan:
[...]

Issue #619, opened 2017-09-24
Improve interaction with tooltips (popups)


braver:
- tooltips follow the cursor, which is super annoying
- they don't disappear when the lint-issue is resolved
- when you move the mouse the tooltip disappears for some reason
- but hovering over an error-region doesn't open the popup (which would be so much more useful than on_selection_change)
- ~~when the popup disappears it hides all popups, making it fight with plugins like GitGutter~~
- ~~for some unclear reason it displays the line number (it is *at the line*!) but it's off by one~~

Also, there are 2 themes with no apparent advantage of one over the other. You can configure the font-size via a setting, but if the template would [follow the new guidelines](https://www.sublimetext.com/docs/3/minihtml.html#best_practices) there would be no need for that. 

Suggestions:

- hover over the gutter to display all errors for that line
- hover over an error range to display that specific error
- remove templates and create 1 very simple tooltip

This supersedes issues: #568, #565, #564, #560

#471 has some more info about making the tooltips open on_hover instead of on_selection_change

PR #579 will fix some of this, as will #617, however I don't think fixing this issues by issue is going to work. Some of the code is probably salvageable, but overall tooltips need to be completely re-implemented for the new stable 3.0 api.

Edit: updated because since recent events it's not so bad anymore

Cerberus-tm:
I mostly agree with the suggested changes. What made me turn off tooltips is that they activate when you click on the line (but outside the error area), and when you are still typing when you pause to think or look at your code. That tends to obscure my view of the line I'm typing and the one above/below. And it makes e.g. autocomplete disappear, when the tooltip appears as you're typing.

Instead of 'on hover', I'd prefer for the tooltip to appear when you **click** on the error (the code inside the red or yellow rectangle); and for all tooltips of the line to appear when you click on the error symbol in the margin.

dsl101:
Sorry if this isn't the right place—I got a bit lost in the thread of related issues. I've recently added this flag at line 469:
```
active_view.show_popup(tooltip_content,
   flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY | sublime.COOPERATE_WITH_AUTO_COMPLETE,
   location=-1,
   max_width=600)
```
Works much better now when linting picks up a half-finished bit of typing, and then won't go away. I really wish there was HIDE_ON_TYPING—that would solve it for me :).

dsl101:
So,, you don't like that extra flag?

braver:
[...]

Issue #618, opened 2017-09-24
Improve status bar messages


braver:
- split into **count** and **message**
- disable either part
- also display **count** when zero
- configurable template for **count**

Example:

`Errors: 0, Warnings: 1, this is a linter error message `

Or

`e:1 w:2, this is a linter error message`

Consolidates issues #563, #542, #338, #324, #523

pykong:
Assign that to me. Consider it done.

braver:
Looking forward to the PR! GitGutter does this really well IMO, that's what I had in mind.

pykong:
Thanks for the tip. I have a look into GitGutter.
I am going to start working on this issue once my other update has been successfully merged.

pykong:
@braver Put the assign label on me.

braver:
Can’t seem to do that, but it’s all yours!

PR #666, opened 2017-11-13
WIP: Next

### LINKED ISSUE #115: list errors in an output panel
The error list in [AtomLinter](https://atom.io/packages/linter) looks much better than SublimeLinter's.

SublimeLinter:
[SublimeLinter](https://fs.fnkr.net/LdifCunV)

AtomLinter:
[AtomLinter](https://fs.fnkr.net/KczF7LLo)
AtomLinter

braver:
A PR to track a possible upcoming (pre)-release where we combine some new features with a bunch of cleaning up.

Most of the work done by @pykong at PR #665 _(closed)_

fixes #662,  #661, #649, #639, #619, #618, #609, #274, #292, #299, #569, #663, #279, #348, #394, #457, #460, #461, #524, #557, #570, #452, #212, #115, 

The number of this PR should tell you something 🤘 

---------------

Open issues:

- [x] Bugs  
  - [x] check correct working of icon fetcher
  - [x] https://github.com/SublimeLinter/SublimeLinter3/issues/671
  - [x] https://github.com/SublimeLinter/SublimeLinter3/issues/675
  - [x] user settings backup fails if it isn't valid json
  - [x] sporadic empy inline popups (related to zero length error regions?)
  - [x] icon options "none", "dot", "circle", or "bookmark" don't work, which also means there is no way to disable the icons.
- [x] Refactoring:
  - [x] write gutter icon information right to style store instead of persist.gutter_marks as intermediate
  - [x] refactor lint/persist.py and check with linters
- [x] ~~check if end col number is correct~~ fixed
- [x] Code quality review, including:
  - [x] single-use constants
  - [x] stray TODO and other comments
  - [x] remove `force_xml_scheme` option
  - [x] something is printing `[]`
  - [x] global search for unused code (via _vulture_)
- [x] lint on load/startup in background mode
- [x] Error panel
  - [x] ~~show only start col?~~ yes
  - [x] panel does not show if console (other panels?) are opened (LSP doesn't have this problem)
  - [x] cleanup panel commands
  - [x] update panel content on lint
  - [x] provide options via keybinding
  - [x] fix filtering by provided options
  - [x] remove `select` option
  - [x] make panel toggleable
  - [x] decide on view/panel line highlight synchronisation
  - [x] implement region addition for styling
  - [x] check overall styling, modify if necessary
  - [x] show only errors for the current window, current file first
  - [x] fix panel bugs:
    - [x] bug related to `base_dir` finder
    - [x] shouldn't list files that don't have errors
- [x] Commands and Keybindings
  - [x] improve command name
    - [x] ~~should internal commands be named by a different scheme, for encapsulation?~~ not necessary 
  - [x] review keybindings
    - [x] ~~eventually unified key binding file for all platforms?~~ can't be done and there shouldn't be many bindings anyway
- [x] backup old user settings on upgrade
- [x] update docs

- [x] next/previous error crashes ST (3154)
- [x] In settings: styles -> After entering a scope that isn't in the current color scheme (e.g. `bla`), it will complain about every next change even if you enter a valid scope. 
- [x] After every change, all highlights are removed and then restored, making them "blink" under certain circumstances.
- [x] singular/plural in-line report 
- [x] check GitGutter gutter hover interop
- [x] changing gutter themes seems to be broken
- [x] all default settings are copied to the user settings
- [x] check if disabling linters still works
- [x] can not focus panel
- [x] error when user settings file doesn't exist (thanks @darkred for finding this!)

Some core issues we want to register:

- [x] ~~inconsistent type when set vs get to settings~~ 

Feature requests:

- [x] Tooltip on hover over a highlight, displaying the errors for that specific position (which may be multiple, but it shouldn't show all errors for the line)
- [x] It should be possible to disable each popup.
- [x] Replace `Report (Open Files)` and `Show All Errors` with an output panel, like the one in [LSP](https://packagecontrol.io/packages/LSP). Will fix #115 and #212
 
We could consider dropping the new line report popup in favour of these two features, but it's ok to first try and see how it feels and then decide. I think reporting per line isn't all that natural, but perhaps that depends on the linters you're using.


jrappen:
@groteworld Looking at eda3a8a1fa there is still a Gratipay reference in the README at the bottom. Also you can register a `paypal.me` URL (and use it like `paypal.me/groteworld/5usd`), which I personally find to be easier to use. If you want to keep the recurring option for PayPal though, you'll need the button (if they didn't change that meanwhile for the `paypal.me` site, didn't check).

braver:
@jrappen not sure this PR is the place for this discussion, but I'll remove it from the README too.

jrappen:
The changes here do not address `*.hidden-color-scheme` files, only `*.sublime-color-scheme` files. Compare the last paragraph from Jon's message here: <https://forum.sublimetext.com/t/dev-build-3154/33286>.

braver:
I'm not sure what you mean. > 3148 We don't create color schemes at all anymore, instead we fully rely on scopes already available in a given color scheme (or the region.colorish scopes). < 3148 hidden tmTheme files are created.

FichteFoll:
Glad to see changes happening to SublimeLinter. This opportunity should probably be used to solve the awkward situation of Python linters being evaluated within the built-in environment, which causes various issues and is generally quite hack with any version that is not 3.3. I don't think there is an issue for this yet, but it's been a while since I last looked for one. This has been on my list for a while now, but I won't be able to file a proper issue until Sunday.

FichteFoll:
I can imagine that help is always welcome, but I'm unable to provide it here currently. There are just too many things I'm dingbat already.

FichteFoll:
*doing. Thanks autocorrect and thanks github for making editing comments on mobile a pain.

braver:
Yes, dingbat is most likely to be the word you were looking for 😀 

Just taking a peek at it and having some feedback is also good!

gerardroche:
Seems to ok for me with just one issue:

Please add more granularity to the default style scopes. Specifically a namespace, but also add types to the scopes.

The reason is to allow color schemes to enhance support specifically for the plugin, because while a color scheme can define styles for the colorish regions it may want to do something specific for a plugin.

I've already talked elsewhere on this so I'm not going to rehash it all. I recommend postfixing a namespace so that color schemes can do this:
 
```
    <dict>
        <key>scope</key>
        <string>region.yellowish sublimelinter</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>#fd971fa6</string>
        </dict>
    </dict>
```

Appending it may cause potential issues later. Postfixing is more powerful.

Also add types but I suggest not coupling it to the namespace for the most flexibility:

```
    <dict>
        <key>scope</key>
        <string>region.yellowish warning.sublimelinter sublimelinter</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>#fd971fa6</string>
        </dict>
    </dict>
```

One other thing that would be nice is to differentiate between debug and verbose for plugin debug messages and lint warnings. For example when you enable debug mode you get messages like:

```
stdin:1:1: F401 'sublime' imported but unused
stdin:3:1: E302 expected 2 blank lines, found 1
stdin:7:1: F821 undefined name 'f'
stdin:8:1: W391 blank line at end of file 
```

I often want to see those messages because I want to lookup the error code (there's probably several features in there). Enabling debug mode gives too much noise really all you want is verbose linting information.

FichteFoll:
Error codes are within the domain of plugins. SublimeLinter-flake8 has an option for this, for example (though imo it should be enabled by default).

Regarding more specific scopes, what about prefixing the sublimelinter scope?

braver:
You’ll be able to view a list of errors per line but hovering over the gutter, and there are report options already in place. The console is not the place for such features, so we definitely won’t be enhancing the logging options there.

I need to check how it works, but if I’m right we’re already baking in a scope that will allow you to target the highlights from a color scheme. I don’t think you want users to have to “fight” against a color scheme just to get the colors they want. Region.colorish is already in hands of the color scheme, as are things like markup.error. A color scheme can advertise to the user which scopes are suitable for linter highlights, but we’re explicitly putting this in the hands of users. This will allow them to differentiate between different linters in a single file, which is one of the most frequently recurring feature requests. A color scheme cannot reasonably provision for such situations.

messages/4.0.0-rc.1.txt
```diff
@@ -0,0 +1,64 @@
...
+
+
+Changes to commands and menus
+-------------------------------
+
+The context menu now only list commands relevant for the view context.
+The tools menu now only lists the clear_cache command.
+Commands are now named sublime_linter_* to comply with naming conventions.
+
+Commands to change settings or toggle linters have been removed.
```
FichteFoll: What about the situation when you want to have a linter disabled by default but enable it for certain projects only?

braver: Sure, nothing about the enabled/disabled settings has changed, just the command has been removed. You can enable linters in project settings or in sublimelinterrc files.

FichteFoll: Right, didn't read this correctly. All good then.

pykong:
**Teaser...**
custom_highlights

darkred:
Sorry for bothering you, but I've been trying to test this but without success. 
**Could you please tell me what I'm doing wrong?**

STR:
- Using ST3 3154 x64 (clean installation).
- Download https://github.com/SublimeLinter/SublimeLinter3/archive/next.zip
- inside the resulting `SublimeLinter3-next.zip` there's a folder `SublimeLinter3-next` with the content. 
So, I copy it as `SublimeLinter` into `C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages`
- Then, as I run ST3 3154 x64 (clean installation) I get 

```
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\Kostas\\AppData\\Roaming\\Sublime Text 3\\Packages\\User\\SublimeLinter.sublime-settings'`
```

- I open: Preferences | Package Settings | SublimeLinter  -> Settings
and I copy-paste the settings from the left (Default settings) to the right (USer).
Then, as I restart ST, I get this:
    ```
      File "C:\Program Files\Sublime Text 3\sublime_plugin.py", line 210, in on_api_ready
        m.plugin_loaded()
      File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\sublime_linter.py", line 35, in plugin_loaded
        @legacy_check
      File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\legacy.py", line 210, in legacy_check
        update_settings(min_version, usr_dir_abs)
      File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\legacy.py", line 189, in update_settings
        settings = json.load(rf)
      File "./python3.3/json/__init__.py", line 271, in load
      File "./python3.3/json/__init__.py", line 316, in loads
      File "./python3.3/json/decoder.py", line 351, in decode
      File "./python3.3/json/decoder.py", line 367, in raw_decode
    ValueError: Expecting property name enclosed in double quotes: line 4 column 5 (char 27)
    ```

- _In the end, I'd like to try it with `Sublime​Linter-contrib-eslint`, if possible_.

pykong:
@darkred Thanks for your interest in the next SublimeLinter.

The current commit is not working, as it serves as a backup push.
Much is WIP as of now, albeit maturing each day.
Try the commit before: 667acf8

Your error log implies a problem with the settings conversion.
I need both your default and user version of `SublimeLinter.sublime-settings`, to have a look into this.
Alternatively, try to truncate your user settings to an empty object `{}`.

@braver What is a good place to track issues/bug reports on the upcoming release?
If we collect them in this single issue it will become messy pretty soon.

gerardroche:
> Regarding more specific scopes, what about prefixing the sublimelinter scope? @FichteFoll 

You mean instead of `region.yellowish sublimelinter` do `sublimelinter region.yellowish`.

That doesn't work. The reason it doesn't work is because the scopes at the end are more specific and whatever is defined last wins.

It's easier to understand by looking at what problem is being solved.

Let's say plugin sets a `region.redish` scope for errors. By default ST will use the closes approximation of red in the currently selected color scheme. So far so good. All color schemes are supported using a heuristic which makes things look better by default.  

The color scheme wants to provide specific styles for the `region.redish`. It doesn't matter why. Perhaps the color scheme uses `#FF0000` and that means `region.redish` will use that color, but the color scheme wants to set a lighter shade of red for errors. So it defines a style for the redish region:

```
    <dict>
        <key>scope</key>
        <string>region.redish</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>#F900000</string>
        </dict>
    </dict>
```

Great. The color scheme has progressively enhanced the redish region colors. But that style is now set for *all* redish regions. That means all plugins that use the redish regions get that style. That's not bad. That's good. That's what the color scheme wants. However, it wants to a slightly diferent style for sublimelinter. Prefixing won't work i.e. `sublimelinter region.redish` won't override the previous style. We need a postfix:
 
```
    <dict>
        <key>scope</key>
        <string>region.redish sublimelinter</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>#FF00000</string>
        </dict>
    </dict>
```

We might not be done yet. The above style is great. It specifically targets the sublimelinter plugin redish regions. But SL might use redish region for several different elements. Let's imagine a plugin used redish regions for both warnings and errors (as a simple easy to understand example).

 
```
    <dict>
        <key>scope</key>
        <string>region.redish warning sublimelinter</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>#F500000</string>
        </dict>
    </dict>
    <dict>
        <key>scope</key>
        <string>region.redish error sublimelinter</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>#F700000</string>
        </dict>
    </dict>
```

The color scheme can now control all the elements.

> Region.colorish is already in hands of the color scheme, as are things like markup.error. A color scheme can advertise to the user which scopes are suitable for linter highlights, but we’re explicitly putting this in the hands of users.

I'm disappointed. Putting it in the hands of the users is useless. Allowing schemes to progressively enhance the colors of the scheme doesn't take anything away from the user. They still have full control. Allowing color schemes control (without taking it away from the user) enhances users experience without having to ask them to do anything.

darkred:
@pykong Thank you for the help:  with the latest functional commit and an empty settings `{}` it loads ok, no problems. 
It's just that sublimelinter-contrib-eslint and sublimelinter-jshint that I tried, both cause an exception in the console, and so they don't work ).

So I'll just wait for the next release 👍

pykong:
@darkred You are welcome. What type of exceptions do `sublimelinter-contrib-eslint` and` sublimelinter-jshint` cause? State the error logs, so I can gauge whether those errors have anything to do with any changes I made to **SL**.

darkred:
@pykong Here: (while opening a .js file) both cause the same exception:

```
SublimeLinter: error in SublimeLinter daemon: 
SublimeLinter: -------------------- 
SublimeLinter: Traceback (most recent call last):
  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\queue.py", line 57, in loop
    item = self.q.get(block=True, timeout=self.MIN_DELAY)
  File "./python3.3/queue.py", line 175, in get
queue.Empty

During handling of the above exception, another exception occurred:

  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\queue.py", line 65, in loop
    self.lint(view_id, timestamp)
  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\queue.py", line 111, in lint
  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\sublime_linter.py", line 263, in lint
  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\linter.py", line 1008, in lint_view
    callback(cls.get_view(vid), linters, hit_time)
  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\sublime_linter.py", line 316, in highlight
    highlights.draw(other_view)
  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\highlight.py", line 133, in draw
    all.draw(view)
  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\highlight.py", line 477, in draw
    icon = self.get_style("icon", style, error_type)
  File "C:\Users\Kostas\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\highlight.py", line 184, in wrapper
    return persist.gutter_marks["icons"][error_type]
KeyError: 'warning'
 
SublimeLinter: -------------------- 
```

FichteFoll:
@gerardroche The color scheme could define a color for the selector `sublimelinter region.redish` just fine, afaik. I didn't test it, but I believe it should work just like the other way around. If it doesn't, it's probably a bug.

Either way, I believe that the optimal scope names are `region.redish markup.error.sublimelinter` and `region.yellowish markup.warning.sublimelinte` via https://github.com/sublimehq/Packages/issues/1036, where the regionish scopes server as a default fallback and targetted colors could be defined for `markup.warning` (or `markup.warning.sublimelinter` even).
SL could probably be using the respective error type for this scope directly, so that people could get `markup.info.sublimelinter` for the "info" type. As for the regionish fallback color, yellowish seems appropriate.

gerardroche:
> The color scheme could define a color for the selector `sublimelinter region.redish` just fine, afaik. I didn't test it, but I believe it should work just like the other way around. If it doesn't, it's probably a bug.

`sublimelinter region.redish` will work, but it won't override a style `region.redish` because `region.redish` is more specific.  An example, the intention below is that `sublimelinter region.redish` should override the `region.redish`, but it doesn't. Maybe it is a bug. I've tested various ways of solving these issues and the postfix is most consistent, most powerful, and flexible.

```
    <dict>
        <key>scope</key>
        <string>region.redish</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>red</string>
        </dict>
    </dict>
    <dict>
        <key>scope</key>
        <string>sublimelinter region.redish</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>blue</string>
        </dict>
    </dict>
```

> Either way, I believe that the optimal scope names are `region.redish markup.error.sublimelinter` and `region.yellowish markup.warning.sublimelinter` via sublimehq/Packages#1036, where the regionish scopes serve as a default fallback and targetted colors could be defined for `markup.warning` (or `markup.warning.sublimelinter` even).

Is the intention of falling back to the `markup` to support older color schemes?

If you use the scope: `region.yellowish markup.warning.sublimelinter` then the styles below won't work i.e. you can't style them seperately. That could be another bug, but I don't think it is:

```
    <dict>
        <key>scope</key>
        <string>region.yellowish</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>yellow</string>
        </dict>
    </dict>
    <dict>
        <key>scope</key>
        <string>markup.warning</string>
        <key>settings</key>
        <dict>
            <key>foreground</key>
            <string>red</string>
        </dict>
    </dict>
```

markup.warning` styled then it will override the `region.yellowish`. You have to target `region.yellowish markup.warning` in the first style.

I also don't like the appending of the namespace because it means you either then need to add it to every scope or it limits what you can target.

I've tested many of the different options available and the postfix is best option. 

 > SL could probably be using the respective error type for this scope directly, so that people could get markup.info.sublimelinter for the "info" type. As for the regionish fallback color, yellowish seems appropriate.

The point of adding the type or whatever the relevant semantic element name, is so that it can be targeted for customisation.

braver:
@darkred thanks for your enthusiasm with regards to the upcoming release. I do try to keep the `next` branch in a usable state (I use it myself every day), so it's a good place to get an early look. But we shouldn't be that far off from a releasable version. contrib-eslint works just fine on my machine by the way, but there have been a lot of people that have trouble with it (or any other linter really, eslint is just one of the most popular). If you don't have this problem on the regular release of SublimeLinter, could you open a new issue about it so we can track it?

@pykong We'll track issues in the beta just as we would other issues, but I'll label them so we can find them easily. For now, while we're still making big changes let's try to keep it to this PR. 

@FichteFoll 

I agree with https://github.com/sublimehq/Packages/issues/1036 and will add the scopes you propose to the defaults style settings. I tested it and it allows a color scheme to override the colorish scope by simply supplying a color for markup.error/warning, and optionally specifically that for sublimelinter. 

@gerardroche 

> Putting it in the hands of the users is useless

Color schemes select colors for different semantic scopes. These can be plugin specific. Users may run any combination of linters. The defaults for the colors of those linters are in the hands of the color scheme via semantic scopes. This is already the reality of the current state of this PR. Users may then customise this in several ways, one of which is assigning colors to each linter to keep them separated. How they do this is up to the user, it's called customisation for a reason. This is not useless, there is no need to be disappointed.

> Is the intention of falling back to the markup to support older color schemes

No, it's to support new color schemes that adopt the scope names proposed in https://github.com/sublimehq/Packages/issues/1036, region.colorish will handle all backwards compatibility.

> markup.warning will override the region.yellowish style

That's the point.

gerardroche:
> The defaults for the colors of those linters are in the hands of the color scheme via semantic scopes. This is already the reality of the current state of this PR.

This is only a reality since the recent change adding the markup scopes with a namespace https://github.com/SublimeLinter/SublimeLinter3/pull/666/commits/12a4cc154e6ddb70e546f103d5aa451fd9d674cd. Prior to that change you couldn't target sublimelinter elements.

braver:
@gerardroche sorry, I messed something up in editing that post. I was under the impression that there was already a `sublimelinter` pre- or postfix hardcoded. Apparently there wasn't. I made the changes while editing the post, but didn't correctly update that section.

darkred:
> contrib-eslint works just fine on my machine by the way, but there have been a lot of people that have trouble with it (or any other linter really, eslint is just one of the most popular). If you don't have this problem on the regular release of SublimeLinter, could you open a new issue about it so we can track it?

@braver Yes, of course. Here: https://github.com/SublimeLinter/SublimeLinter3/issues/671
btw, could you please share your SublimeLinter settings (User) ?

kaste:
Set reminder: This will break the `flake8` plugin bc of simple import changes, so don't forget to leave a note over there, when this will land.

braver:
I think it needs the clarification of what's going on because it's detached from context and you hopefully don't see it a lot. Perhaps we should make all of it configurable so everyone can make it to their taste, or we should have a shared "busy" indicator like they do in Atom. Things can always be improved but we might not always agree on how to do so.

braver:
Ok, it's clear we won't do anything in SL3 anymore, and if we do we can always branch out again. I'm going to merge this behemoth, it'll make a couple of things easier to manage.

kaste:
😭 I will miss you, branch.

braver:
Yeah, it was a pretty good branch...

👋🏻

PR #670, opened 2017-11-27
Use current shell path if use_current_shell_path is true

douglas:
Right now, I'm working on some Ruby/Rails projects using an in-house developed tool to aid in development environment setup/creation.

Basically, it installs the right version of Ruby, the needed gems and finally it add the project ruby to the PATH before the system binaries dir.

My environment is as follows:

- ruby 2.4.2 in /usr/local/bin with rubocop installed and my default PATH:

```bash
» env | grep PATH
PATH=/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin:/usr/local/heroku/bin:/Users/douglas/work/golang/bin:/Users/douglas/.composer/vendor/bin:/Users/douglas/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/douglas/Library/Android/sdk/platform-tools:/Users/douglas/Library/Android/sdk/tools/bin:/usr/local/opt/fzf/bin
```
- ruby 2.3.5 when I ```cd ~/work/project```:

```bash
» env | grep PATH
PATH=/Users/douglas/.gem/ruby/2.3.5/bin:/opt/rubies/2.3.5/lib/ruby/gems/2.3.0/bin:/opt/rubies/2.3.5/bin:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin:/usr/local/heroku/bin:/Users/douglas/work/golang/bin:/Users/douglas/.composer/vendor/bin:/Users/douglas/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/douglas/Library/Android/sdk/platform-tools:/Users/douglas/Library/Android/sdk/tools/bin:/usr/local/opt/fzf/bin
```

With the current implementation of ```get_shell_path```, when I _cd_  to the ```~/work/project``` directory and open SublimeText using ```» subl .```, I get the first shell PATH (my default shell path) instead of the current session PATH, so, it will not find the rubocop binary installed for the project.

With this PR, it will be possible to tell SublimeLinter to just return the PATH defined in the current session, allowing _SublimeLinter-rubocop_ to find the right executable.

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -526,6 +526,11 @@ def get_shell_path(env):
 
     """
 
+    from . import persist
+
+    if persist.settings.get('use_current_shell_path') is True:
+        return env['PATH']
+
     if 'SHELL' in env:
         shell_path = env['SHELL']
         shell = os.path.basename(shell_path)
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -6,6 +6,7 @@
         "gutter_theme": "Packages/SublimeLinter/gutter-themes/Default/Default.gutter-theme",
         "gutter_theme_excludes": [],
         "lint_mode": "background",
+        "use_current_shell_path": false,
         "mark_style": "outline",
         "no_column_highlights_line": false,
         "passive_warnings": false,


lint/util.py
```diff
@@ -526,6 +526,11 @@ def get_shell_path(env):
 
     """
 
+    from . import persist
+
+    if persist.settings.get('use_current_shell_path') is True:
+        return env['PATH']
+
     if 'SHELL' in env:
```
kaste: Is this branching here useful anymore? Or should we just use env['PATH']?

lint/util.py
```diff
@@ -526,6 +526,11 @@ def get_shell_path(env):
 
     """
 
+    from . import persist
+
+    if persist.settings.get('use_current_shell_path') is True:
+        return env['PATH']
```
kaste: Note that we return env['PATH'] at the end of the function anyway.

kaste:
@braver Honestly, why is that needed? Why not just use `env['PATH']` like we do on Windows anyway?

braver:
A lot of things happen here, before the PR, that I find really questionable. I don’t understand why we wouldn’t just use env[‘PATH’] and I don’t know what all the guesswork that happens here solves. I’ve not been involved that long. This part of the SL3 codebase looks bad to me though.

kaste:
Maybe git blame can help us.

braver:
> Maybe git blame can help us.

Nope, it come down to this very helpful first commit:

> Restore everything

Can’t find anything useful in the old repo, so it seems to be by design. 

In cases like these I tend to go with “let’s take it all out and see what breaks”.

kaste:
Hm, actually e2cbad82425f8312d3c7810b78c6138a004c563c introduced it. I usually open sublime via `subl .`. 

When you just open the console in Sublime, and then `import os; os.environ['PATH']`, do you see a reasonable output. Bc on my machine I get exactly what I get outside of sublime in a shell.

Of course, we should take it out, if it breaks for people. I think this is even Sublime2 code. 🤷‍♂️

braver:
Good find. Would it be because of https://github.com/SublimeTextIssues/Core/issues/1877?

> When you just open the console in Sublime, and then import os; os.environ['PATH'], do you see a reasonable output

I get good output here, but plugins may load before all env vars are loaded. I run into this with LSP sometimes. 

Oh I don't know. It runs just fine on my system without the entirety of `get_shell_path`. Anyone want to give me a good reason to not just remove it?

kaste:
🙌 Yes, that seems to be the reason. But then it's MacOS, not 'posix'. And if you start sublime via `subl .`, it's always correct? How does a corrupt path look like? Can you try the steps to reproduce the issue?

kaste:
Unfortunately we don't have a `Restart servers` command like in LSP. Instead, we have a `lrucache` in front of that. That means, if you unlucky and hit this sublime bug, you have to restart and restart and restart.

FichteFoll:
`lrucache` is also quite unfortunate in case you added/removed an executable to your PATH, as SL currently requires a restarte for that to take effect (for obvious reasons). Might be better suited for a separate issue though.

braver:
> And if you start sublime via subl ., it's always correct?

Yes, it inherits the PATH in that case

> How does a corrupt path look like? Can you try the steps to reproduce the issue?

Yes, the steps in that Core issue work fine. `/usr/bin:/bin:/usr/sbin:/sbin` in my case... pretty useless

With LSP I run into this regularly, but LSP also needs to have a server running in the background and we just run commands and can easily just run them again. I haven't run into this problem yet after taking out `get_shell_path`, but who knows. Maybe there is just enough of a delay baked into SL for it to not happen.

Should we even have a cache here? Seems like an optimisation with enough risks to do without unless we really need to optimise this.

braver:
`get_shell_path` is gone for SL4. Let's just see what happens, and looks a lot more sane than what was in there before. Thanks for reviewing and commenting @kaste and @FichteFoll. 

For SL3, I'm like 🤷‍♂️ I don't mind merging this even if just one guy uses it. Am not going to update docs or anything though.

sebastienbarre:
I was auto-upgraded to SL4 from SL3, and my linter binaries are no longer found. I assume this is because of the changes mentioned in this thread? Is there a workaround? Thanks.

FichteFoll:
Can you give some specific examples and include  a debug log? Or was this already addressed in a different issue?

sebastienbarre:
I use Sublime and SublimeLinter everyday. One day (yesterday) it upgraded from SL3 to SL4 and none of my linters were found. I added `"~/.composer/vendor/bin/"`, which *is* in my PATH, to `paths.osx` to fix it in the SL4 settings, but that's not ideal, and probably burned/will burn other people. 

Here is what SL4 displays in debug mode now:
```
SublimeLinter: SublimeLinter.sublime-package:103: PATH:
    /usr/bin
    /bin
    /usr/sbin
    /sbin
```

@kaste seems to hint that [it auto-fixes itself](https://github.com/SublimeLinter/SublimeLinter/issues/1066#issuecomment-370163506), though I'm not sure what it meant.

FichteFoll:
Okay, let's keep this in #1066 then. Having multiple topics in the same issue is confusing.

sebastienbarre:
Actually it seems this PR is the better place? :)  The other is related to how filenames are passed to linters. @braver seems to say that `get_shell_path` is gone for SL4. Maybe I'm minsterpreting this function... Is it getting the shell's PATH? Or the path to the shell executable?

FichteFoll:
This pr is merged and closed. If you have an issue with something that the poor changed, you should create a new one that can be handled separately with a reference to the relevant change.

PR #672, opened 2017-12-01
Fix issue with unassigned variable

EmmaSimon:
I was getting the following error, and it was preventing some files from being linted.
```
  File "/opt/sublime_text/sublime_plugin.py", line 538, in on_activated_async
  File "/home/emma/.config/sublime-text-3/Packages/SublimeLinter/sublimelinter.py", line 280, in on_activated_async
    self.check_syntax(view)
  File "/home/emma/.config/sublime-text-3/Packages/SublimeLinter/sublimelinter.py", line 207, in check_syntax
    Linter.assign(view, reset=True)
  File "/home/emma/.config/sublime-text-3/Packages/SublimeLinter/lint/linter.py", line 773, in assign
    if not linter_class.disabled and linter_class.can_lint(syntax):
  File "./python3.3/functools.py", line 251, in wrapper
  File "/home/emma/.config/sublime-text-3/Packages/SublimeLinter/lint/linter.py", line 1641, in can_lint
    status = 'WARNING: {} deactivated, cannot locate \'{}\''.format(cls.name, executable)
UnboundLocalError: local variable 'executable' referenced before assignment
```
Assigning `executable` before the if seems to fix things.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1590,9 +1590,9 @@ class Linter(metaclass=LinterMeta):
                 can = cls.syntax.match(syntax) is not None
 
         if can:
-            if cls.executable_path is None:
-                executable = ''
+            executable = ''
 
+            if cls.executable_path is None:
                 if not callable(cls.cmd):
                     if isinstance(cls.cmd, (tuple, list)):
                         executable = (cls.cmd or [''])[0]


braver:
Isn't this the same problem as #491?

EmmaSimon:
Ah, it seems so. I tried to find it it had already been done but didn't notice that one. Is there something preventing the merge of the other fix?

braver:
@EmmaSimon Well, we've been a bit short on maintainers for one. Also, I'd like to test and verify a solution, but I don't know how to reproduce this. However, now that you are, perhaps we can get somewhere. #491 seems like a cleaner solution to me, can you verify if it solves the problem for you too?

EmmaSimon:
Just for reference, it looks like it's happening to [jshint](https://github.com/SublimeLinter/SublimeLinter-jshint), with the following `user.linters` dict,
```
"jshint": {
    "@disable": true,
    "args": [
      "--config",
      "/home/emma/.config/jshint/.jshintrc"
    ],
    "excludes": []
}
```
`cls.syntax = ('javascript', 'html', 'javascriptnext', 'javascript (babel)`
`cls.executable_path = ''`
jshint is not installed on the system.

#491 works perfectly, it prints out `SublimeLinter: WARNING: jshint deactivated, cannot locate ''`

Issue #466, opened 2016-06-06
linux zsh util.py


victorzhuk:
line 537, calc PATH variable. with archlinux x64 on zsh returns empty, and throw error. with bash all ok.

victorzhuk:
zsh and bash are openning with '-l' (login) key, thats why if you are executing commands in .zprofile or .bash_profile, SublimeLinter execute it too. Is it so necessary call shells with login key?

braver:
Not a single clue as to what you're trying to do and what does or doesn't work.... sorry.

victorzhuk:
You can read [this](https://unix.stackexchange.com/a/46856) first of all. And then look at [this code](https://github.com/SublimeLinter/SublimeLinter3/blob/20a5dde78c1f7cbdbe73dcdebca3f554d88efc68/lint/util.py#L537). I`m (and maybe someone else) using i3wm as my DE for my linux. So some scripts, like set desktop wallpapers, I place into .zprofile file at home dir. When you call shell with login param, it auto execute this script, and if have some error during execution getting your path fail.

braver:
Well, perhaps you should fix or move some of those scripts then, because reading the profile when a shell is started is pretty important for things like PATH configuration.

victorzhuk:
I don`t think than it is a good practice to depend on user shell path. Maybe it`s better to move user paths to sublime config?

braver:
It is a good practice to use the PATH if it is set and there are already several options to augment it.

PR #676, opened 2017-12-02
#466 fix, remove "-l" arg from shell call

victorzhuk:


braver:
That’s not really a good idea, we need to respect the profile.

PR #677, opened 2017-12-03
Fixes and enhancements

gerardroche:
* Fix AttributeError: 'dict' object has no attribute 'set'

  This error causes migrating old settings to fail.

* Skip migrating settings that are the same as the default values.

  This results in less noise in the user settings file.

It looks like there's other unused settings that could be removed, but I don't know enough about the current refactoring to know if they can be removed safely.


braver:
@pykong it looks good to me, but want to check with you if it intersects with what you're working on right now before I merge it.

pykong:
@gerardroche Thank you very much for your interest and contribution into SublimeLinter 4.
We have decided to not include the settings migrator.

PR #701, opened 2017-12-16
Match entire result panel line

FichteFoll:
Causes the entire line to be highlighted, which helps with locating the
currently viewed error report in the panel.

Also strip whitespaces (notably trailing newlines) since they are
useless.


braver:
Sweet! Thanks!

PR #702, opened 2017-12-16
Print source pattern of failed regex match

FichteFoll:
The string representation of a compiled regex object isn't that exciting.

```py
>>> import re
>>> re.compile("abc")
<_sre.SRE_Pattern object at 0x7f694c33f630>
>>> _.pattern
'abc'
```


braver:
Thanks!

PR #704, opened 2017-12-16
Fix: Panel flickers when typing *anywhere*

deathaxe:
This commit introduces the "sublime_linter_replace_all" command to replace the whole content of a view with only one TextCommand to avoid repainting the view several times while updateing its whole content.


braver:
fixed via #705

PR #705, opened 2017-12-16
Update panel in a single TextCommand

### LINKED ISSUE #700: [next] Panel flickers when typing *anywhere*
When typing something, the errors in the panel flicker, presumably when the view is re-linted.

Now, the interesting part is that this also happens when typing in the **Command Palette**. I've verified (in the console, with debug prints) that no linting happens when I type into the command palette panel.

I recorded a video of this (does not play in FireFox). You can see the flicker happening at ~0:14. It happened a few additional times but those weren't picked up in the video. It's also inconsistent.

https://x0.at/s3c.mp4

FichteFoll:
Prevents panel from flickering due to multiple commands updatating its
contents (fixes #700).

Also try to restore the previous selection to allow the user to continue
browsing errors exactly were they left off. Reset to the beginning
otherwise.

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -1,3 +1,4 @@
+import sublime
 import sublime_plugin
 
 from .lint import persist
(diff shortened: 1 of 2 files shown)


braver:
This would replace #704 then I presume?

FichteFoll:
Haha, looks like I wasn't fast enough.

braver:
the presumption was pretty strong ;) Thanks for this!

PR #712, opened 2017-12-17
Panel path display and result_*_regex fixup

FichteFoll:
See commit messages for details.


braver:
Something I noticed while testing this:

lints

It already had this, where linting an untitled view would somehow update the paths even though the untitled view has no errors. Now that you're touching this perhaps you have an idea why that would happen. Or @pykong has?

Looks super good though, thanks.

FichteFoll:
Fixed that as well.

PR #713, opened 2017-12-18
Add NewPythonLinter

kaste:
NewPythonLinter is taken from flake8 and was tested in public for about a month.

### DIFF
--- a/lint/__init__.py
+++ b/lint/__init__.py
@@ -20,6 +20,7 @@ from . import (
 
 from .linter import Linter
 from .python_linter import PythonLinter
+from .new_python_linter import NewPythonLinter
 from .ruby_linter import RubyLinter
 from .node_linter import NodeLinter
 from .composer_linter import ComposerLinter
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -60,7 +60,7 @@ class LinterMeta(type):
         if bases:
             setattr(cls, 'disabled', False)
 
-            if name in ('PythonLinter', 'RubyLinter', 'NodeLinter', 'ComposerLinter'):
+            if name in ('PythonLinter', 'NewPythonLinter', 'RubyLinter', 'NodeLinter', 'ComposerLinter'):
                 return
 
             cls.alt_name = cls.make_alt_name(name)
--- a/lint/new_python_linter.py
+++ b/lint/new_python_linter.py
@@ -1,0 +1,302 @@
+from functools import lru_cache
+import os
+import subprocess
+
+import sublime
+from . import linter, persist, util
+
+
+class NewPythonLinter(linter.Linter):
+    """New Python Linter [WIP].
+
+    """
+
+    @classmethod
+    @lru_cache(maxsize=None)
+    def can_lint(cls, syntax):
+        """Determine optimistically if the linter can handle the provided syntax."""
+
+        can = False
+        syntax = syntax.lower()
+
+        if cls.syntax:
+            if isinstance(cls.syntax, (tuple, list)):
+                can = syntax in cls.syntax
+            elif cls.syntax == '*':
+                can = True
+            elif isinstance(cls.syntax, str):
+                can = syntax == cls.syntax
+            else:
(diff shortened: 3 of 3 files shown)


braver:
@kaste It might be a bit more work, but would it be possible to base this on the `next` branch instead of master? I'd like to not touch SL3 if we can help it.

kaste:
Okay, if we want this as a breaking change for SL4, we can replace the old PythonLinter with the new one herein. The old PythonLinter only adds linting using the builtin python (the python33 nobody wants).

And it adds a bug where you cannot have `self.cmd` as a callable anymore. Reason:

https://github.com/SublimeLinter/SublimeLinter3/blob/2d6667f41f4f81e214021bb41ea7ffcdc016dfce/lint/base_linter/python_linter.py#L292 

within `run` but `run` already has the signature `def run(self, cmd, code)` and the correct `cmd`. I think this is why @FichteFoll had to patch `build_cmd` (here: https://github.com/fredcallaway/SublimeLinter-contrib-mypy/commit/e36dd92e0ba6e6c72cfcf0f75d96ee94b5783f50#diff-6c1f4f1574c85da37dc4dd59341c9a39R144) for `mypy`.


We _just_ need to find the correct entry point to remove a given `@python` intent from `cmd[0]` bc the new python linter does not support that. (I mean it is already a PythonLinter so we don't need that intent anymore.)

How many linters are out there that still want the builtin python? Is this still too risky?

kaste:
Haha, seems rather like flake8 is the only linter which is NOT using the internal python.

kaste:
`next` doesn't work that well. Don't know if I can do this then. Bc I have to stay productive as well.

FichteFoll:
How many plugins do we have currently subclassing PythonLinter and which of these do not run internally?

I agree that this should go into SL4. With all the work already put into it, SL3 can already be considered deprecated.

kaste:
A well I'm a bit pissed off. It couldn't convert my settings. I had a lot of import errors. `eslint` had import errors. I had to restart sublime a dozen times. All my view plugins are broken.

kaste:
Okay, it doesn't seem to reload the settings after edit. Should it?

Whats worse, it doesn't reload the source code. If I edit say new_python_linter or `highlight` it does not reload that file. So, this is unworkable.

braver:
Well that's odd, eslint for one runs just fine on my machine. It might bork out on some linters but none that I tried and it shouldn't be that bad. 

It doesn't upgrade your settings, it backs them up. What do your settings look like right now?

Module reloading (e.g. everything in the linter dir) doesn't work in Sublime, I believe you can set it up with Package Reloader or something to do that. I don't run into this often enough to have set that up yet though (I just `CMD+Q` and restart).

I'm sorry if it's killing your work flow right now.

braver:
> How many plugins do we have currently subclassing PythonLinter

I haven't run any analytics but it's about 7 that are in the SL org,m plus mypy.

> and which of these do not run internally?

What do you mean with "internally"?

FichteFoll:
Internally as in "inside the Python 3.3 env that ST provides". I should revise that I meant to ask for which plugins *do* that, not those that don't.

Regarding the PR at hand, I'm confident that this is an improvement (I didn't compare to the current PythonLinter class), but I think we can do even better. Eventually.

kaste:
Ok, it's running. 👍   Package Reloader is the king 👑 

But the transition should be way smoother. It should migrate the settings if possible.

braver:
> Ok, it's running

Pfew, good to hear 😅 

> It should migrate the settings if possible

If things settle a bit I'm not against it, but we had trouble keeping it working reliably so we ditched it for now. Another problem is lot of people have bogus settings because the old version of the plugin wrote *everything* to the user settings. So we need a pretty smart migrator and we need someone to write it and test the hell out of it. Really open for a PR here. Right now though, the issue at hand is where I'd like to spend energy because it helps more people.

braver:
@kaste shall we close this one then?

PR #714, opened 2017-12-18
Trivial adjustments to result_line_regex and panel template

FichteFoll:


pykong:
Those changes make sense. Thanks!

PR #718, opened 2017-12-19
Diagnostics panel improvements/suggestions

deathaxe:
This commit enables the gutter area of the output panel and indents each line by a single `tab` in order to enable ST´s code folding capabilities. This way a file which is out of interest can be collapsed. In order to ensure correct indention, the line:col pair is moved to the second column and the severity to the first one and displayed as unicode character. As a result the empty line between each file seems no longer required and is therefore removed. To ensure code folding works properly the very last line must end with a `\n`.

### DIFF
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -9,15 +9,25 @@ OUTPUT_PANEL_SETTINGS = {
     "auto_indent": False,
     "draw_indent_guides": False,
     "draw_white_space": "None",
-    "gutter": False,
+    "gutter": True,
     "is_widget": True,
     "line_numbers": False,
     "margin": 3,
     "match_brackets": False,
+    # path/file.ext:
+    "result_file_regex": r"^(.*):$",
+    # type  row:col  linter  code  message
+    # <col> is optional
+    "result_line_regex": r"^\s+[⨻⨂]\s+(\d+)(?::(\d+))?\s+.*$",
     "scroll_past_end": False,
-    "tab_size": 4,
+    "tab_size": 2,
     "translate_tabs_to_spaces": False,
     "word_wrap": False
+}
+
+ICONS = {
+    "warning": "⨻",
+    "error": "⨂"
 }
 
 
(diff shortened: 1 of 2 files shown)


braver:
Thanks! The folding looks like it's super useful.

Some feedback:
- I'd rather not use the unicode characters for Error/Warning because they're rendered too small in most cases. They end up being just a color blob. I could go with `E` and `W` as we do in the statusbar, or simply full `error` and `warning`.
- Why are the line numbers in italics? I modeled the styling after the find results panel and that doesn't use italics either.

panel/panel.py
```diff
@@ -9,17 +9,27 @@
...
     "margin": 3,
     "match_brackets": False,
+    # path/file.ext:
+    "result_file_regex": r"^(.*):$",
+    # type  row:col  linter  code  message
+    # <col> is optional
+    "result_line_regex": r"^\s+[⨻⨂]\s+(\d+)(?::(\d+))?\s+.*$",
     "scroll_past_end": False,
-    "tab_size": 4,
+    "tab_size": 2,
```
braver: Does this even matter? I'm a 4-spaces guy usually..

deathaxe:
It controls how much the error/warning lines are indented. I like 4 spaces in code as well, but in the output panel it just felt too much.

braver:
> Everyone cries for icons, but no one wants to implement them. :-D

Yeah, graphics are super limited in ST, but I kinda like the bare-bones-ness of it.

braver:
Do you want to update this PR or shall I just go an pick what I like from it?

deathaxe:
The main intention of this PR was to suggest folding. Maybe this feature requires some more work in the future to work properly with updating messages. 

You can pick what you want. It is more or less a set of suggestions. Up to you what you like to use.

FichteFoll:
I liked the long-ish `warning` and `error` type indicators. They are really prominent with background color.

img

Folding seems like a pretty good idea, but functionality to persist the folding state across refreshes of the panel should be added. This is possible by inspecting the folded regions of a view and determining the filename that the region was folded at. Then just reapply the fold if the same file occurs in the new contents again.

braver:
> but functionality to persist the folding state across refreshes of the panel should be added

Oh... you're right, would be sure annoying otherwise. I'm not sure I'm willing to really build out the functionality here right now, I feel like we should do a bit better with regards to internal logic first. But let's keep this open to revisit later.

panel/panel.py
pykong: Me, too. Vouch for 4-spaces.

braver:
Ok, I'm keeping the folding in mind and perhaps revisit this later. For now, it's not going anywhere. Thanks a lot though, I really appreciate the input.

PR #723, opened 2017-12-19
Replace old python linter

kaste:
Instead of #713 

#713 was a non-breaking change bc it only added. It was asked to do a breaking change instead based on the `next` branch.

The PR replaces the old PythonLinter with the new one used by flake8. 

❗️  It removes the ability to use the internal python for linting. So basically all ol'school plugins should be broken. ❗️ The last save commit is https://github.com/SublimeLinter/SublimeLinter3/pull/723/commits/141cdc376eedc21d2db3d3bba62ad9049560c154 bc after that the removal process begins.

- For compatibility, a cmd = ['flake8@python', ...] is allowed, but it is not needed, and in fact we have to strip it away. (https://github.com/SublimeLinter/SublimeLinter3/pull/723/commits/8de77789947b04541cda4c16bc238bde68d475ed)


braver:
Wooo, sweet. I'll have to ponder how we're going to upgrade everything, but definitely looking forward to working with this.

kaste:
ATT: The execution path for `cmd` is now always the same, regardless if you set it statically to a `str` or `tuple` or you generate it dynamically (lazy) via a method. 

This basically means that `context_sensitive_executable_path` will always be called, and thus you don't have to fill in e.g. `self.executable_path` in cmd[0] (Example: https://github.com/fredcallaway/SublimeLinter-contrib-mypy/blob/cd6fbaf0cd9c8f52b23df621537fa9276a92bae1/linter.py#L81-L89) 

(MyPy will BTW either be lucky or break bc it overrides an internal linter method. https://github.com/fredcallaway/SublimeLinter-contrib-mypy/blob/cd6fbaf0cd9c8f52b23df621537fa9276a92bae1/linter.py#L144)

/cc @FichteFoll

kaste:
The most important changes https://github.com/SublimeLinter/SublimeLinter3/pull/723/files?diff=split#diff-8120b617e4c6c91f43e1aec1ba391da7R1025-R1055 (Must use split diff, uni-diff is too confusing.)

* We never call `insert_args` from `get_cmd`
* We normalize `cmd` **within** `get_cmd`
* We **pass** `cmd` to `build_cmd`
*

kaste:
:bug: E.g. https://github.com/SublimeLinter/SublimeLinter-annotations uses the builtin python for good reasons.

FichteFoll:
This is honestly too much to review for me right now, but I'm glad that the Python-specific executable locating stuff has been moved out of util into the python linter class. I believe that the general tools should remain in util though.

braver:
I think we'll just have to test against a bunch of setups, i.e. run it through its paces via the beta, and go from there. I already feel better now that there is *someone* who seems to know what all of this does 😄

lint/base_linter/python_linter.py
```diff
@@ -1,314 +1,330 @@
...
-        that version, the module will be used.
-
-      - Otherwise the executable will be used with the python specified
-        in the "@python" setting, the cmd attribute, or the default system
-        python.
-
-    """
-
-    SHEBANG_RE = re.compile(r'\s*#!(?:(?:/[^/]+)*[/ ])?python(?P<version>\d(?:\.\d)?)')
+    """(New) Python Linter [WIP]."""
```
braver: Is it still WIP or are you ready for this to get merged? It seems to work fine on my machine.... We can always make fixes and adjustments later, but I'd like to expose it to a slightly wider audience.

lint/base_linter/python_linter.py
```diff
@@ -1,314 +1,330 @@
...
+            cmd, env=env, startupinfo=info, universal_newlines=True
+        )
+    except Exception as err:
+        persist.debug(
+            "executing {} failed: reason: {}".format(cmd, str(err))
+        )
+        return ''
 
-        Return a tuple of (have_path, path).
+###
```
braver: Some stray comment leftover?

lint/base_linter/python_linter.py
kaste: Done

lint/base_linter/python_linter.py
kaste: Done

Issue #711, opened 2017-12-17
[beta] Define a namedtuple type to be returned from the split_match plugin API


FichteFoll:
The current interface expects a 7-tuple, which is quite messy. Defining a namedtuple class (with default values) somewhere in the linter module that plugins can instantiate with named kwargs would improve readability significantly.

Additionally, this change would be completely backwards-compatible since you can index namedtuples just like normal tuples in addition to attribute names. (To always work with those internally, a normal tuple could be converted internally using the `*` unpacking operator.)

pykong:
I have opened a PR for the implementation of this enhancement request.
https://github.com/SublimeLinter/SublimeLinter3/pull/732

Maybe you want to have a look into it.

Issue #709, opened 2017-12-17
[beta] Show some indicator whether lint processes are currently active


FichteFoll:
In the panel or status bar, but panel seems to be ideal. Some linters take a little longer to run (e.g. mypy) and it would be useful to know whether there are still some that haven't finished yet instead of assuming there are no more issues with your code.

pykong:
We could borrow the busy indicator from PackageControl.
I think it is a good one.

FichteFoll:
Might make sense to show it in the panel directly, if it is open. Maybe inside a phantom, which doesn't require updating contents of the entire panel.

I'm not sure whether you need to see that lintings is still in progress when that is closed, to be honest.

PR #732, opened 2017-12-23
Implemented namedtuple, lint indicator and some minor cleanup.

pykong:
This PR addresses: 
#709  
#711 
#736


braver:
@pykong can you pull `next` into your branch and resolve the merge conflicts?

pykong:
Is mergeable now.

lint/linter.py
```diff
@@ -15,6 +16,18 @@
...
+    "match": None,
+    "line": None,
+    "col": None,
+    "error": None,
+    "warning": None,
+    "message": '',
+    "near": None
+})
+
+LintMatch = namedtuple("LintMatch", MATCH_DICT.keys())
```
FichteFoll: You can provide default arguments by assigning a tuple to `LintMatch.__new__.__defaults__`, i.e. 

```py
LintMatch = namedtuple("LintMatch", MATCH_DICT.keys())
LintMatch.__new__.__defaults__ = tuple(MATCH_DICT.values())
```

lint/linter.py
```diff
@@ -1628,27 +1651,28 @@ def split_match(self, match):
...
                     col = int(col) - self.line_col_base[1]
                 else:
                     col = len(col)
+                match_dict["col"] = col
 
-            return match, line, col, error, warning, message, near
-        else:
-            persist.debug('No match for regex: {}'.format(self.regex.pattern))
-            return match, None, None, None, None, '', None
+        return LintMatch(*match_dict.values())
```
FichteFoll: Could also use `LintMatch(**match_dict)` here. It's an ordered dict anyway, but this seems more explicit to me (despite having fewer characters).

pykong: Thanks. Didn't know that one!

lint/linter.py
pykong: Thanks. Didn't know that one!

lint/linter.py
pykong: Agreed.

pykong:
Closing this PR until branch is merge ready again, as in the meantime I have added further changes.

pykong:
Reopened.

pykong:
Closed. Again, as we decided to move jsonschema into a sperate package.

pykong:
Reopening after removing jsonschema.

braver:
Lol, yeah, this looks kinda messy. I prefer it this way though, because it gives me a single iteration to test and we don't have to deal with conflicting or overlapping changes. Once you know the rules well enough you get to break them ;).

lint/linter.py
```diff
@@ -1050,9 +1064,9 @@ def get_cmd(self):
         # For backwards compatibility: SL3 allowed a '@python' suffix which,
         # when set, triggered special handling. SL4 doesn't need this marker,
         # bc all the special handling is just done in the subclass.
-        which = cmd[0]
-        if '@python' in which:
-            cmd[0] = which[:which.find('@python')]
+        # which = cmd[0]
+        # if '@python' in which:
+        #     cmd[0] = which[:which.find('@python')]
```
braver: @pykong should this be commented out?

sublime_linter.py
```diff
@@ -316,6 +300,10 @@ def highlight(self, view, linters, hit_time):
             panel.fill_panel(window, update=True)
 
         for view in window_views.values():
+            # stop lint indicator
+            ind = persist.indicators[view.id()]
+            ind.stop()
```
braver: I dislike short variable names if the abbreviation isn't common (e.g. "img" is fine). This particular one (and a similar one a few lines down) should also work without introducing a variable right?

pykong: I think those are fine as long as the abbreviation can be deduced from context.
In this case it is clear `ind` stands for *ind*icator. Nonetheless have changed it to a longer variable name. Yes, we could just chain the `start()`/`stop()` method, but I find it less readable.

braver:
The W/E counters in the statusbar now blink on every change: the numbers disappear for a split second, and then re-appear with the new value.

pykong:
I see what I can do about that blinking.

braver:
Also, sometimes no lint errors show up, not even after running "lint all views". I'm running in debug and see lot's of output but the counters stay at zero and the panel is empty as well...

kaste:
I don't think the indicator should be part of this PR.

Note: If you patch `hit`, the indicator starts if we record (queue) the hit. The queue callback `lint` seems better. But `Linter.lint_view` can still return early so that its callback `highlight` is never called.

lint/linter.py
pykong: @FichteFoll  `.values()` ? Shouldn't it be keys?

lint/linter.py
```diff
@@ -15,6 +16,18 @@
 ARG_RE = re.compile(r'(?P<prefix>@|--?)?(?P<name>[@\w][\w\-]*)(?:(?P<joiner>[=:])(?:(?P<sep>.)(?P<multiple>\+)?)?)?')
 BASE_CLASSES = ('PythonLinter',)
 
+MATCH_DICT = OrderedDict({
```
pykong: Who spots the problem here? :-)

pykong:
Closing this PR, as branch `next_namedtuple` has been merged into branch `braver`.

PR #733, opened 2017-12-23
Confirm to `context_sensitive_executable_path` type

kaste:
`context_sensitive_executable_path` should return

	(True, None) if it cannot find an executable
	(True, <path>) if it has an executable path
	(False, None) if it wants the default behavior of the standard linter

If we return (False, '') we basically just do a `which(cmd[0])` again.
If we return (False, \<path>), the returned '\<path>' is actually ignored.


kaste:
Note that nothing really changed here recently. It was just poorly, confusing implemented and worked.

Issue #541, opened 2017-02-14
Broadcast an event when linting has finsihed


kaste:
I suggest you emit an event whenever SublimeLinter has finished linting a view. A user can listen to these events and react e.g. by annotating the view, showing phantoms etc. A showcase plugin can be found [here](https://github.com/kaste/SublimeLinter-Annotations). This plugin monkeypatches `SublimeLinter.highlight` which I found to be a good first candidate to send these events. [See the simple patch.](https://github.com/kaste/SublimeLinter-Annotations/blob/master/monkeypatch_sublimelinter.py)

Following this way, the builtin 'views' (regions marker, status bar updater), could be decoupled from SublimeLinter. In the long run.

[See a simple tooltip 'viewer'.](https://github.com/kaste/SublimeLinter-Annotations/blob/master/show_tooltips.py)

braver:
I'm not sure about the "why" here? To separate the front-end/UI from the running of the linters? What does that accomplish?

kaste:
I have actually three separate view plugins running with this monkeypatch system. I have custom phantoms, tooltips, and a patched status bar indicator. 

The question was how I could actually hook into SublimeLinter, and unless I'm wrong or everything changed since then, there was no official way to write custom "views". Then the first request would be, you provide such a hook, and `WindowCommand` is the best you can do here IMO.

But of course, you could use such a hook internally as well. Back in Feb the actual data model was not ready before `highlight` for which the name suggests it is actually a view function. So naturally, as a system grows and evolves, you can always factor things differently and choose to decouple more. I think it would be great if we would just install view plugins on top of SublimeLinter, a bit more competition helps. If I remember correctly, at the time I wrote my phantoms view, SublimeLinter actually had none to offer.

kaste:
PS Does SublimeLinter actually show phantom/inline annotations? 😟

braver:
> Does SublimeLinter actually show phantom/inline annotations?

No, although they sound nice for linting, they're actually super annoying in practice. In combination with a command that is run manually it might be useful, but we're working on some other stuff first.

kaste:
I've not told you to do that, okay? I have already written a phantoms view. 

![image](https://user-images.githubusercontent.com/8558/33223834-c510c186-d163-11e7-8528-c3e19094c5a2.png)

I also colorize the complete status bar on hard errors. (Some Linters have a good separation of what is a warning and what is actually an error.)

[...]

PR #744, opened 2017-12-27
Strawman: A status bar view

kaste:
Following a bit my own #541

**I** think if we decouple the views and send events around, we get a better architecture and can more easily add features. Here and as a strawman, bc @braver asked what we can accomplish with #541, I implement the status bar (including the busy indicator) as a separate view.

Of course, we can bikeshed about everything in here. Events can be implemented in different ways. But overall we just have a cleaner approach, and as you can see with the second commit, one can easily add new features like an indicator, without getting a mess and a lot of bugs. Note that I don't instantiate an abstract `Indicator()` for every view. This is way less overhead IMO.


status_bar_view.py
```diff
@@ -0,0 +1,120 @@
...
+
+import time
+from itertools import cycle
+
+from .lint import persist, util
+from .lint.const import STATUS_KEY
+from . import events
+
+
+State = {
```
FichteFoll: This should be lower-cased.

braver:
I'm not sure about this. For one I don't feel the need to add features or to prepare to facilitate that, although an improved architecture to support existing features could be welcome. As for the indicator itself it has a similar problem to the one @pykong proposed, where it starts to blink and attract attention if there are edits in quick succesion, e.g. when holding backspace to remove a bunch of code. 
Do we even need an indicator? I understand some linters may be slow (primarily mypy it seems), but perhaps we should approach this differently. Maybe it just need to be a different design that doesn't replace the counters when it runs.

kaste:
As for the indicator, I only added that bc I saw it in the other branch and thought this is the consensus. I'm relatively 👎 At least the design had to change and the INITIAL_DELAY of 1 second just produces noise. 

The strawman here is more about architecture.

braver:
Yeah, if I didn't design it wrong I might've been happier about it 😄 . Anyway, I'm not against the indicator and I'm pro this approach, so if anyone really wants an indicator they can use this as a starting point. I'm just going to let it sit here for a while.

pykong:
**I** want the LintIndicator. @FichteFoll told me the same on Discord.
As a user of the notoriously slow MyPy I would appreciate if we find a general way to subtly signalize linting activity. 

In this regard:
What are your suggestions for a possible look of this indicator?
Also how could we avoid that "blinking" misbehaviour @braver did not like 

@kaste Kudos for this PR. It really gives us a cleaner architecture.

braver:
I think we should simply add the static text “Linting...” when it’s busy for more than 1sec, and not touch the counters.

kaste:
Let's see how this one goes.

kaste:
Fixed buffers with multiple views.  Sorry, but why do we organize all our data around view ids if they share the same buffer? I just followed the usual stuff in `sublime_linter.py` and now I'm feeling stupid. E.g. our *queue* is organized around *vid* 

FWIW I don't get the "blinking" behavior. At least not with a simple "hold backspace down".

pykong:
I think we should approach the lint indicator design from the perspective of universal design ideas.
Movement is ubiquitously used to signal some form of ongoing process.
More specifically circular movement, i.e. a spinner, is used to indicate site loading.

**I** believe we should stick to such common design elements. Hence I fully support @kaste 's most recent commit, featuring a spinner in the status bar.

Another advantage of this spinner over a static text, is that it is far less bulky, fitting into a single char width. We all know how crammed up a status bar can get, therefore we may take up as little space as from the status bar as possible, while still maintaining comprehensibility. The spinner does well in this respect.

kaste:
This took a relative long time to build, and the Sublime API was not my friend here, but after all, I actually like the current design as well. 👍 

Really a TIL day

pykong:
@kaste Regarding the `vid centricity` of our data structures. I am the culprit. At least to a degree.
I started to build new dictionaries, in the like of those already present (e.g. "persist.errors"). Three month back, no one even thought of SL4, but maintaining backwards compatibility was the foremost priority. 
Hence I did nothing to refurbish those old structures.

I have not been able to reproduce the blinking behavior with my own implementation of the indicator, either. This proves the blinking is not an insurmountable problem.

braver:
Pretty awesome guys. 💯

braver:
Ugh, I pulled too fast. @kaste what to do about this lint error:

screen shot 2018-01-03 at 12 59 07

FichteFoll:
`except Exception:`

kaste:
Fixup is #762

kaste:
My spinner is gone. 😢 @pykong Please tell @braver that my spinner was 🔥

PR #745, opened 2017-12-28
Lint the visible views from the active window on startup

kaste:
This is still an improvement over SL3 which only linted(?) the active (focused) view. 

It is way better for startup of course. But since Sublime never opens a panel on startup, we have plenty of time after startup where we can lazy lint other views. If at all.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -68,17 +68,17 @@ def plugin_loaded():
 
 def visible_views():
     """Yield all visible views of the active window."""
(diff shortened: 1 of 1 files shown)


braver:
> If at all.

Indeed. I'm not convinced it is necessary. 

I really like this approach. Could you give the single-character variables a more readable name though? They're short lived, but I don't want to encourage using super short names. That's all that's stopping me from merging this.

kaste:
Done.

PR #746, opened 2017-12-29
Do not 'hit' the active_view twice at startup

kaste:
Sorry, you merged too fast.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -55,11 +55,6 @@ def plugin_loaded():
     plugin = SublimeLinter.shared_plugin()
     queue.start(plugin.lint)
 
-    # This ensures we lint the active view on a fresh install
-    window = sublime.active_window()
-    if window:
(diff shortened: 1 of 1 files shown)


braver:
Ohoh, now everyone knows I don’t review and test everything...

kaste:
Still not working as expected. I 'hit' the active view before the other views, which will `queue.put()`. Still, the active view will be linted last. Or second at least, I only test with two view_groups. But a python Queue is FIFO, isn't it?

braver:
My cat is currently preventing me from getting out my chair, but I’ll have a look later. I didn’t notice anything weird in your previous PR though.

kaste:
Okay, got it. Our internal code runs tasks added to the queue within the same tick in random order.

braver:
Ah. I just confirmed the behaviour here. I don't think it's much of a problem really.

PR #748, opened 2017-12-29
Queue replacement

kaste:
The old queue implementation used its own `threading.Thread` to do async
background work. It had to use polling for that.

Here, we use the builtin async threading that comes with Sublime. This has
the advantage that we respect the event system of Sublime. Sublime may
decide that it is busy right now, and will therefore delay our workers.

It is also an easier and shorter implementation.

A small trick is applied for startup so that the initial linting of the
visible views actually occurs even without user interaction.

### DIFF
--- a/lint/queue.py
+++ b/lint/queue.py
@@ -1,116 +1,70 @@
-"""This module provides a threaded queue for lint requests."""
+import sublime
 
-from queue import Queue, Empty
-import threading
-import traceback
+from . import persist
+
 import time
 
-from . import persist, util
+
+# Map from view_id to timestamp of last 'hit' (e.g. an edit)
+last_seen = {}
 
 
+# For compatibility this is a class with unchanged API from SL3.
 class Daemon:
-    """
-    This class provides a threaded queue that dispatches lints.
-
-    The following operations can be added to the queue:
-
-    hit - Queue a lint for a given view
-    delay - Queue a delay for a number of milliseconds
-    reload - Indicates the main plugin was reloaded
-
-    """
-
-    MIN_DELAY = 0.1
-    running = False
-    callback = None
-    q = Queue()
-    last_runs = {}
-
     def start(self, callback):
-        """Start the daemon thread that runs loop."""
-        self.callback = callback
-
-        if self.running:
-            self.q.put('reload')
-        else:
-            self.running = True
-            threading.Thread(target=self.loop).start()
-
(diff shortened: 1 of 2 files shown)


kaste:
For review convenience, I left some `print` statements here 😉

lint/queue.py
```diff
@@ -1,116 +1,70 @@
...
-        a "delay" setting is not available in any of the settings, MIN_DELAY is used.
-        """
-        if persist.settings.get('lint_mode') != 'background':
-            return 0
+def get_delay():
+    """Return the delay between a lint request and when it will be processed.
 
-        delay = persist.settings.get('delay', self.MIN_DELAY)
+    If the lint mode is not background, there is no delay. Otherwise, if
+    a "delay" setting is not available in any of the settings, MIN_DELAY is used.
```
braver: Delay is guaranteed to be available because it's in the default settings and those are immutable now. The `MIN_DELAY` can go.

braver:
Pretty neat and seems to work at least as well as the old implementation. Perhaps I need to test it further, but it looks really good right now.

FichteFoll:
Just to confirm, linter processes aren't run or polled synchronously from the async thread now?

kaste:
This was pulled too early as well. I try a quick fix, otherwise we MUST revert. 

👍 
- Very simple implementation of a debouncer. 
- FIFO. 
- The queue itself runs on the *shared* worker thread Sublime provides, so Sublime will delay us if it has to much to do

👎 
- The whole linting runs on the *shared* worker, which is not OK. We cannot block here for 1 sec. 

What I wanted to do here is to swap

```
callback(vid, hit_time)
```
with
```
threading.Thread(lambda: callback(vid, hit_time)).start()
```

So we immediately return control back to Sublime. We do not block. After that, multiple linters can run in parallel. At any time, we can go back to the Sublime thread, by invoking `set_timeout_async` again. 

This works totally okay, but I need an additional `running` lock per view_id/buffer_id bc of mutation hell in the linter. And I was totally lost in doing that quickly, bc I actually never did that in python before. So I need to read the python manual and see what they have. 

But if anyone here just knows how to it, you're welcome.

braver:
Ok, I can't judge this well enough. Let me know if I need to revert (or open a PR that does so if needed).

kaste:
Fixup is #763

PR #751, opened 2018-01-01
Settings validation

pykong:
**Up for review and testing.**

This PR implements the validation of settings via `jsonschema`.

**Note**: Before merge, the availability of the `jsonschema` package over PackageControl must be ensured.


FichteFoll:
This PR is largely identical to #752?

Default.sublime-commands
```diff
@@ -3,6 +3,10 @@
         "caption": "SublimeLinter: Lint This View",
         "command": "sublime_linter_lint"
     },
+    {
+        "caption": "SublimeLinter: Lint All Views",
+        "command": "sublime_linter_lint_all"
```
braver: What is this command doing here?

commands.py
```diff
@@ -35,7 +35,22 @@ def is_enabled(self):
...
         """Lint the current view."""
         from .sublime_linter import SublimeLinter
-        SublimeLinter.shared_plugin().hit(self.view)
+
+        plugin = SublimeLinter.shared_plugin()
+        plugin.hit(self.view)
+
+
+class SublimeLinterLintAllCommand(sublime_plugin.WindowCommand):
+    """A command that lints all views that are lintable."""
```
braver: Could you check if your change set here matches your expectations?

sublime_linter.py
```diff
@@ -15,30 +15,8 @@
...
-            new_name = "SublimeLinter (old).sublime-settings"
-            new_path = os.path.join(usr_dir_abs, new_name)
-            os.rename(settings_file, new_path)
-            msg = "SublimeLinter\n\nYour settings have been backed up to:\n{}\nin Packages/User/".format(new_name)  # noqa: 501
-            sublime.message_dialog(msg)
-
-
 def plugin_loaded():
-    backup_old_settings()
+    util.backup_old_settings()
```
braver: I don’t want this moved to util because nothing else should use it and keeping it close together makes it easier to remove later on.

braver:
I think the default output of schema validation is a bit verbose. We could do with just the first line, `'scope' is a required property` in the example below.

```
'scope' is a required property

Failed validating 'required' in schema['properties']['styles']['items']:
    {'additionalProperties': False,
     'properties': {'icon': {'type': 'string'},
                    'mark_style': {'pattern': '^(fill|outline|solid_underline|squiggly_underline|stippled_underline|none)$',
                                   'type': 'string'},
                    'priority': {'minimum': 0, 'type': 'integer'},
                    'scope': {'type': 'string'},
                    'types': {'items': {'pattern': '^(error|warning)$',
                                        'type': 'string'},
                              'minItems': 1,
                              'type': 'array',
                              'uniqueItems': True}},
     'required': ['mark_style', 'priority', 'scope', 'icon', 'types'],
     'type': 'object'}

On instance['styles'][1]:
    {'mark_style': 'squiggly_underline', 'priority': 1} 

```

The schema should also be updated to remove the settings that have been remove recently.

braver:
Also please note the flake8 error: `E722 do not use bare except'`

pykong:
Done Only the first line of the exception message will be printed out to console.

The `bare except` is from events.py. kaste has fixed that in one of the most recent PRs.
Thereby it should resolve on merging.

braver:
That last commit isn’t necessary, the dependency is published as “jsonschema”, the name of the repo isn’t used. I tested it, it works.

pykong:
Do not know how to revert it by git means.

braver:
I think it's `git revert ff18cbf`, but I tend to use a GUI or GitHub for the less obvious operations.

braver:
squash merged it in 8a830dac6cbce5bfab6970ac58984bee5a455ea7!

PR #752, opened 2018-01-01
Panel select

pykong:
**Up for review and testing.**

This PR updates panel selection based upon caret position in corresponding view.
Thereby the current file is automatically focused and GoTo Next/Prev does not lead to any context switch.

Hence this PR greatly improves user experience.


FichteFoll:
What feature (set) does this PR add now? Maybe update OP.

settings-schema.json
```diff
@@ -0,0 +1,84 @@
+{
```
pykong: This shouldn't be part of this PR.

braver:
@pykong please check your PR’s with regards to the notes I put in #751 , each branch should just have the minimum changes related to the feature you’re adding, but you’ve got some other cruft in there as well.

braver:
@pykong can you resolve the merge conflict for me?

pykong:
@braver `on_selection_modified_async` was removed when we switched to the event driven architecture. This led to the new panel update method introduced with this PR never been called. It now works as of: ef701ac

braver:
Ah, makes sense. Thanks!

braver:
I get this really weird behaviour where in util.py it jumps from line 72 to line 69 instead of to the next error at line 83. I can go forwards or back as much as I like but it trips up there for some reason.

weird

pykong:
I can not reproduce this issue. Do you got another example?
Why close this PR?

braver:
I guess it was auto-closed because I removed `next`.

pykong:
panel_select

I am still trying to get any output from `pycodestly`. In lieu of the latter all I can only guess what the described issue is.

I have so far in my testing never encounter a trip up like described. Is it maybe linter related?

FichteFoll:
I noticed in your preview that you have two errors on a line, yet only one is selected in the panel. Other than that this looks very neat.

pykong:
@FichteFoll Thanks. Great you like what you see!
The selection dependends on the column position as well.
But even if more than a single error encompass the column, only the first error will be selected in the panel.

It is a kind of UI question what is best here.

I suggest that we get the error @FichteFoll describes fixed and then merge.
Other features like the folding or changes to the panel selection can follow in their own PRs.

FichteFoll:
>The selection dependends on the column position as well.

Good point. In your particular example, both errors are on the same column though.

I guess the most sensible approach would be to select the nearest error if the caret is between two errors on the same column and hl multiple when they are on the same column.

pykong:
> I guess the most sensible approach would be to select the nearest error if the caret is between two errors on the same column

This is how it is done in the current implementation

> [...] and hl multiple when they are on the same column.

Yes. I think this makes sense to highlight multiple.

braver:
It's broken for everything but the first file in the panel. The second file requires 2 actions to go to the next error, the third file it just jumps back and forth between the file name and the first result.

This is with just hitting "next_result" continuously:

weird

pykong:
I'll check that next week.
Rather busy this WE with other stuff.

panel/panel.py
```diff
@@ -169,10 +196,95 @@ def fill_panel(window, types=None, codes=None, linter=None, update=False):
...
+        colno = None
+
+    full_line = False
+    view_dict = persist.errors.get_view_dict(vid)
+    if not view_dict or util.is_none_or_zero(view_dict["we_count_view"]):
+        return
+
+    line_dicts = view_dict["line_dicts"]
+
+    if lineno:
```
kaste: What if `lineno == 0`?

panel/panel.py
```diff
@@ -169,10 +196,95 @@ def fill_panel(window, types=None, codes=None, linter=None, update=False):
...
+
+    if not region_dict:
+        return
+
+    if full_line and colno is not None:
+        full_line = region_dict["start"] <= colno <= region_dict["end"]
+
+    panel_lineno = region_dict.get("panel_lineno")
+
+    if panel_lineno:
```
kaste: The same. Is this really a truthy test or a `None` test?

sublime_linter.py
```diff
@@ -169,6 +169,15 @@ def on_hover(self, view, point, hover_zone):
             if persist.settings.get('show_hover_region_report'):
                 SublimeLinter.shared_plugin().open_tooltip(view, point, True)
 
+    def on_selection_modified_async(self, view):
```
kaste: I would really appreciate if we don't immediately start and fill sublime_linter.py with stuff that belongs to a special view? I've shown with the statusbar view and the busy indicator, that this really helps readability. 

I think this is really important especially for open source projects, where a user must find a bug and a way to help and contribute very fast, not knowing the code base.

sublime_linter.py
```diff
@@ -169,6 +169,15 @@ def on_hover(self, view, point, hover_zone):
...
+        if not view:
```
kaste: Just curious. Is that really that Sublime calls us sometimes with no view? Just with nothing?

panel/panel.py
kaste: In other words: It is important to state if your data is (0, 0) based and when it is (1, 1) base. (1, 1) based is what we present the user. Bc something like this is known to err very easy.

sublime_linter.py
braver: I believe because the async-ness the view may no longer exist by the time this code is executed. It’s the reason behind a lot of the `if view` checks in this and other packages.

FichteFoll: Indeed, if the async thread was blocking on something earlier while the view was closed (which hopefully doesn't happen but could), then the view would most likely not exist anymore. However, I'm not sure if the function is even called at this point and, in addition, whether the view's ID might just be invalid instead. It's usually safe to check for `view.window()` for when a view object is invalid.

kaste: Yeah okay, but you can't just `None` a var that is in our scope. python will not garbage collect this.

So I tested.

```
class UpdateState(sublime_plugin.EventListener):
    def on_selection_modified_async(self, view):
        current_pos = get_current_pos(view)
        import time
        time.sleep(1)
        print('test_view', view.id(), current_pos, view.file_name() or view.name(), view.window())


def get_current_pos(view):
    try:
        return view.rowcol(view.sel()[0].begin())
    except (AttributeError, IndexError):
        return -1, -1

```

If you run this, every callback blocks for a second. So you have plenty of time to close a view. Sublime will then still call you, even after the view is gone. The view is still valid, you can call `id()` and even `rowcol()`, but it is detached so `window()` returns `None`.

sublime_linter.py
```diff
@@ -169,6 +169,15 @@ def on_hover(self, view, point, hover_zone):
...
+            return
+
+        lineno, colno = self.get_line_and_col(view)
```
kaste: The `view` here doesn't have to be the `active_view`. 

- You can modify views in the background, and this will trigger.
- If you have multiple views into the same buffer, the view here will always be the *primary* one.

That's why I have the `State` object in `status_bar_view.py`, which holds the `active_view` and `current_pos`.

pykong:
Good points @kaste 
I have moved the call of the `update_panel_selection` function to `status_bar_view`, where it should have been in the first place. Works like a charm.
The function will now simply return early, if `if current_pos == (-1, -1)`, this allowed me to cleanup its logic.

braver:
Ok, seems to work pretty good now. One thing: it doesn't update the selection in the panel as I switch files. Only after I move my cursor does it sync up. Everything else = 🌶

braver:
Please see #797

Issue #740, opened 2017-12-25
[beta] Cleaner syntax map for settings


pykong:
The current syntax map quickly grows unreadable when adding new entries.
A cleaner structure may consist of an object mapping `syntax_name` to and array of `syntaxes`.
_(See proposed syntax map below.)_

**Current:**
```
    "syntax_map": {
        "html (django)": "html",
        "html (rails)": "html",
        "html 5": "html",
        "javascript (babel)": "javascript",
        "magicpython": "python",
        "php": "html",
        "python django": "python",
        "pythonimproved": "python"
    }
```

**Proposed:**
```
"syntax_map": {
        "html":  [ "html (django)",  "html (rails)", "html 5", "php"],
        "javascript":  ["javascript (babel)"],
        "python": [  "magicpython" , "python django", "pythonimproved"]     
    }
```

FichteFoll:
Related proposal: https://github.com/SublimeLinter/SublimeLinter3/issues/691

braver:
See #691

Issue #691, opened 2017-12-10
Base linter selection on scope instead of syntax file name


FichteFoll:
Basing the linter selection on the base scope is much more flexible as opposed to having a large `syntax_map` mapping, since most JavaScript-like languages *will* be using `source.js` as the base scope and possibly add more leaves to that, e.g. `source.js.babel`. Or `source.json.sublime.settings`.

Kind of nullifies #431 too, since all use cases I had in mind for that would be solved by this solution.

To outline the extend of this problem, this is my current `syntax_map` settings (which I didn't even realize to have grown so much – probably the work of a plugin, and I suspect a theme providing icons, like @ihodev's A File Icon):

```
    "syntax_map": {
        "coffeescript (gulpfile)": "coffeescript",
        "html (django)": "html",
        "html (rails)": "html",
        "html 5": "html",
        "javascript (babel)": "javascript",
        "javascript (eslint)": "javascript",
        "javascript (gruntfile)": "javascript",
        "javascript (gulpfile)": "javascript",
        "javascript (postcss)": "javascript",
        "javascript (stylelint)": "javascript",
        "javascript (webpack)": "javascript",
        "json": "javascript",
        "json (babel)": "json",
        "json (bower)": "json",
        "json (composer)": "json",
        "json (eslint)": "json",
        "json (npm)": "json",
        "json (postcss)": "json",
        "json (settings)": "json",
        "json (stylelint)": "json",
        "json (sublime)": "json",
        "json (tern js)": "json",
        "php": "html",
        "python django": "python",
        "sublime commands": "javascript",
        "sublime completions": "javascript",
        "sublime key map": "javascript",
        "sublime macros": "javascript",
        "sublime settings": "javascript",
        "sublime text build system": "javascript",
        "sublime text syntax def (json)": "javascript",
        "sublime text syntax def (yaml)": "yaml",
        "xml (config)": "xml",
        "xml (svg)": "xml",
        "yaml": "yaml",
        "yaml (circleci)": "yaml",
        "yaml (docker)": "yaml",
        "yaml (eslint)": "yaml",
        "yaml (lock)": "yaml",
        "yaml (procfile)": "yaml",
        "yaml (stylelint)": "yaml",
        "yaml (yarn)": "yaml"
    },
```

braver:
Let’s close this once the PR has been merged.

braver:
via @pykong in #740

> The current syntax map quickly grows unreadable when adding new entries.
[...]

PR #753, opened 2018-01-01
New syntax selection based upon scope.

pykong:
Up for review and testing.

Addresses: #691 #740
- Also easier syntax mapping.

Notes:
- due to changes to settings file, a corresponding jsonschema needs to be updated as well.
- ~~the syntax map needs to be expanded to support other linters~~

**Update 1:**
```
       "syntax_map":{
            "type":"object"
        }
```
Will work with this PR.
Still, that `jsonschema` definition could be made more specific/stringent.
  
**Update 2:**
The branch `settings_validation` got merged into this one, as I wanted to use the new `load_json` function and also remove the `syntax_map` entry from the schema file.
  

**Update 3:**
My bad. I should have more clearly stated that the syntax file:
```
resources/syntax_map.json
```
need to be expanded.


lint/util.py
```diff
@@ -49,22 +49,13 @@ def get_syntax(view):
...
-        else:
-            view_syntax = ''
-
-    return mapped_syntax or view_syntax
+    from .persist import settings
+    syntax_map = settings.get('syntax_map')
+    for syntax, selectors in syntax_map.items():
+        for s in selectors:
+            if view.match_selector(0, s):
+                return syntax
```
FichteFoll: I wonder whether this should return the syntax with the maximum selector match score. Like follows:

```py
try:
    syntax, score =  max(
        (max(view.match_selector(0, s) for s in selectors, default=0),
         syntax)
        for syntax, selectors in syntax_map.items()
    )
except ValueError:
    return ""
else:
    return syntax
```

lint/util.py
```diff
@@ -49,22 +49,13 @@ def get_syntax(view):
...
-            view_syntax = match.group(1).lower()
-            from .persist import settings
-            mapped_syntax = settings.get(
-                'syntax_map', {}).get(view_syntax, '').lower()
-        else:
-            view_syntax = ''
-
-    return mapped_syntax or view_syntax
+    from .persist import settings
+    syntax_map = settings.get('syntax_map')
```
FichteFoll: Default value?

SublimeLinter.sublime-settings
```diff
@@ -119,14 +119,10 @@
...
     "syntax_map": {
-        "html (django)": "html",
-        "html (rails)": "html",
-        "html 5": "html",
-        "javascript (babel)": "javascript",
-        "magicpython": "python",
-        "php": "html",
-        "python django": "python",
-        "pythonimproved": "python"
+        "html": [ "source.html"],
```
FichteFoll: Leading space inside the array here.

FichteFoll:
The change is still backwards incompatible for linters that do not target a syntax contained in the syntax map, like say yamllint which [targets the YAML syntax](https://github.com/thomasmeeus/SublimeLinter-contrib-yamllint/blob/76f7079526ec83e8fb83d08185dec828300a4a1c/linter.py#L19).

A fallback similar to the one provided by @gwenzek for the Discord Rich Presence plugin could be implemented: https://github.com/Snazzah/SublimeDiscordRP/pull/15

lint/util.py
pykong: We have gone great l length to separate data from code, which have been intertwined pre-SL4.

The default settings are now read-only. Also, we are going to implement a settings validation via the new `jsonschema` dependency.

Hence there will always be a syntax-map provided.

pykong:
@FichteFoll Thank your review and code suggestions.

> The change is still backwards incompatible for linters that do not target a syntax contained in the syntax map [...] Technically, this would only require us to specify javascript in the default map since the others already match the second base element

Are you suggesting we should simply extend the syntax map?
(A need which I have already pointed out.)

lint/util.py
pykong: Just to be certain, we like to use `view.match_selector` not `view.score_selector()` here?

FichteFoll: Yes, that should be `score_selector`.

FichteFoll:
No, I'm saying that the "fallback" for no matching selector should be to select the second segment of the base scope (i.e. the thing after `text` and `source`), *unless* a selector in our map matched. 

That way we only need to specify `source.js => javascript`, as `source.python` gets mapped to `python` automatically, and similar.

lint/util.py
pykong: I'll take that comprehension as a suggestion, because it is very unreadable.
Check whether you agree with my implementation of scoring based syntax selection.

lint/util.py
```diff
@@ -44,18 +44,20 @@ def printf(*args):
...
-        for s in selectors:
-            if view.match_selector(0, s):
-                return syntax
-    return ""
+    try:
+        scoring = {
+            max(view.score_selector(0, s) for s in selectors if s): syntax
+            for syntax, selectors in syntax_map.items()
+        }
+        max_score = max(scoring, key=scoring.get)
```
FichteFoll: Pretty sure the provided `key` is doing the opposite of what you want here. It will order the scores after the associated syntax string and not by their numeric value.

lint/util.py
```diff
@@ -44,18 +44,20 @@ def printf(*args):
...
     from .persist import settings
     syntax_map = settings.get('syntax_map')
-    for syntax, selectors in syntax_map.items():
-        for s in selectors:
-            if view.match_selector(0, s):
-                return syntax
-    return ""
+    try:
+        scoring = {
+            max(view.score_selector(0, s) for s in selectors if s): syntax
```
FichteFoll: If the iterable for `max` is of zero length, it raises a `ValueError`, but I think we should handle this case by providing a `default` kwarg.

pykong: Which one?

lint/util.py
pykong: Good spot. I have switched to score: syntax.
That handles the unlikely case we get two syntaxes with the same score.
The latter in our list overwriting the former.

pykong: @FichteFoll based on the current state of code. Could you make a suggestion how to implement that fallback mechanism? Is it as simple as, splitting - the - scope on whitespace and iterating the scoring algorithm over its fragments? What am I missing here?

lint/util.py
FichteFoll: well, if `selectors` is empty, `max` will err out. But this is the inner max for the entries' selectors, not as a whole, so when there's an empty list for one map, the function will always return `""`.

The solution would be to invoke `max(iterable, default=0)`. Regardless, this will only happen for invalid configurations (i.e. not having a single selector entry in the array).

lint/util.py
FichteFoll: Pretty much? I mean, I linked a reference implementation.

braver:
Uhm, it's using the python linters to lint a less file?

screen shot 2018-01-06 at 11 49 13

Also, stylelint doesn't seem to work at all.

pykong:
Hm, might be related to the scoring based syntax detection.
I should define a cutoff.

> stylelint
Did you extend the syntax_map, at all?

braver:
> Did you extend the syntax_map, at all

No, I didn't. stylelint declares less as one of it's syntaxes, so I think it should match `source.less` automatically? I don't have to do anything for it to work pre this PR.

pykong:
@braver As I told you before, we are going to need a huge mapping file. Try with adding an respective entry to our new mappings file:
```
resources/syntax_map.json
```

a line like:
```
"less": ["source.less"]
```

might do

braver:
Oh sure, right... Building that mapping would definitely have to be part of this PR then.

pykong:
@braver My bad. I should have also repeated that info together with this PR. 

I wonder whether there is any more elegant way than this long syntax map?
I mean basically for most syntaxes the name is exactly like the second segment of the base scope:

e.g.:
```
"python": ["source.python"]
````

braver:
We could attempt naively if `source.<syntax>` is the full base scope and if so run the linter?

pykong:
I have too little experience with Sublime to tell whether this approach is reliable. Maybe @FichteFoll can provide some guidance in these matters?

FichteFoll:
Uhm, did you even read through the reference implementation I linked? I mean, I could take care of this as it's quite isolated, but probably not until two weeks or so.

I'm only reading issues I'm some spare time currently and also have to take care of pull requests to the PCC as the next priority.

pykong:
Replaced by #799

Issue #710, opened 2017-12-17
[beta] Add blacklist to hide errors from the panel


FichteFoll:
For some linters or errors, I don't want to see a summary in the panel, e.g. the annotations plugin.

braver:
You can add filtering arguments when you call toggle_panel. See the keybindings settings for how to do that. Would that work for you?

FichteFoll:
Can you exclude certain linters? From how the invocation looks like, I assume it's an "include only" type of filter.

braver:
Yes, I think it’s a whitelist. @pykong

pykong:
Indeed it is whitelisting only.

pykong:
Solved by referenced PR. Discussion may continue over there.

braver:
fixed in #798

PR #754, opened 2018-01-02
Panel: Implement blacklisting.

pykong:
Addresses: #710 

**NOTE:** #752 should be merged before this.


braver:
I notice that you may filter it so that no errors remain in the panel. However, a file *is* listed if it has errors, regardless of filtering. This is of course also an issue with the existing whitelist. Think we might be able to put a message in there, e.g. "No errors due to filtering"?

FichteFoll:
I would just filter the file as well.

braver:
Yes, that would also be good :)

pykong:
The best way to implement the change requested requires the `RenderLines` class from the branch `panel_select`. What is the best way to fare, *git* wise here?

braver:
Well, because it's git there are at least 3000 ways to do it, but I'm fine if you just merge the two branches and present a single PR for both features.

FichteFoll:
I would try to get `panel_select` merged first. In the meantime, you can start implementing this on a separate branch starting from panel_select and then rebase from `master` once that is merged (and submit a PR).

pykong:
Closed. Changeset migrated to #798

PR #756, opened 2018-01-02
Do not `update_gutter_icons` per `activate`

### LINKED ISSUE #749: [beta] perf: `update_gutter_icons` called per `on_activated_async`
https://github.com/SublimeLinter/SublimeLinter/blob/c6438d25748f2ad50e61ec96570e12b715d8fe93/sublime_linter.py#L97-L102

https://github.com/SublimeLinter/SublimeLinter/blob/c6438d25748f2ad50e61ec96570e12b715d8fe93/lint/settings.py#L58-L64

I don't know why we do `settings.load()` per activate anyway. It doesn't take an argument, and seems global. Maybe the call in `plugin_loaded` is enough.

kaste:
Fixes #749

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -120,7 +120,7 @@ class LinterMeta(type):
 
             # The sublime plugin API is not available until plugin_loaded is executed
             if persist.plugin_is_loaded:
-                persist.settings.load(force=True)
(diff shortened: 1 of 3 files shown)


braver:
Thanks!

PR #761, opened 2018-01-03
[beta] Fix bug in match dictionary

willrowe:
A bug was introduced in af5527935df20303ae4afbdf8476aafb33cde0ac which causes an error when an individual linter's regex matches groups that are not defined in the [MATCH_DICT](https://github.com/SublimeLinter/SublimeLinter/blob/af5527935df20303ae4afbdf8476aafb33cde0ac/lint/linter.py#L19-L29).

For example, this happens in both the [pylint](https://github.com/SublimeLinter/SublimeLinter-pylint) and [php](https://github.com/SublimeLinter/SublimeLinter-php) linters.

The pylint linter matches a [`code` group](https://github.com/SublimeLinter/SublimeLinter-pylint/blob/bc29a397d7eedbf02711ae5db9a2553a02ad5408/linter.py#L33) and the php linter matches a [`type` group](https://github.com/SublimeLinter/SublimeLinter-php/blob/cd81fe9c6e7bb664722b35716ef72a7930905958/linter.py#L21)

This pull request fixes the bug by only updating the dictionary to be passed to the tuple with keys that are already available on it. The match object from the linter supplied regex will still continue to work and allow it to perform linter-specific actions using that information, but the tuple will only be provided with the groups that are relevant to it.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1388,9 +1388,11 @@ class Linter(metaclass=LinterMeta):
         if not match:
             persist.debug('No match for regex: {}'.format(self.regex.pattern))
         else:
-            match_dict.update({key: value
-                for key, value in match.groupdict().items()
-                if key in match_dict})
+            match_dict.update({
+                k: v
+                for k, v in match.groupdict().items()
+                if k in match_dict
+            })
             match_dict["match"] = match
 
             # normalize line and col if necessary
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1388,7 +1388,7 @@ class Linter(metaclass=LinterMeta):
         if not match:
             persist.debug('No match for regex: {}'.format(self.regex.pattern))
         else:
-            match_dict.update(match.groupdict())
+            match_dict.update({key: value for key, value in match.groupdict().items() if key in match_dict})
             match_dict["match"] = match
 
             # normalize line and col if necessary
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1388,7 +1388,9 @@ class Linter(metaclass=LinterMeta):
         if not match:
             persist.debug('No match for regex: {}'.format(self.regex.pattern))
         else:
-            match_dict.update({key: value for key, value in match.groupdict().items() if key in match_dict})
+            match_dict.update({key: value
+                for key, value in match.groupdict().items()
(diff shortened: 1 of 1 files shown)


pykong:
@willrowe Thank you for your interest in SL4 and this PR.
Could you in short, explain how the potential bug you describe actually affects your usage?

willrowe:
Depending on what linters are enabled, it will throw an error. For example, when the `pylint` linter is enabled it will throw this error:

```
  File "[REDACTED]/Sublime Text 3/Packages/SublimeLinter/lint/queue.py", line 37, in worker
    callback(vid, hit_time)
  File "[REDACTED]/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 235, in lint
  File "[REDACTED]/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 748, in lint_view
    linter.lint(hit_time)
  File "[REDACTED]/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1103, in lint
  File "[REDACTED]/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1370, in find_errors
    yield self.split_match(error)
  File "linter in [REDACTED]/Sublime Text 3/Installed Packages/SublimeLinter-pylint.sublime-package", line 276, in split_match
  File "[REDACTED]/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1407, in split_match
    return LintMatch(**match_dict)
TypeError: __new__() got an unexpected keyword argument 'code'
```

lint/linter.py
```diff
@@ -1388,7 +1388,7 @@ def split_match(self, match):
         if not match:
             persist.debug('No match for regex: {}'.format(self.regex.pattern))
         else:
-            match_dict.update(match.groupdict())
+            match_dict.update({key: value for key, value in match.groupdict().items() if key in match_dict})
```
pykong: Could you please wrap that long dict comprehension according to the pattern:
```
[expression
for value in collection
if condition]
```
Readability counts.

willrowe: No problem, this has been updated in the latest commit.

pykong:
@willrowe Thank you! Have fun with SL4!

PR #763, opened 2018-01-04
Replacement for the new queue

kaste:
Now, the strategy is just to use the Python standard library.

### DIFF
--- a/lint/queue.py
+++ b/lint/queue.py
@@ -1,11 +1,15 @@
 from . import persist
 
+from collections import defaultdict
 import time
 import threading
 
+# TODO: Both `timers` and `running` herein grow unbounded. T.i. a proper
+# cleanup must be implemented.
 
 # Map from view_id to threading.Timer objects
 timers = {}
+running = defaultdict(threading.Lock)
 
 
 # For compatibility this is a class with unchanged API from SL3.
@@ -21,11 +25,13 @@ class Daemon:
         return queue_lint(vid, delay, self._callback)
 
 
-def queue_lint(vid, delay, callback):
+def queue_lint(vid, delay, callback):  # <-serial execution
     hit_time = time.monotonic()
 
-    def worker():
-        callback(vid, hit_time)
+    def worker():                      # <-concurrent execution
+        with running[vid]:  # <- If worker runs long enough
+                            #    multiple tasks can wait here!
+            callback(vid, hit_time)
 
     try:
         timers[vid].cancel()
--- a/lint/queue.py
+++ b/lint/queue.py
@@ -38,9 +38,6 @@ def queue_lint(vid, delay, callback):
     return hit_time
 
 
-MIN_DELAY = 0.1
-
-
 def get_delay():
     """Return the delay between a lint request and when it will be processed.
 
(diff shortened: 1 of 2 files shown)


kaste:
I still forgot something ... 🤣

braver:
Good busy! 😄

kaste:
Okidoki. 

Let's have true parallel linting. 🔥 🎆

kaste:
We're several seconds (!) faster at startup now.

braver:
Wow, it *is* fast now. Excellent!

braver:
@pykong or @FichteFoll want to say/ask anything smart about this before I pull it?

FichteFoll:
The diff is kinda hard to read without all the surrounding architectural changes (or how it's been before in the first place).

I'll take a note to look at it properly in full when I have more time available next week.

kaste:
Since this is a hotfix, I'll do a quick walkthrough.  The overall diff is kinda useless bc this is already a complete replacement of the queue from last week. We have two relevant commits here: https://github.com/SublimeLinter/SublimeLinter/pull/763/commits/a369376f85a5cedeef80ffdb2f042f9782337755 which introduces the new debouncer and https://github.com/SublimeLinter/SublimeLinter/pull/763/commits/1b1d5a21682aa988355e7cfe4159ea8d53d8fcec which finally introduces a lock to limit the parallelism.

The disadvantage of using `sublime.set_timeout_async` was that it runs on a *shared* thread. If we block here, all other events like `on_selection_modified_async` etc. will not run before we finish.

The implementation here thus follows the python standard library. We replace 

```
sublime.set_timeout_async(worker, delay * 1000)
```

directly with

```
threading.Timer(delay, worker).start()
```

This will already solve most of our problems. `threading.Timer` has a nice feature ` sublime.set_timeout_async` does not have. We can actually abort or cancel scheduled tasks. So most of the logic, we had before just disappears. Every time we 'hit', we just `cancel()` the previous task if any. 

After that, we basically can run `lint` in parallel. This is safe for different views which have each their own linters. So we just use a `threading.Lock` per `vid` in https://github.com/SublimeLinter/SublimeLinter/pull/763/commits/1b1d5a21682aa988355e7cfe4159ea8d53d8fcec

lint/queue.py
```diff
@@ -1,12 +1,15 @@
-import sublime
-
 from . import persist
 
+from collections import defaultdict
 import time
+import threading
 
+# TODO: Both `timers` and `running` herein grow unbounded. T.i. a proper
```
braver: Shouldn't we do something about this now?

kaste: No, it's a good solution and a hotfix. Holding 100 or so locks should be no problem. Don't know. Is it?

It's one of the moving parts. Can we make the debouncer self-contained, so the debouncer decides when to clear up old timers, or do we need to trigger the clearance from outside? 

This async/parallel stuff just takes a week to sink. This is really hard to think about. At least for me.

kaste: Okay, should I just do it? I add it to `def on_pre_close(self, view)`?

braver: > This async/parallel stuff just takes a week to sink. This is really hard to think about. At least for me.

Definitely.

braver:
To me this seems to work fine, a lot more performant than what we had first. I’m going to merge this so it spreads to more users. We can always fix or discuss or do whatever to improve it.

lint/queue.py
braver: > Okay, should I just do it?

Add the cleanup to `on_pre_close` you mean? Doesn't sound like a bad idea.

kaste: Ah, let's do it

PR #766, opened 2018-01-04
Small refactor in highlight

kaste:
Remove some cruft from `highlight.py`

- The `reset` methods were unused
- The `clear` methods were duplicated
- `persist.highlights` is unused
- The RegionStore is a singleton, only used within *highlight.py* itself, so I removed it from *persist* as well.

### DIFF
--- a/lint/highlight.py
+++ b/lint/highlight.py
@@ -22,12 +22,19 @@ NEAR_RE_TEMPLATE = r'(?<!"){}({}){}(?!")'
 
 
 class RegionStore:
-    def __init__(self):
+    _memory = None
+
+    @property
+    def memory(self):
         """structure: {"view.id": [region_keys ... ]}."""
-        self.memory = sublime.load_settings("sl_regions.sublime-settings")
-        views = self.memory.get("views")
-        if not views:
-            self.memory.set("views", {})
+        memory = self._memory
+        if not memory:
+            memory = sublime.load_settings("sl_regions.sublime-settings")
+            views = memory.get("views")
+            if not views:
+                memory.set("views", {})
+            self._memory = memory
+        return memory
 
     def add_region_keys(self, view, new_keys):
         view_id = view.id()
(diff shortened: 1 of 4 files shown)


lint/highlight.py
```diff
@@ -22,12 +22,19 @@
...
+    @property
+    def memory(self):
         """structure: {"view.id": [region_keys ... ]}."""
-        self.memory = sublime.load_settings("sl_regions.sublime-settings")
-        views = self.memory.get("views")
-        if not views:
-            self.memory.set("views", {})
+        memory = self._memory
+        if not memory:
+            memory = sublime.load_settings("sl_regions.sublime-settings")
```
kaste: Don'

kaste: We never call `save_settings` if that matters.

braver: first question: I don't know

second: we only ever *read* settings, so we shouldn't have a need for `save_settings `

kaste: But it looks like someone wanted to persist these regions across sessions. And we could use `save_settings` to do that. But it is not implemented. @pykong did this.

If we just use a normal *dict*, we don't need a lazy *property* here.

braver: Right, that sounds like a remnant of our earlier iterations. I think SL3 had a lot of caching and persistence, etc., that we don't need anymore. IMO you're free to change this.

pykong: @kaste @braver we could change that indeed if everything works as before.

The idea behind the in-memory settings was to persist region throughout a session.
So, when there is an exception that SublimeLinter will be given the chance to pick up those then remnant highlights from the previous run and clear them.

braver:
@kaste is this still WIP?

kaste:
Done. 

I found a bug here as well. https://github.com/SublimeLinter/SublimeLinter/pull/766/commits/23e431c48ae526181043b245a11b593841f4192f

PR #771, opened 2018-01-05
Translate start position for the panel view

### LINKED ISSUE #768: Generally, the col is wrong by 1 in the panel
Note that the underlining (region mark) and the status bar info is correct, but panel column is off by one.

image

kaste:
Generally, our data is (0, 0) based. We inc (line:col) by one for the
user.

Fixes #768


kaste:
image

PR #773, opened 2018-01-05
Implement simple cleanup for the queue

kaste:
Since we have a good signal when views actually get unreachable, we just
use that `on_pre_close` to trigger a clearance.

### DIFF
--- a/lint/queue.py
+++ b/lint/queue.py
@@ -4,8 +4,6 @@ from collections import defaultdict
 import time
 import threading
 
-# TODO: Both `timers` and `running` herein grow unbounded. T.i. a proper
-# cleanup must be implemented.
 
 # Map from view_id to threading.Timer objects
 timers = {}
@@ -23,6 +21,9 @@ class Daemon:
         vid = view.id()
         delay = get_delay()  # [seconds]
         return queue_lint(vid, delay, self._callback)
+
+    def cleanup(self, vid):
+        cleanup(vid)
 
 
 def queue_lint(vid, delay, callback):  # <-serial execution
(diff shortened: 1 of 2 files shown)


kaste:
That's probably good enough

braver:
Yeah, looks good to me.

Is it me or is linting a bit slower with the new queue? It doesn't block at startup so much which is great, and it seems to be cleaner architecturally, but I do think I see the statusbar indicator more often than before.

kaste:
Slower compared to the old SL3 queue or my first attempt which I had to use today and yesterday.

braver:
Compared to the old SL3 queue

kaste:
The indicator has strictly speaking nothing to do with the queue. We fire BEGIN just before `Linter.lint_view` and stop via FINISH at the end of `highlight`. 

I do see the indicator right now at startup but that is bc we're so fast that Sublime didn't finished booting itself. So we actually hang in `highlight` and wait for the regions to draw. 

I do think though that the 1 second rule will not work out. On my system, `flake8` is actually very fast - roughly 0.6sec per run. But `eslint` is slower and always takes about 1 second. (Maybe our NodeLinter is also a bit wasteful here.)

kaste:
Apply this.

[patch.txt](https://github.com/SublimeLinter/SublimeLinter/files/1609018/patch.txt)

```
@events.on(events.FINISHED_LINTING)
def on_finished_linting(buffer_id):
    start = State['running'].pop(buffer_id, None)
    if start:
        print('Elapsed time', buffer_id, time.time() - start)
```

kaste:
Update: Actually eslint is super fast, there is roughly 0.5sec overhead added by SublimeLinter. The overhead for flake8 is more like 0.01sec.

braver:
I never benchmarked SL before, so I can’t say if we regressed at all. 🤷‍♂️ It’s fine to me.

PR #774, opened 2018-01-06
Hotfix: Status bar doesn't update if you have multiple views into the same buffer

kaste:
…tive_view"

This reverts commit 9be3eb55d6025e52e61578e82be58cae227a7f42.

### DIFF
--- a/status_bar_view.py
+++ b/status_bar_view.py
@@ -49,11 +49,8 @@ class UpdateState(sublime_plugin.EventListener):
 
         draw(**State)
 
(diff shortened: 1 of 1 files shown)


kaste:
Bug: If you have multiple views into the same buffer, the simple check `view != active_view` doesn't hold event though we must update our status bar. 

A complete check here would be `view != active_view and view.buffer_id() != active_view.buffer_id()` but I don't know if this is really necessary.

braver:
> I don't know if this is really necessary.

I guess we'll find out sooner or later, or not... :)

Issue #758, opened 2018-01-02
Last hit_time is irrelevant to invalidate/abort a lint


kaste:
Throughout our code we do something like

https://github.com/SublimeLinter/SublimeLinter/blob/d2f87bace77e520488ba4b37edf01c460f6c0125/sublime_linter.py#L227-L228

to decide if we should abort the linting. But the hit_time is irrelevant. a) it is coupled with the view.id() which is irrelevant in this regards bc only the underlying buffer matters. b) if you don't lint in `background` mode, you can easily alter the view before linting has finished: 

![image](https://user-images.githubusercontent.com/8558/34500313-83808376-f00a-11e7-8f20-a70ae9b6e90f.png)

(above: \<ctrl-s> then (quickly) add new lines above)

We should explore if `view.change_count()` fulfills our needs.

kaste:
Candidates for this check: change_count(), just the (source) code, or a md5/sha hash of the source code.

FichteFoll:
Is change_count view- or buffer-specific?

kaste:
buffer-specific

From Sublime docs

> Returns the current change count. Each time the buffer is modified, the change count is incremented. The change count can be used to determine if the buffer has changed since the last it was inspected.

PR #777, opened 2018-01-08
Refactor `Linter.lint_view`

kaste:
This is basically a refactor to make our main `Linter.lint_view` idempotent and reduce the mutable data structure. After this it should be basically okay, to run `lint_view` concurrently with the same view. (So the `queue.running` locks could go away.)

- Removed `self.code`, `self.highlight`, and `self.errors`
- Really simplify Highlight
- Bc we use a flat data structure, merging the results, is a simple list.extend

TODO/MAYBES

- [x] Name bike-shedding and comments
- [x] Move highlights creation up out of Linter.lint_view
- [x] Change type `highlight.Highlight(code)` ->`Highlight(view)` - As long as we don't plan to somehow keep a *virtual view* up-to-date in the sense that we update/invalidate error lines and columns while the user is typing, we can just use a real view to compute region boundaries
- [x] Remove `LinterStyleStore` from linter bc it belongs to highlight. Do we use the *style* somewhere else? 
- [x] ~~Maybe simplify `lint_view` if we find it's too long and unreadable~~ Implemented, but goes into another PR
- [x] ~~Use the new raw errors list, linting now produces. For compatibility, I convert the flat list to a struct grouped by line and error_type. But it is really trivial to get e.g. the `we_count` from the flat list.~~ Out of scope. But I used it for the goto commands, yet to PR
- [x] ~~Concurrently run the linters inside `lint_view`. E.g. run `flake8`and `mypy` in parallel~~ Implemented but becomes a new PR
- [x] ~~Use one instance per linter, not one per view~~ Out of scope, we have a `self.view`, which is used by `get_view_settings()` which is used in user-land

### DIFF
--- a/lint/highlight.py
+++ b/lint/highlight.py
@@ -41,42 +41,6 @@ def clear_view(view):
         view.erase_regions(key)
 
 
-class VirtualView:
-    def __init__(self, code=''):
-        self._code = code
-        self._newlines = newlines = [0]
-        last = -1
-
-        while True:
-            last = code.find('\n', last + 1)
-
-            if last == -1:
-                break
-
-            newlines.append(last + 1)
-
-        newlines.append(len(code))
-
-    def full_line(self, line):
-        """Return the start/end character positions for the given line."""
-        start = self._newlines[line]
-        end = self._newlines[min(line + 1, len(self._newlines) - 1)]
-
-        return start, end
-
-    def select_line(self, line):
-        """Return code for the given line."""
-        start, end = self.full_line(line)
-        return self._code[start:end]
-
-    # Actual Sublime API would look like:
-    # def full_line(self, region)
-    # def full_line(self, point) => Region
-    # def substr(self, region)
-    # def text_point(self, row, col) => Point
-    # def rowcol(self, point) => (row, col)
-
-
 def get_line_start(view, line):
     return view.text_point(line, 0)
 
--- a/lint/highlight.py
+++ b/lint/highlight.py
@@ -212,6 +212,12 @@ class Highlight:
 
         pos += start
         region = sublime.Region(pos, pos + length)
+
+        self.add_mark(error_type, style, region)
+
+        return length
+
+    def add_mark(self, error_type, style, region):
         other_type = ERROR if error_type == WARNING else WARNING
 
         for scope, marks in self.marks[other_type].items():
@@ -219,13 +225,13 @@ class Highlight:
             for i, mark in enumerate(marks):
                 if (mark.a, mark.b) == (region.a, region.b):
                     if error_type == WARNING:
-                        return length
+                        # ABORT! We found an error on the exact same position
+                        return
                     else:
                         self.marks[other_type][scope].pop(i - i_offset)
                         i_offset += 1
 
         self.marks[error_type].setdefault(style, []).append(region)
-        return length
 
     def near(self, line, near, error_type=ERROR, word_re=None, style=None):
         """
--- a/lint/highlight.py
+++ b/lint/highlight.py
@@ -80,7 +80,8 @@ class Highlight:
     """This class maintains error marks and knows how to draw them."""
 
     def __init__(self, code=''):
-        self.code = code
+        self.vv = VirtualView(code)
+
         # Dict[error_type, Dict[style, List[region]]]
         self.marks = defaultdict(lambda: defaultdict(list))
         self.style_store = HighlightStyleStore()
@@ -89,28 +90,6 @@ class Highlight:
         # lines to mark in the gutter.
         # Dict[error_type, Dict[lineno, style]]
         self.lines = defaultdict(dict)
-
-        # Linting runs asynchronously on a snapshot of the code.
-        # Marks are added to the code during that asynchronous linting,
-        # and the markup code needs to calculate character positions given
-        # a line + column. By the time marks are added, the actual buffer
-        # may have changed, so we can't reliably use the plugin API to
-        # calculate character positions.
-        # The solution is to calculate and store the character positions for
-        # every line when this object is created, then reference that
-        # when needed.
-        self.newlines = newlines = [0]
-        last = -1
-
-        while True:
-            last = code.find('\n', last + 1)
-
-            if last == -1:
-                break
-
-            newlines.append(last + 1)
-
-        newlines.append(len(code))
 
     @staticmethod
     def strip_quotes(text):
@@ -122,18 +101,6 @@ class Highlight:
 
         return text
 
-    def full_line(self, line):
-        """
-        Return the start/end character positions for the given line.
-
-        This returns *real* character positions (relative to the beginning
-        of self.code) base on the *virtual* line number.
-        """
-        start = self.newlines[line]
-        end = self.newlines[min(line + 1, len(self.newlines) - 1)]
-
-        return start, end
-
     def range(self, line, pos, length=-1, near=None, error_type=ERROR, word_re=None, style=None):
         """Mark a range of text."""
         raise Exception('`range` has been removed')
@@ -143,9 +110,8 @@ class Highlight:
         raise Exception('`near` has been removed')
 
     def add_error(self, line, start, end, error_type, style):
-        a, b = self.full_line(line)
+        a, b = self.vv.full_line(line)
         region = sublime.Region(a + start, a + end)
-        print('new: ', region)
         self.add_mark(error_type, style, region)
         self.line(line, error_type, style)
         return region
@@ -232,7 +198,7 @@ class Highlight:
             for line, style in self.lines[error_type].items():
                 if not self.style_store.has_style(style):
                     continue
-                pos = self.newlines[line]
+                pos, _ = self.vv.full_line(line)
                 region = sublime.Region(pos, pos)
                 gutter_regions.setdefault(style, []).append(region)
 
--- a/lint/highlight.py
+++ b/lint/highlight.py
@@ -76,11 +76,15 @@ class VirtualView:
     # def rowcol(self, point) => (row, col)
 
 
+def get_line_start(view, line):
+    return view.text_point(line, 0)
+
+
 class Highlight:
     """This class maintains error marks and knows how to draw them."""
 
-    def __init__(self, code=''):
-        self.vv = VirtualView(code)
+    def __init__(self, view):
+        self.view = view
 
         # Dict[error_type, Dict[style, List[region]]]
         self.marks = defaultdict(lambda: defaultdict(list))
@@ -100,7 +104,7 @@ class Highlight:
         raise Exception('`near` has been removed')
 
     def add_error(self, line, start, end, error_type, style, **kwargs):
-        line_start, _ = self.vv.full_line(line)
+        line_start = get_line_start(self.view, line)
         region = sublime.Region(line_start + start, line_start + end)
         self.add_mark(error_type, style, region)
         self.line(line, error_type, style)
@@ -188,7 +192,7 @@ class Highlight:
             for line, style in self.lines[error_type].items():
                 if not self.style_store.has_style(style):
                     continue
-                pos, _ = self.vv.full_line(line)
+                pos = get_line_start(self.view, line)
                 region = sublime.Region(pos, pos)
                 gutter_regions.setdefault(style, []).append(region)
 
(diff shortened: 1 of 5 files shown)


kaste:
That worked out pretty well. Decoupled linter.py from highlight.py.

But if I follow that path, this PR just gets too big. It is already not reviewable IMO.

braver:
I use a bunch of linters and should be able to verify this works. I’m not much of a reviewer anyway... That said it is better to have multiple short cycles rather than one really big one, so if you’re at a point where we should test and review, let’s do that and perhaps do some more work later.

kaste:
The purpose of this PR is to decouple linter.py and highlight.py which I did today, and remove the mutable data structure stuff when we lint, which I did yesterday(?). So, this is it. Of course, you get my usual fixups as always.

kaste:
New behavior:

- If there is no `m.col` but `m.near` and we can't match `m.near` it highlights the whole line. Before it completely dropped that error. (Bug fix)
- New extension point for implementors: `reposition_match(line, col, match, virtual_view) -> (line, start, end)`. E.g. for `flake8` if I have an indentation error, currently SL highlights the next word. A linter could overwrite `resposition_match` to highlight the whitespace that actually yields the error. So it would match from the beginning of that line to `m.col` in that case. 

For the testers: I don't actually change linter behavior. You just have to look if the highlights and gutter marks are on the same or a reasonable position.

- Check if highlights are okay, if you lint a whole file vs. selector based linting (e.g. script in html)
- Check if highlights are okay if only `m.near` is given. (I wasn't able to produce errors with m.near! Please tell me how to do that.)  
- Check if highlights are okay, if no `m.col` is given, where it should highlight the whole line or just the first char on that line depending on `settings.get('no_column_highlights_line')`

braver:
> I wasn't able to produce errors with m.near! 

I don't even know what that is man! 

I did get an error

```
Exception in thread Thread-80:
  File "./python3.3/threading.py", line 901, in _bootstrap_inner
  File "./python3.3/threading.py", line 1142, in run
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/queue.py", line 35, in worker
    callback(vid, hit_time)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 226, in lint
    Linter.lint_view(view, hit_time, self.highlight)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 793, in lint_view
    errors = linter.lint(code, hit_time)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1155, in lint
    error = self.process_match(m, vv)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1223, in process_match
    line, start, end = self.reposition_match(m.line, col, m, vv)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1297, in reposition_match
    match = self.word_re.search(text)
AttributeError: 'NoneType' object has no attribute 'search'

```

when linting some javascript with eslint:

```js
(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
```

The eslint plugin is probably also a bit buggy, but this did work before.


I got another one when linting sublime_linter.py:

```
Exception in thread Thread-4:
  File "./python3.3/threading.py", line 901, in _bootstrap_inner
  File "./python3.3/threading.py", line 1142, in run
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/queue.py", line 35, in worker
    callback(vid, hit_time)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 226, in lint
    Linter.lint_view(view, hit_time, self.highlight)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 793, in lint_view
    errors = linter.lint(code, hit_time)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1147, in lint
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1172, in find_errors
    yield self.split_match(error)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter-pylint/linter.py", line 274, in split_match
    match, line, col, error, warning, message, near = super().split_match(match)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1209, in split_match
    return LintMatch(**match_dict)
TypeError: __new__() got an unexpected keyword argument 'code'
```

The error goes away when I disable the pylint plugin (I think by now I run all the python linters I can find...), which does a lot of very funky processing so perhaps also something particular to this linter.

kaste:
Is that really my branch? I actually tested and developed this with `eslint` running. And I did not touch `split_match` and `find_errors` at all. Also, the lines in the tracebacks don't match: `return LintMatch(**match_dict)` is on line 1203

braver:
It’s your branch merged into master. I’ll check again though just to be sure.

kaste:
I remember we had a patch in `split_match` https://github.com/SublimeLinter/SublimeLinter/pull/761/files
It says merged but I don't have it.

kaste:
And the first error shouldn't be there either bc of that change https://github.com/SublimeLinter/SublimeLinter/pull/777/commits/ed8ce1ba73ebd2c7aa02e7dc569e429ab0974f29

T.i. `self.word_re` shouldn't be None it's defined on the base class.

kaste:
Okay, I fucked up the split_match patch

braver:
Yeah, I'm pretty confident I'm looking at the correct code here.

kaste:
Okay, I pushed the `split_match` fix.

But I cannot reproduce the eslint error.

kaste:
Can you do a find all for `word_re` bc if it's defined on the base class, where does it get None'd. I don't have any references to `word_re` within SublimeLinter, just in linter.py:

image

braver:
Yeah, just that one in linter.py.

Just found out I also have the jshint plugin running and actually that's the one that breaks. That plugin is also a bit of a mess. It has this:

```py
 def split_match(self, match):
        """
        Return the components of the match.

        We override this to catch linter error messages and return more presise
        info used for highlighting.

        """
        # restore word regex to default each iteration
        self.word_re = None

```

kaste:
WDYT, should be fixed over there? I actually wanted to add a fixup here, so that when `word_re` is set to None, you just opt-out of this functionality. Currently, you cannot opt-out bc when you set None, a default kicks in.

So, here jshint interprets None as opt-in to a default.

This is like monkey-patching, what they do. Usually you would store the old value in a safe place, and restore it later. They just saw, ah it's None, we use that. 

So **I** actually wanted this line to be 

image

but I forgot about this.

kaste:
And if we patch jshint, it would look like

image

kaste:
Trivial stuff.

kaste:
Ah, but jshint uses the `near` stuff like they're crazy. So you can quickly check if that `near` stuff works. 👍

kaste:
Okay, pushed again to allow setting `word_re` to None.

braver:
> WDYT, should be fixed over there

We could definitely change the jshint plugin, it’s a bit wild IMO.

braver:
Check, works again with JSHint 👍

kaste:
😌

braver:
Also, all the jshint stuff seems to be correctly highlighted and positioned as far as I can tell.

javascript inside html is off by the indentation (e.g. script that was indented 10 spaces has highlighting off by -10 characters). But this isn't a regression in this PR, master has it too and I bet SL3 isn't different either.

screen shot 2018-01-10 at 22 56 45

I'll see if I run into anything tomorrow, or if a code review results in anything, otherwise I'm about to merge this.

lint/highlight.py
```diff
@@ -40,304 +37,78 @@ def clear_view(view):
...
-            text = text[1:-1]
-
-        return text
-
-    def full_line(self, line):
-        """
-        Return the start/end character positions for the given line.
+    def range(self, line, pos, length=-1, near=None, error_type=ERROR, word_re=None, style=None):
+        """Mark a range of text."""
+        raise Exception('`range` has been removed')
```
braver: Do we need to keep these around?

FichteFoll: I guess otherwise any attempts to call it would just fail with an AttributeError.

FichteFoll: The proper way would probably be to raise a DeprecationWarning, but I'm not sure whether these are ignored by default

braver: If it's a deprecation, I should be able to check if any linters still call it and then notify the respective repos. Otherwise just failing with an error is just fine with me, no need to be fancy about this.

braver: Ok, not sure if I'm searching correctly, but:

- `highlight.near` has no results
- `highlight.range`:
  - [SublimeLinter-contrib-jsdebugmarkers](https://github.com/LeoLeal/SublimeLinter-jsdebugmarkers/blob/master/linter.py#L99)
  - [SublimeLinter-annotations](https://github.com/SublimeLinter/SublimeLinter-annotations/blob/master/linter.py#L99)
  - SublimeLinter-rst (has been deprecated)
- `self.error`:
  - Again SublimeLinter-contrib-jsdebugmarkers, SublimeLinter-annotations (and SublimeLinter-rst), one line down from `highlight.range`.

So it seems we should update SublimeLinter-annotations and give SublimeLinter-contrib-jsdebugmarkers a notice. @kaste could you perhaps update SublimeLinter-annotations?

kaste: I couldn't throw a DeprecationWarning bc I already removed it 🍭 

Freudian slip, I knew that annotations used this API - @braver sent me a list of API usage before. But this plugin didn't work on master already. This is just a result of not using the official API but fiddling around. jsdebugmarkers is a fork of annotations. Both reimplement `split_match` within `run` just for fun. I think `run` returns `None`, so SublimeLinter thinks anyway that this lint was clean and does a fast exit. 

Fuck it; I'm really bad at regexes.

kaste:
Basically at https://github.com/SublimeLinter/SublimeLinter-annotations/blob/master/linter.py#L97 

```output.append('line {}, col {}: {}: {}'.format(i, col, error_type, message))```

of course `append` unconditionally. Now you have to write a `regex` for `split_match` that parses this fake linter message back at https://github.com/SublimeLinter/SublimeLinter-annotations/blob/master/linter.py#L24. I don't know how regexes go `r'`line \d, col \d ...'` ❓ 

At the end of run just return `''.join(output)` and you're done.

braver:
Alright, I'll see how far I get with that. 👍

braver:
Jep, got it. Thanks!

kaste:
Okay, you know your regexes.

braver:
You do a few syntaxes and you'll learn quickly enough ;)

FichteFoll:
Shouldn't there be an API to skip having to emit fake output and parse that with regular expressions again? I'm not on top of the current API but if there isn't, this should be possible.

braver:
I kinda like that it pretends to be a command line utility just like pretty much every other linter. There aren’t many other linters that aren’t a util with output that we parse, and there shouldn’t be.

kaste:
Well, the API. We have a lot of surface area here. 

``` run(cmd, code) -> output ```

And as you can see annotations doesn't even use cmd (it should be `None`).

Since this refactoring:

``` lint(code) -> [errors] ```

So this is actually the right candidate for this, you only have to see how the error-dict looks like. Before this PR this didn't work. So, you're basically right @FichteFoll, I just so used to look at the old code, that I forgot that I just implemented that. (Currently the signature is actually `lint(code, hit_time)`, but we have this #758, so `hit_time` might go.)

Issue #729, opened 2017-12-20
Run subprocess.Popen with universal_newlines and encoding parameters


FichteFoll:
Running `subprocess.Popen` with the `universal_newlines` and specifying the enconding instead of manually decoding everywhere (basically everywhere that `util.run_shell_command` is run and in `util.combine_output`) is much cleaner and centralizes the point of concern regarding encoding issues and CR presence (see #703).

This would probably need some testing, but would be the most proper way to handle the situation.

FichteFoll:
I agree, this is quite sub-optimal since it utilizes a fallback to `locale.getpreferredencoding()` instead of utilizing "the right one". That makes the entire situation a lot more complicated.

Maybe the better way for now is to just do newline handling inside  `combine_output` as well and change its signature to accept two (or multiplek `bytes` buffers, where it would only concatenate those if provided. Then use that function within `run_shell_command` as well.

kaste:
I just doubt that `javac` still has utf-8 issues on windows. Couldn't google anything either.

kaste:
Okay, javac doesn't seem to handle utf-8 as I would expect it.

![image](https://user-images.githubusercontent.com/8558/34302544-b1c97fe4-e731-11e7-894f-9bcd9375dff5.png)

kaste:
Okay, asked on stackoverflow https://stackoverflow.com/questions/47943780/javac-does-not-output-unicode-on-command-line. You actually have to pass an argument to get this working:

```
javac -J-Dfile.encoding=UTF-8 <your_file.java>
```

![image](https://user-images.githubusercontent.com/8558/34311111-dc3ac628-e75b-11e7-936c-c29dc0feff8e.png)

kaste:
So, this would mean, instead of

```
return (self.executable_path, xlint, '-encoding', 'UTF8', '*')
return (self.executable_path, xlint, '-J-Dfile.encoding=UTF-8', '*')
``` 
https://github.com/SublimeLinter/SublimeLinter-javac/blob/15b921f74de1bdb4146ef7d0771d5fa120f998d3/linter.py#L57

/cc @braver Do we know enough Java people to propose this change?

braver:
My neighbour uses Java. I don't know anyone else in the world that does. 🤷‍♂️

braver:
closed via #783

PR #783, opened 2018-01-10
Remove newline characters in panel view

kaste:
I don't find the issue right now, but it's about 

image

After:

image

This is completely the KISS approach.

### DIFF
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -112,8 +112,10 @@ def format_header(f_path):
 def format_row(lineno, error_type, dic):
     lineno = int(lineno) + 1
     start = dic['start'] + 1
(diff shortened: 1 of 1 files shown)


FichteFoll:
This should be tackled at the root and not to cover up the symptoms: https://github.com/SublimeLinter/SublimeLinter/issues/729

kaste:
Totally agree, but this is a fix for a visual bug. A obvious visual bug is a release blocker. #729 is correct, but didn't move forward for three weeks. 

Okay, I give #729 15 minutes and we will see

kaste:
Okay after 15 minutes I was interested enough, then it took 30 minutes.

kaste:
Okay, #784 is here to test.

kaste:
`universal_newlines` doesn't work on python 3.3. PTAL @FichteFoll and see comment in #784 

#729 can then be closed.

PR #784, opened 2018-01-11
Use universal newlines

### LINKED ISSUE #729: Run subprocess.Popen with universal_newlines and encoding parameters

kaste:
fixes #729 

Also, move two functions from util to ruby_linter

### DIFF
--- a/lint/base_linter/ruby_linter.py
+++ b/lint/base_linter/ruby_linter.py
@@ -4,8 +4,11 @@ import os
 import re
 import shlex
 import sublime
+import subprocess
 
-from .. import linter, util
+from functools import lru_cache
+
+from .. import linter, util, persist
 
 CMD_RE = re.compile(r'(?P<gem>.+?)@ruby')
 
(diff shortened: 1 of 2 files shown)


kaste:
`universal_newlines` doesn't work. On Python 3.3, we cannot set an `encoding` using `Popen`, and it just uses `locale.getpreferredencoding(False)`. On 3.6 this approach would work. *Changed in version 3.6: Added encoding and errors parameters* https://docs.python.org/3/library/subprocess.html 

So you can take the first commit here for the refactoring. But then the rest is 💩 and #783 is 👑

FichteFoll:
Oh, encoding was only added for 3.6? That's quite late. I don't remember it being that way, but I also won't bother checking on mobile.

FichteFoll:
Well the docs seem to indicate this being true. Guess we'd need a manual implementation then.

Issue #782, opened 2018-01-10
Not showing any error


wellyal:
I wasnt able to make it work. I followed the instructions on the [documentation](http://sublimelinter.readthedocs.io/en/latest/installation.html) installed through package install, restarted sublime but it does not show the lint errors. On console loads everything ok like:

System: Mac OS EL Capitan v10.11.6
Sublime 3

```
reloading plugin SublimeLinter.commands
reloading plugin SublimeLinter.sublimelinter
plugins loaded
```

eslint working just fine though
![image](https://user-images.githubusercontent.com/3047016/34778397-62b68d2a-f604-11e7-8374-27123f60db0f.png)

FichteFoll:
You should enable debug logging for SublimeLinter, restart ST and post the entire console's contents here.

wellyal:
```
...
reloading plugin SublimeLinter.commands
reloading plugin SublimeLinter.sublimelinter
plugins loaded
SublimeLinter: debug mode: on 
SublimeLinter: temp directory: /var/folders/r5/mq06kxp92njczjyb1qv56qzc0000gp/T/SublimeLinter3-wellyngton.leitao 
Package Control: Skipping automatic upgrade, last run at 2018-01-10 12:34:22, next run at 2018-01-10 13:34:22 or after
environment variables loaded using: /bin/zsh -l
```

thezimmee:
SublimeLinter also broke for me when I upgraded to `v4.0.0-rc.9.4`, where it was working prior to the upgrade (with the exception of known issues like the inline diff, .sublime-color-theme, etc., all which I believe `v4.0.0-rc.9.4` is supposed to address). Now the only linter that "kinda" works is annotations. No errors show for other linters; annotations shows the error but doesn't highlight anything as it used to. I saw this error in the console:

```
Traceback (most recent call last):
  File "/Volumes/Home/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/util.py", line 264, in generate_color_scheme_async
  File "./python3.3/xml/etree/ElementTree.py", line 1356, in XML
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 1, column 0
```

And this is the console output when starting up:

```
startup, version: 3156 osx x64 channel: dev
executable: /Applications/Sublime Text.app/Contents/MacOS/Sublime Text
working dir: /
packages path: /Volumes/Home/Library/Application Support/Sublime Text 3/Packages
[...]

PR #786, opened 2018-01-11
Hotfix: Do not cache result of `create_environment`

kaste:
As discussed #782

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -235,16 +235,11 @@ def get_environment_variable(name):
     return value
 
 
-@lru_cache(maxsize=None)
 def create_environment():
-    """
-    Return a dict with os.environ augmented with a better PATH.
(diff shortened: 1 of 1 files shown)


FichteFoll:
That's an interesting way to only print the path when it changes.

braver:
eslint was pretty reliably failing because it would start before my PATH was corrected. Now it works every time. 

🎉 🕺 

Where do I send the flowers?

kaste:
@FichteFoll pending patent on that trick. 

Most people think of memoize functions if they want to be more efficient about computing a result. But in fact, you can control and reduce side-effects. So if you have a fn `store_user(db, user_name) -> None` and you decorate it with a memoize fn like lrucache or so, you can call it without an `if` statement, unconditionally, as often as you want. It will only do the side-effect, that is actually hit the db or the file, if one of the args changed.

🔚

PR #788, opened 2018-01-11
Goto commands

### LINKED ISSUE #715: [beta] Add command Goto next/previous from current position
Key command for goto next/previous error is broken and really a regression from SL3. 

- As a user, I want to go to the next/previous error *from my current position*.
- ~~If I never opened the new panel the keybindings do not work.~~
- After an edit, hitting e.g. F4 start the list from the beginning again, which can be in any file.

[edit: removed separate issues listed elsewhere, tweaked tone to be a bit more inviting for contributions ]

kaste:
Fixes #715

### DIFF
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -104,30 +104,10 @@ def goto(view, direction, count, wrap):
 
 
 def move_to(view, line, col):
-    loc = view.text_point(line, col)
-    region = sublime.Region(loc, loc)
-    view.sel().clear()
-    view.sel().add(region)
-    center_region_in_view(view, region)
-
-
-def center_region_in_view(view, region):
-    """Center the given region in view.
-
-    There is a bug in ST3 that prevents a selection change
-    from being drawn when a quick panel is open unless the
-    viewport moves. So we get the current viewport position,
-    move it down 1.0, center the region, see if the viewport
-    moved, and if not, move it up 1.0 and center again.
-    """
-    x1, y1 = view.viewport_position()
-    view.set_viewport_position((x1, y1 + 1.0))
-    view.show_at_center(region)
-    x2, y2 = view.viewport_position()
-
-    if y2 == y1:
-        view.set_viewport_position((x1, y1 - 1.0))
-        view.show_at_center(region)
+    window = view.window()
+    filename = view.file_name() or "<untitled {}>".format(view.buffer_id())
+    target = "{}:{}:{}".format(filename, line + 1, col + 1)
+    window.open_file(target, sublime.ENCODED_POSITION)
 
 
 def get_current_pos(view):
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -80,6 +80,8 @@ def goto(view, direction, count, wrap):
 
     reverse = direction == 'previous'
     jump_positions = list(previous_positions if reverse else next_positions)
+    if reverse:
+        jump_positions = list(reversed(jump_positions))
 
     if not jump_positions:
         if wrap:
@@ -91,14 +93,11 @@ def goto(view, direction, count, wrap):
             )
             return
     elif len(jump_positions) <= count:
-        # If we cannot jump wide enough, pin to top/bottom to reduce
-        # disorientation.
-        error = errors[0] if reverse else errors[-1]
+        # If we cannot jump wide enough, do not wrap, but jump as wide as
+        # possible to reduce disorientation.
+        error = jump_positions[-1]
     else:
-        # This is a bit hard to read, but note that for any list we get the
-        # *first* item via `list[0]`, but going reverse, we get the *last*
-        # item via `list[-1]` (not via `list[-0]`).
-        error = jump_positions[-count] if reverse else jump_positions[count - 1]
+        error = jump_positions[count - 1]
 
     line, start = error
     move_to(view, line, start)
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -68,11 +68,11 @@ def goto(view, direction, count, wrap):
             # filtered errors directly under cursor (!)
             line == current_line and start < current_col)
 
-    next = dropwhile(before_current_pos, errors)
-    previous = takewhile(before_current_pos, errors)
+    next_positions = dropwhile(before_current_pos, errors)
+    previous_positions = takewhile(before_current_pos, errors)
 
     reverse = direction == 'previous'
-    jump_positions = list(previous if reverse else next)
+    jump_positions = list(previous_positions if reverse else next_positions)
 
     if not jump_positions:
         if wrap:
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -60,6 +60,13 @@ def goto(view, direction, count, wrap):
                      error['start'] <= current_col <= error['end'])}
     errors = sorted(errors)
 
+    # Edge case: Since we filtered, it is possible we get here with nothing
+    # left. That is the case if we sit on the last remaining error, where we
+    # don't have anything to jump to and even `wrap` becomes a no-op.
+    if len(errors) == 0:
+        flash(view, 'No more problems.')
+        return
+
     def before_current_pos(error):
         line, start = error
         return (
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -1,0 +1,136 @@
+import sublime
+import sublime_plugin
+
+from itertools import dropwhile, takewhile
+from .lint import persist
+
+
+"""
+Implement typical Goto Next Previous Error Commands. If you will, add
+the following to your key bindings.
+
+```
+  { "keys": ["ctrl+k", "n"],
+    "command": "sublime_linter_goto_error",
+    "args": {"direction": "next"}
+  },
+  { "keys": ["ctrl+k", "p"],
+    "command": "sublime_linter_goto_error",
+    "args": {"direction": "previous"}
+  },
+```
+
+Supported args:
+
+* `direction`: 'previous' | 'next'
+* `count`: How wide you will jump. Defaults to 1.
+* `wrap`: Set to True, to jump to the top if you're on the last error and
+          vice versa. Defaults to False
+
+"""
+
+
+class SublimeLinterGotoError(sublime_plugin.WindowCommand):
+    def run(self, direction='next', count=1, wrap=False):
+        goto(self.window.active_view(), direction, count, wrap)
+
+
+def goto(view, direction, count, wrap):
+    bid = view.buffer_id()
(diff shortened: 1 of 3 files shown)


kaste:
Already force pushed a fixup. It's really always the same.

goto_commands.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def before_current_pos(error):
+        line, start = error
+        return (
+            line < current_line or
+            # `start < current_col` is only safe bc we already
+            # filtered errors directly under cursor (!)
+            line == current_line and start < current_col)
+
+    next = dropwhile(before_current_pos, errors)
```
FichteFoll: `next` is a built-in function.

goto_commands.py
```diff
@@ -0,0 +1,136 @@
...
+    region = sublime.Region(loc, loc)
+    view.sel().clear()
+    view.sel().add(region)
+    center_region_in_view(view, region)
+
+
+def center_region_in_view(view, region):
+    """Center the given region in view.
+
+    There is a bug in ST3 that prevents a selection change
```
FichteFoll: It seems the currently preferred way for this is to perform selection within in a text command so they can be soft-undone.

goto_commands.py
```diff
@@ -0,0 +1,136 @@
...
+  },
+```
+
+Supported args:
+
+* `direction`: 'previous' | 'next'
+* `count`: How wide you will jump. Defaults to 1.
+* `wrap`: Set to True, to jump to the top if you're on the last error and
+          vice versa. Defaults to False
+
```
FichteFoll: I can't jump to the next file with this if I want to, can I?

braver: Doesn't seem like it no.

goto_commands.py
```diff
@@ -0,0 +1,136 @@
...
+  { "keys": ["ctrl+k", "p"],
+    "command": "sublime_linter_goto_error",
+    "args": {"direction": "previous"}
+  },
+```
+
+Supported args:
+
+* `direction`: 'previous' | 'next'
+* `count`: How wide you will jump. Defaults to 1.
```
braver: What's the idea behind this? Is that so, perhaps from the command palette, you can say "jump 5 errors down"? Or is it just for the vim-style chords for neovintageous?

sublime_linter.py
```diff
@@ -241,6 +241,10 @@ def highlight(self, view, errors, hit_time):
         if hit_time and persist.last_hit_times.get(vid, 0) > hit_time:
             return
 
+        bid = view.buffer_id()
+        persist.raw_errors[bid] = errors
+
+        # For compatibility we store the errors SL3 style as well.
```
braver: Compatibility with what exactly?

sublime_linter.py
```diff
@@ -251,8 +255,9 @@ def highlight(self, view, errors, hit_time):
             vid = view.id()
             persist.errors[vid] = errors_by_line
 
-        events.broadcast(events.FINISHED_LINTING, {'buffer_id': view.buffer_id()})
+        events.broadcast(events.FINISHED_LINTING, {'buffer_id': bid})
 
+        # TODO: These should go to their own modules.
```
braver: I'd prefer it if the TODO was done, or if it were an issue so we can track it.

goto_commands.py
```diff
@@ -0,0 +1,136 @@
...
+            error = errors[-1] if reverse else errors[0]
+        else:
+            flash(
+                view,
+                'No more problems {}.'.format('above' if reverse else 'below')
+            )
+            return
+    elif len(jump_positions) <= count:
+        # If we cannot jump wide enough, pin to top/bottom to reduce
+        # disorientation.
```
braver: Awesome 👾

goto_commands.py
```diff
@@ -0,0 +1,136 @@
...
+    "command": "sublime_linter_goto_error",
+    "args": {"direction": "previous"}
+  },
+```
+
+Supported args:
+
+* `direction`: 'previous' | 'next'
+* `count`: How wide you will jump. Defaults to 1.
+* `wrap`: Set to True, to jump to the top if you're on the last error and
```
braver: Perhaps we need to flash a message on the status bar that says we wrapped?

goto_commands.py
kaste: That was not specced. You can do this with `F4`, maybe it could be done, that `F4` immediately jumps to the first error of the next file.

The problem with view jumping is, that we yet don't have an ordered, stable list of views to jump to. That's also a problem of the panel. I would prefer if the panel showed the views in a lru way. But I file an issue about that.

goto_commands.py
kaste: I don't know how to do that. I thought about this, but failed bc there is no documentation of what commands Sublime provides. 

You could fix this in a separate PR.

sublime_linter.py
kaste: We have the new style, flat errors on `raw_errors` which have the type `[buffer_id, [errors]]`. On `persist.errors` we store `[view_id, [line, [error_type, [errors]]`. 

This is merely a safe refactoring strategy. We can slowly change the consumers to use `raw_errors`, and if nobody uses `errors` anymore, we flip names.

goto_commands.py
kaste: I think I change that.

sublime_linter.py
kaste: Then I open an issue.

goto_commands.py
kaste: I don't think so, bc we already have the visual feedback. So the user has the immediate feedback that the command has been received. 

With wrap `false`, I flash bc the user cannot decide if the computer actually received the command bc nothing happens otherwise.

goto_commands.py
kaste: Yes, you could do that. 

I merely implemented that bc it was specced, and I immediately thought that it is worth the extra effort.

goto_commands.py
braver: Indeed, I'd like this to *not* jump to another file. You can use the panel. for that. So as it is now is excellent IMO.

goto_commands.py
braver: Sounds good, agreed.

kaste:
Done? Working as intended?

sublime_linter.py
braver: Should we have todo to do this refactoring?

kaste: Dunno. I actually like `TODO` and comments in the source. It's open source so a reader sees that there is something that can be done.  

I actually wanted to commit more (welcoming) comments about fishy code, bc there are parts in this software ... really, really hard to read, really not welcoming to do a PR and fix a problem.

kaste: The code sometimes says: Please, go away, leave me alone.

goto_commands.py
kaste: Rechecked, I can jump back via `alt-left` just fine.

sublime_linter.py
braver: Sure, “leave me alone” is perfectly ok to have in a comment. But if it’s something that really should be be done, ie. it’s technical debt, I want to have a single list of issues that I can manage. Todo comments are not in that list, and have a lower likelihood of getting done.

kaste: No, I meant, we have a lot of actual code that says "Leave me alone", not comments. 😄 

Anyway, IMO this is a good comment. You to decide if it should go away.

braver: oh, aha, yes we have that too 😄 

I'm very much 🤷‍♂️ right now, so yeah. I'll just merge this. :)

goto_commands.py
FichteFoll: I think you'd need your own command for this.

goto_commands.py
FichteFoll: Well, I could use F4, but the selection within the panel is not updated if I use the commands added here, so I can only use either entirely (or fall back to using the mouse). I guess the selection updating stuff could be inherited from #752, but that doesn't need to happen right here in this PR.

goto_commands.py
FichteFoll: Otoh, you could try View.open_file with `sublime.ENCODED_POSITION`.

pykong:
#788 is another great one by @kaste 
Implementing our own GoTo function gives us more control.

Still, I agree with @FichteFoll that the panel selection should be updated.
The panel is one of the defining features of SL4 and we should really make it shine.

The respective function to update the panel from #752 can be easily interlinked with any other code, as long as a `view` can be provided as an argument.

goto_commands.py
kaste: Okay, got it

kaste:
You cannot stop me, Sublime. I know my Internet Explorer.

goto_commands.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def move_to(view, line, col):
+    window = view.window()
+    filename = view.file_name() or "<untitled {}>".format(view.buffer_id())
+    target = "{}:{}:{}".format(filename, line + 1, col + 1)
+    window.open_file(target, sublime.ENCODED_POSITION)
+
+
+def get_current_pos(view):
```
pykong: There is the method `get_line_and_col` in `sublime_linter.py` that is very similar. 
https://github.com/SublimeLinter/SublimeLinter/blob/e14e93bc666c044419f0b70e7e7ea62a681d70e6/sublime_linter.py#L316
Maybe we could honor DRY here and refactor to `utils.py`?

goto_commands.py
```diff
@@ -0,0 +1,122 @@
...
+        line, start = error
+        return (
+            line < current_line or
+            # `start < current_col` is only safe bc we already
+            # filtered errors directly under cursor (!)
+            line == current_line and start < current_col)
+
+    next_positions = dropwhile(before_current_pos, errors)
+    previous_positions = takewhile(before_current_pos, errors)
+
```
pykong: Here we iterate over errors twice. What about slicing `errors` based on `next_positions`?

kaste: These are lazy. Both lines are declarative.

goto_commands.py
kaste: Self-contained overrules DRY. This way anyone can understand everything without looking here and there. I can just rewrite this fn and don't have to look if I break anyone. Etc. I think I actually copy-pasted this from my own `status_bar_view.py`. 🍭

braver: Especially for small utilities like this, not having it tangled up with usages elsewhere is probably best.

PR #789, opened 2018-01-11
Implement backend_async (as replacement for Linter.lint_view)

kaste:


FichteFoll:
Yay for refactoring. (Didn't review in detail.)

Is there a specific reason for the sync backend still being included? As a fallback? I mean, it's not used as far as I can tell.

Also, I wish I could react with other emojis than just 6, so have a 🔥 .

braver:
Look good man. I think we can remove the sync version then?

kaste:
Okay, I do this in a commit bc I want a backup for the sync backend.

kaste:
Done. This went out really readable, I think.

PR #797, opened 2018-01-13
sync panel selection with cursor

braver:
replaces #752 

> Ok, seems to work pretty good now. One thing: it doesn't update the selection in the panel as I switch files. Only after I move my cursor does it sync up. Everything else = 🌶 

@pykong please continue work on this feature in this branch and PR.

### DIFF
--- a/status_bar_view.py
+++ b/status_bar_view.py
@@ -47,7 +47,7 @@ class UpdateState(sublime_plugin.EventListener):
             'we_count': we_count,
             'current_pos': current_pos
         })
-
+        panel.update_panel_selection(**State)
         draw(**State)
 
     def on_selection_modified_async(self, _primary_view_):
--- a/status_bar_view.py
+++ b/status_bar_view.py
@@ -4,6 +4,7 @@ import sublime_plugin
 from .lint import persist
 from .lint.const import STATUS_KEY
 from .lint import events
+from .panel import panel
 
 
 State = {
@@ -56,7 +57,7 @@ class UpdateState(sublime_plugin.EventListener):
             State.update({
                 'current_pos': current_pos
             })
-
+            panel.update_panel_selection(**State)
             draw(**State)
 
 
--- a/commands.py
+++ b/commands.py
@@ -88,5 +88,4 @@ class SublimeLinterUpdatePanelCommand(sublime_plugin.TextCommand):
             if new_selected_region:
                 sel.add(new_selected_region)
                 return
-
         sel.add(0)
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -1,5 +1,6 @@
 import os
 import sublime
+import bisect
 
 from ..lint.const import WARN_ERR
 from ..lint import util, persist
@@ -19,6 +20,22 @@ OUTPUT_PANEL_SETTINGS = {
     "translate_tabs_to_spaces": False,
     "word_wrap": False
 }
+
+
+class RenderLines:
+    def __init__(self):
+        self.lines = []
+        self._line_counter = 0
+
+    def append(self, str):
+        self.lines.append(str)
+        self._line_counter += 1
+
+    def current_lineno(self):
+        return self._line_counter
+
+    def render(self):
+        return "\n".join(self.lines)
 
 
 def dedupe_views(errors):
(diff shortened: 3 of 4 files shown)


pykong:
@braver Good spot! You got an eye for those little things that matter.
Fixed it with f0df236

braver:
giphy

braver:
@kaste I’ll be honest here, and hopefully you already noticed the absence of serious review and feedback from me in earlier PR’s: I don’t have the knowledge and experience to guide the architecture. I can design the features and decide what stays and goes, do all the menial tasks that come with maintaining this repo and the plugins. I hope to support anyone who wants to contribute in that way and by testing if it works correctly. 

Basically I’m a designer (with basic knowledge of ~10 languages but expert only in CSS) with experience in guiding development teams and managing a product, including release management and community engagement. That seems to have worked out in making SL4 happen, but I did none of the coding for it.

I’m super grateful for your reviews and those of @FichteFoll for instance. I trust you guys, and @pykong, to really do your best and make SL4 awesome. I would love to grow the maintenance team and more formally review contributions, but I cannot assume you’re up for the responsibility. If anyone wants to have access to the repo and/or help reviewing, let me know. SL needs an architect, not just a designer, to help keep it stable in the future.

Now that you point it out I do see the flaw here. I totally welcome a PR that changes it. @pykong, do you agree and want to attempt a refactor for it?

status_bar_view.py
```diff
@@ -46,7 +47,7 @@ def on_activated_async(self, active_view):
             'we_count': we_count,
             'current_pos': current_pos
         })
-
+        panel.update_panel_selection(**State)
```
braver: Basically, these calls should be triggered from somewhere else.

pykong:
@kaste You are the one that did an awesome job refurbishing the internals of SublimeLinter. (Kudos!)
I think therefore you got a better grasp at the architecture of SublimeLinter, than @braver or myself.

Maybe I misunderstood your suggestion on the predecessor PR as an invitation to move that function call over.

Could you make a suggestion where to call the `panel.update_panel_selection` then? Or is there some more profound change you may suggest? _Importantly performance wise I got the impression it was running even smoother when called from `status_bar_view`!_ 

**As an additional note:** Now the panel selection is dynamically updated, I see a potential overlap of function with the newly implemented "old style" GoTo @kaste implemented. For example have we even considered to continue using <kbd>F4</kbd> /  <kbd>shift</kbd> + <kbd>F4</kbd>, instead of introducing two new commands? We could maybe use an `on_command` listener and intercept `"next_result"` and `"prev_result"`. If our panel is the active we can execute additional logic, required for the behaviour added in #788.

Right now we end up with two different GoTos. I am sure I am not the only one seeing some redundancy here. What can we do to make this harmonious?

braver:
> Right now we end up with two different GoTos. I am sure I am not the only one seeing some redundancy here. What can we do to make this harmonious?

Please open a new issue for this so we can discuss it.

braver:
About the refactor proposal: https://github.com/SublimeLinter/SublimeLinter/issues/800

kaste:
Well, first of all, the functionality here is just awesome. I did not expect that this would work out so well.

I also did not foresee that you could just reuse the *whole* controller from the status bar. For now, the 'correct' way is to just copy-paste all the bits you need from `status_bar_view.py` into say `panel_view.py`. 

So now, we obviously already see a reusable pattern here. But we can still wait, I can imagine that the panel view still grows and needs more and more `state` for that. The transition right now is more into KISS, not into DRY. Reduce coupling, reduce avoidable abstractions.

kaste:
Would be cool if you could remove these two lines as well 

https://github.com/SublimeLinter/SublimeLinter/blob/5ee12b9256605ff57115686b3c61876645de021b/sublime_linter.py#L269-L270

Because if you listen to the FINISHED_LINTING event, you already have a place to react. Also, this seems inefficient. We probably don't need to refresh all windows if one linter finishes.

braver:
I can totally get behind those proposals 👍🏻

pykong:
@kaste great you like the functionality!

Thanks for your guidance on the topic. I am going to refactor my code to incoorporate a `panel_view.py`.

As I am rather busy right now doing other things, it may eventually take me a couple days until I can attend to the topic.

PR #798, opened 2018-01-13
Blacklist filtering for panel 

pykong:
This PR replaces #754.
It extends the options for filtering panel output with blacklisting of `codes` and `linters`.

Instruction to make custom command with provided args for filtering:
```
    // Show all errors
    { "keys": ["ctrl+k", "a"], "command": "sublime_linter_panel_toggle" },
    // To display filtered results in the panel add arguments:
    // { "keys": ["ctrl+k", "a"], "command": "sublime_linter_panel_toggle", "args": {
    // whitelist via:
    //         "types": ["error"],
    //         "codes": [],
    //         "linter": [],
    // blacklist via:
    //          "exclude_codes": [],
    //          "exclude_linter": []
    //     }
    // }
```

### DIFF
--- a/keymaps/Default (OSX).sublime-keymap
+++ b/keymaps/Default (OSX).sublime-keymap
@@ -6,14 +6,9 @@
     // Show all errors
     { "keys": ["ctrl+super+a"], "command": "sublime_linter_panel_toggle" },
     // To display filtered results in the panel add arguments:
-<<<<<<< HEAD
-    // { "keys": ["ctrl+k", "a"], "command": "sublime_linter_panel_toggle", "args": {
-    // whitelist via:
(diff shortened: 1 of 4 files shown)


keymaps/Default (OSX).sublime-keymap
```diff
@@ -6,10 +6,19 @@
     // Show all errors
     { "keys": ["ctrl+super+a"], "command": "sublime_linter_panel_toggle" },
     // To display filtered results in the panel add arguments:
+<<<<<<< HEAD
```
braver: There are conflict markers left in this file!

braver:
Whoop whoop!

PR #799, opened 2018-01-13
Scope based syntax detection

pykong:
### Objective
As of now SL, uses the name of a syntax file for detecting a syntax.
That method lacks flexibility and also necessitates long syntax maps.
See: #740 #691 

This PR implements a new syntax selection using the base scope of a view instead. (Replaces #753)
The regarding steps are:

### Algorithm
0. A syntax map to reference internally usable aliases to a scope selectors.
Like ```"javascript": ["source.js"]```
1. The main scope of a view is retrieved. Examples:
  **a)** `source.python`
  **b)** `source.json.sublime-xxx`
  **c)** `text.html.markdown`
2. The base scope is iteratively split on `.` from right to left. If a resulting tail fragment is identicaly to an alias in the syntax map it is returned. (Like `markdown` from **c)**)
3. If 2. has not been successful in returning a syntax, now a function build around [score_selector](https://www.sublimetext.com/docs/3/api_reference.html) is employed, which tries to find the syntax alias owning the highest scoring scope selector.
4. As a final resort the last tail determined in 2) is returned. Like `python` from `source.python`.

Main parts of the code are derived from https://github.com/Snazzah/SublimeDiscordRP/pull/15

#### Limitations
*I like however to stress that the PR at hand represents my own interpretation of that linked code.
My understanding of Sublime's syntax system is limited at this point. Hence, I invite constructive reviews and critique.*

For example my hypothesis is that most base scopes look like `source.usable_alias` like **a)**. Hence
step 4) will give us `usable_alias` as the syntax.  Thereby we do not need to explicitly put `python` into our syntax map. *That can make the map shorter as we only need to put in entries for required cases.*
I could be wrong about this though. Maybe we need a much larger syntax map.

### TODO:
- test and review
- extension of the syntax map if required

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -42,30 +42,6 @@ def printf(*args):
     print()
 
 
-def get_syntax(view):
-    """
-    Return the view's syntax.
-
-    or the syntax it is mapped to in the "syntax_map" setting.
-    """
-    syntax_re = re.compile(r'(?i)/([^/]+)\.(?:tmLanguage|sublime-syntax)$')
-    view_syntax = view.settings().get('syntax', '')
-    mapped_syntax = ''
-
-    if view_syntax:
-        match = syntax_re.search(view_syntax)
-
-        if match:
-            view_syntax = match.group(1).lower()
-            from .persist import settings
-            mapped_syntax = settings.get(
-                'syntax_map', {}).get(view_syntax, '').lower()
-        else:
-            view_syntax = ''
-
-    return mapped_syntax or view_syntax
-
-
 class Borg:
     _shared_state = {}
 
@@ -610,3 +586,45 @@ def load_json(*segments, from_sl_dir=False):
     base_path = "Packages/SublimeLinter/" if from_sl_dir else ""
     full_path = base_path + "/".join(segments)
     return sublime.decode_value(sublime.load_resource(full_path))
+
+
+# syntax functions
+
+
+def yield_tail_scopes(scope):
+    while "." in scope:
+        scope, tail = scope.rsplit(".", 1)
+        yield tail
+
+
+def max_selector_score(view, selectors):
+    return max((view.score_selector(0, s) for s in selectors if s) or [0])
+
+
+def get_syntax_by_score(view, syntax_map):
+    try:
+        scoring = {
+            max_selector_score(view, selectors): syntax
+            for syntax, selectors in syntax_map.items() if selectors
+        }
+    except ValueError:
+        return
+    else:
+        max_score = max(scoring)
+        if max_score > 0:
+            return scoring[max_score]
+
+
+def get_syntax(view):
+    """Return the view's syntax."""
+    syntax_map = load_json('resources/syntax_map.json', from_sl_dir=True)
+    main_scope = view.scope_name(0).split()[0]
+
+    # lazy syntax lookup
+    for tail in yield_tail_scopes(main_scope):
+        if tail in syntax_map:
+            return tail
+
+    # more elaborate lookup via scoring mechanism
+    # or last tail as fallback, e.g. `python` from `source.python`
+    return get_syntax_by_score(view, syntax_map) or tail
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -118,17 +118,5 @@
                 "error"
             ]
         }
-    ],
-
-    // Maps variant syntaxes to the syntax that you have a linter for.
-    "syntax_map": {
-        "html (django)": "html",
-        "html (rails)": "html",
-        "html 5": "html",
-        "javascript (babel)": "javascript",
-        "magicpython": "python",
-        "php": "html",
-        "python django": "python",
-        "pythonimproved": "python"
-    }
+    ]
 }
--- a/resources/syntax_map.json
+++ b/resources/syntax_map.json
@@ -1,0 +1,7 @@
+{
+    "coffeescript": ["source.coffee"],
+    "javascript": ["source.js"],
+    "markdown": [],
+    "pug": ["text.jade", "source.pyjade"],
+    "typescript": ["source.ts"]
+}


lint/util.py
```diff
@@ -610,3 +586,45 @@ def load_json(*segments, from_sl_dir=False):
...
+
+
+def get_syntax(view):
+    """Return the view's syntax."""
+    syntax_map = load_json('resources/syntax_map.json', from_sl_dir=True)
+    main_scope = view.scope_name(0).split()[0]
+
+    # lazy syntax lookup
+    for tail in yield_tail_scopes(main_scope):
+        if tail in syntax_map:
```
FichteFoll: This will test the tailed scopes against the keys of the syntax map, but we want to check against if any of the values contains it, so this should be like follows, no?

```py
for syntax, scopes in syntax_map.items():
    if tail in scopes:
        return syntax
```

lint/util.py
```diff
@@ -610,3 +586,45 @@ def load_json(*segments, from_sl_dir=False):
     base_path = "Packages/SublimeLinter/" if from_sl_dir else ""
     full_path = base_path + "/".join(segments)
     return sublime.decode_value(sublime.load_resource(full_path))
+
+
+# syntax functions
+
+
+def yield_tail_scopes(scope):
```
FichteFoll: This function yields the following valus over `text.html.markdown`:
```
"markdown"
"html"
"source"
```

Shouldn't you be eliminating the tail and yield `scope` here? So the result becomes:
```
"source.html.markdown"
"source.html"
"source"
```

lint/util.py
```diff
@@ -610,3 +586,45 @@ def load_json(*segments, from_sl_dir=False):
...
+    main_scope = view.scope_name(0).split()[0]
+
+    # lazy syntax lookup
+    for tail in yield_tail_scopes(main_scope):
+        if tail in syntax_map:
+            return tail
+
+    # more elaborate lookup via scoring mechanism
+    # or last tail as fallback, e.g. `python` from `source.python`
+    return get_syntax_by_score(view, syntax_map) or tail
```
FichteFoll: `tail` will be `'source'` or `'text'` here and not what we want.

lint/util.py
```diff
@@ -610,3 +586,45 @@ def load_json(*segments, from_sl_dir=False):
...
+        }
+    except ValueError:
+        return
+    else:
+        max_score = max(scoring)
+        if max_score > 0:
+            return scoring[max_score]
+
+
+def get_syntax(view):
```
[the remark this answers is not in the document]
pykong: `get_syntax` is used across three different files. It makes sense to keep them in `util.py`.

Your question points towards the question of whether we should refactor `util.py` altogether. Seeing what code could be better moved elsewhere. And maybe even make a module out of `util.py` and allocate current functions to single files. But this is an issue of its own.

pykong:
@FichteFoll It does not make any sense to go into any details of this PR at this point, when there is a obviously fundamental misunderstanding at hand. 

You pointed me several times to the "full reference implementation". You also asked me in our previous exchange to implement a score selection based upon the API method `score_selector`.

Yet, I can not even find any match for `score_selector` in the entirety of the referenced repo:
[See here](https://github.com/Snazzah/SublimeDiscordRP/search?utf8=%E2%9C%93&q=score_selector&type=)

Are we even talking about the same thing here?

lint/util.py
pykong: No. I will not. ;-)

pykong:
We decided to not include this PR in the upcoming SL4.
Overhauling the syntax detection is too big of a change, with the potential ramification of rendering many linters unusable. It also requires to create and also to maintain a large syntax map.
We are also not aware of any other plugins employing scope based syntax detection across essentially all available syntaxes of ST, as required for SL.

We however do not exclude a renewal of the syntax detection in a future release of SL.
In this regard the PR at hand is up for grabs and can serve as a starting point for the implementation.

FichteFoll:
As explained in #691, I guess this implementation can be thought of more as a middle step or compatibility layer that is backwards compatible. I am okay with skipping this if something scope selector-based was implemented instead.

---

>It does not make any sense to go into any details of this PR at this point, when there is a obviously fundamental misunderstanding at hand.

I think I'd be able to wind up a PR for this on Wednesday or Thursday and base it on this PR as I believe that most parts are right. Assuming this is still a wanted feature. (Please don't delete the branch.)

>You also asked me in our previous exchange to implement a score selection based upon the API method score_selector.

The author of the reference implementation decided to optimize the `score_selector` part away by iterating over the base scope's segments manually. I considered the `score_selector` implementation to be easier to understand, but didn't consider that the reference didn't make use of it. Sorry for the confusion this may have caused.

>Overhauling the syntax detection is too big of a change, with the potential ramification of rendering many linters unusable.

I don't follow this. The change should be mostly transparent, except for exotic linters that target a very specific syntax name (like `JavaScript (Babel)`-specific as opposed to `javascript`). I'd have to go through all linters and check their `syntax` attribute for this.

braver:
> Assuming this is still a wanted feature.

It is. But it definitely needs at least one person to fully understand it and its impact across the plugins. I will try my best to follow along and help where I can, but I don't have a clear picture of it right now. We'll definitely keep the branch as a starting point if someone, e.g. you, wants to have a go at it.

> I'd have to go through all linters and check their syntax attribute for this.

Here you go:
[syntaxes.txt](https://github.com/SublimeLinter/SublimeLinter/files/1634155/syntaxes.txt)

pykong:
@FichteFoll any update on this?

FichteFoll:
Going through @braver's search results just now, this is probably going to be a huge pain. I didn't consider linter plugins to provide `syntax` attributes like:

```
SublimeLinter-contrib-bashate/linter.py:
   20:     syntax = 'shell-unix-generic'

SublimeLinter-contrib-cookstyle/linter.py:
   24:     syntax = ( 'better rspec', 'betterruby', 'cucumber steps', 'rspec', 'ruby experimental', 'ruby on rails', 'ruby' )

SublimeLinter-jsxhint/linter.py:
   25:     syntax = ('jsx', 'javascript_jsx', 'javascript (jsx)', 'javascript (babel)', 'javascript 6to5')
```

Or well, I did expect it somewhat but not to this extend. I believe we can catch all of the special cases through some custom mapping, but it's going to be quite some work to find whether all these syntaxes provide a proper base scope.
Maybe it's not worth spending time on this and instead work on a solely selector-based approach directly. If the selector was specified in and read from a linter's settings, users could then override that again, but I'd need to think about this a bit more.

Issue #800, opened 2018-01-14
[beta] refactor: statusbar shouldn't update panel


braver:
The statusbar and the panel are updated in the same event_listener. That's a good idea from a performance standpoint, but it's not very transparent that that code lives in statusbar. 

See also: https://github.com/SublimeLinter/SublimeLinter/pull/797#pullrequestreview-88666942

kaste:
Fixed in #822

pykong:
Fixed by: #822 
@kaste Just in case you did not know it.

You can reference issues in a PR, like this:
Fixes: #822 

Once the PR gets merged the issue gets closed automatically.

Issue #734, opened 2017-12-23
Consider new data structure for mapping view ids to objects.


pykong:
Right now multiple dictionaries exist mapping `view ids` to other objects.
A comprehensive list of those can be found in `sublime_linter.py`.
```
            self.loaded_views,
            self.linted_views,
            self.view_syntax,
            persist.errors,
            persist.highlights,
            persist.view_linters,
            persist.views,
            persist.last_hit_times
```
Regarding code quality we may consider that data structure bad.
For example, when deleting a view, SL needs to iterate over a list of those dictionaries to individually remove any reference to that deleted view.

Potential fixes are:
1. Merge the current dicts into a single one. Organized by view ids.
2. More preferably extend the view settings, to include the data currently held in individual dict.

FichteFoll:
Note that it may or may not be an advantage that view-specific settings are persisted across sessions.

kaste:
Wait, wait. What are *view specific settings*? 

Does that mean, if we store the last lint result in a special way, we can reuse it (say file mtime did not change or whatever) and draw the regions etc immediately?

FichteFoll:
`view.settings()` is a settings object specific to each view with a fallback to syntax-specific, project-specific user and default preferences. Any changes made to it are persisted across sessions.

kaste:
That sounds really interesting.

braver:
Fixed in #898

Issue #650, opened 2017-10-28
Refactor Module Structure - Remove Mutual Imports


pykong:
In the current implementation **SL** relies on many mutual imports between individual files in the `lint` folder.

This may represent not only a dangerous anti-pattern but makes extending and modifying the code difficult.
As any software need to change over time a refactoring towards a cleaner module structure will help ensure **SL**'s maintainability. 

Therefore such a cleaned up module structure should be investigated.

braver:
@pykong Think we covered this well enough for now?

braver:
I don't see many problems here anymore. Please re-open if you don't agree.

PR #817, opened 2018-01-15
Own controller for updating panel

pykong:
Addresses: #800 (see also: #797)
-  implemented own controller for panel updated based on`status_bar_view`

Also minor refactorings to prevent DRY:
- `get_current_position` is moved to `lint.util`
- `get_current_position` replaces almost identical method `get_line_and_col`
- `get_view_we_count` is now a method of class `Error`
-  renamed `get_we_count_line` to `get_line_we_count` to fit into naming scheme of other methods
- minor refactorings ...


pykong:
@kaste Could you review the code and tell whether you approve of the changes I have made?
(I can not add you via the side bar, somehow)

FichteFoll:
>I can not add you via the side bar, somehow.

I think you can only request reviews from collaborators.

lint/util.py
```diff
@@ -610,3 +610,10 @@ def load_json(*segments, from_sl_dir=False):
     base_path = "Packages/SublimeLinter/" if from_sl_dir else ""
     full_path = base_path + "/".join(segments)
     return sublime.decode_value(sublime.load_resource(full_path))
+
+
+def get_current_pos(view):
+    try:
+        return view.rowcol(view.sel()[0].begin())
+    except (AttributeError, IndexError):
+        return -1, -1
```
FichteFoll: I believe this should return a `sublime.Region` - or the above should return a tuple.

Considering that the default for `current_pos` in the `State` dicts is a tuple, probably the latter.

panel_view.py
```diff
@@ -0,0 +1,60 @@
+import sublime
+import sublime_plugin
+
+from .lint import persist
+from .lint import util
+from .lint import events
+from .panel import panel
+
+
+State = {
```
FichteFoll: I wonder if the state could be shared among the multiple implementations. Maintaining the same state twice does seem redundant and where most of the work ends up in. It would require a dedicated set of event hooks that update the state as well as an event mechanic notifying potential listeners of a state update. I'm not up to date on whether the internal event mechanic still exists.

I think traditionally the persist module was used for storing things like this a lot.

PS: also, can we please not use TitleCase for non-classes?

pykong:
I thought I could anyone belonging to the same org as the repo. Seems not true. 

**Addendum:** I really like to us to get beyond the code repetition introduced with that PR.
I tried to reduce the number of reused lines as much as I could without exact reproduction of the situation that led to this PR in the first place.

kaste:
I love how you really want to be DRY or get something DRY. But you're doing too much work here, too early. We literally want duplication here to occur, and later decide what is the best refactoring we can do. 

Why is that? E.g. since the major refactoring of `lint_view` which I had to do to get async stuff working, `persist.errors` as it stands is deprecated. Why is that? To get it working I actually had to change the data structure. It must be immutable and flat. (So btw the new data is stored in `persist.raw_errors` and only the goto command currently use it.)

It might also be already be deprecated that we have methods on `persist.errors`. Why is that? Data is data, and we have functions to transform it. It is probably that we can persist the data across sessions. So we need the data as a simple data structure, we cannot persist an object. (Side note: That `errors` is actually an instance or class is really confusing, it looks like it was intended as a compatibility layer.)

I asked you to be KISS here *instead* of DRY because of the moving parts. After that, I would have asked you to use your new State object as much as possible, so that we can see if the status bar and the panel really hold the same State. Bc I think as features requests come in, the panel will probably hold way more state. You can see it also from another point: Although *now* both views, staus bar and panel, have literally duplicate code, they don't share that much, if we look at the functionality, the goto and the panel have much more in common. So the principle of coupling or putting things together here should not be, if we literally see a code line twice or thrice, or like functions go in utils, classes each have their own file, html here. We should group them if they work somehow hand-in-hand. 

And then, I would have asked you to try to use the new `raw_errors`. Bc with that you start with a basic data structure, which you can transform as you wish. 

Related issues #650 #734

pykong:
@kaste OK. Makes a lot of sense if you explain it the way you did. :+1: 

You reassured me you have a plan where to move that whole code. We could go there - but step by step. (Really like to avoid those massive changesets of my previous contributions.)

kaste:
I would actually rename `panel/panel.py` to `panel_view.py` and put the State stuff inline, on the top. Then we actually don't fight against sublime's auto-reload feature which makes development much easier anyway.

pykong:
> Then we actually don't fight against sublime's auto-reload feature

@kaste Wait a minute. Did you say, in other words, we can get the SublimeLinter to autoreload on changes to its code? (Like any other plugin?). I killed so much productivity for me having to restart ST to update SL every time and even the reloader plugin wouldn't help.

kaste:
That's why I put 'my' view plugins or the goto file in the root. Sublime wants me to do that. I'm not totally masochistic here. 

The reloader basically doesn't work bc *we* fight against it with our `shared_plugin()` singleton in sublime_linter.py which is totally useless overhead anyway.

pykong:
> I would actually rename panel/panel.py to /panel_view.py and put the State stuff inline, on the top. 

Did that, now the panel seems no longer properly update when a new file finished linting.
Console is clear, no exceptions there.

Also the color scheme apparently changed.

I have pushed a7baaf7  for purposes of diagnosis, as I am not able to work out the underlying issue on my own,

lint/util.py
pykong: Good spot. It is a potential issue.
This pattern has not been introduced by this PR and it works nonetheless.
So, let's just fix this in a consecutive PR.

kaste:
Follow up is probably #822

pykong:
Replaced by: #822

PR #819, opened 2018-01-15
Reimplement RubyLinter conform to our API

### LINKED ISSUE #787: [beta] RubyLinter is not lazy
As mentioned in #782 `RubyLinter` loads eager. It is thus prone to err on fast start-ups on OSX bc if the `env[PATH]` is not immediately ready and good to use, it will disable itself bc it cannot find any reasonable executable.

Note: `RubyLinter` has its own interpretation of how and what extension points `Linter` provides. It should really implement `can_lint` and `context_sensitive_executable_path` to be lazy.

kaste:
Fixes #787

### DIFF
--- a/lint/base_linter/ruby_linter.py
+++ b/lint/base_linter/ruby_linter.py
@@ -4,11 +4,10 @@ import os
 import re
 import shlex
 import sublime
-import subprocess
 
 from functools import lru_cache
 
-from .. import linter, util, persist
+from .. import linter, util
 
 CMD_RE = re.compile(r'(?P<gem>.+?)@ruby')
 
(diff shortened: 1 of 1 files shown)


kaste:
7 minutes to review and test. Kudos.

braver:
Yes, I'm like a review ninja sometimes. The code didn't look wrong and it didn't seem to break anything I care about so I'm like 🎉 

tumblr_inline_nw3mul1x291qafrh6_400

Sorry.


Seriously though, if any future Ruby user has a problem I'll make some extra time to help em out and make up for this one.

FichteFoll:
Testing in production will inevitably yield the most promising results, albeit with a higher noise and annoyance ratio. 😉

lint/base_linter/ruby_linter.py
```diff
@@ -25,30 +24,25 @@ class RubyLinter(linter.Linter):
...
-
-        if not callable(cls.cmd) and cls.cmd:
-            cls.executable_path = cls.lookup_executables(cls.cmd)
-        elif cls.executable:
-            cls.executable_path = cls.lookup_executables(cls.executable)
-
-        if not cls.executable_path:
-            cls.disabled = True
+    @lru_cache(maxsize=None)
+    def can_lint(cls, syntax):
```
FichteFoll: Why do you override this? I'm not seeing it right now.

kaste: They implemented `initialize` which is called eager at Sublime start up. 

`can_lint` is generally lazy, called *once* on `cls` before first use. Bc of that it is still called before we have an instance, and a view. Usually, the default implementation works only with global available executables. 

So, if `can_lint` is relatively lax here (optimistic), you can implement `context_sensitive_executable_path`, which is called before every `lint`. (I just thought `ruby` has probably the same virtual env stuff like all the other languages.) Of course, if you do something before every lint, it should be fast.

Issue #813, opened 2018-01-14
[beta] panel scrolls up if you're below the last error of the current view


kaste:
![image](https://user-images.githubusercontent.com/8558/34921948-5ada45f0-f989-11e7-8712-7c2dbf61f180.png)

pykong:
To ensure I understood the issue correctly:
The panel scrolls to the **first** error of of the respective buffer, while the caret is placed below the last error in the view. *Hence the problem is not the up scroll per se, but that it scrolls to the wrong position in the panel.*

Issue #812, opened 2018-01-14
[beta] RenderLines abstraction is unnecessary


kaste:
Generally, `RenderLines` is only used here:

https://github.com/SublimeLinter/SublimeLinter/blob/5ee12b9256605ff57115686b3c61876645de021b/panel/panel.py#L209-L227

Issues: 

Since `to_render` is a simple object without any `__magic__` overrides, it can never be None: 

https://github.com/SublimeLinter/SublimeLinter/blob/5ee12b9256605ff57115686b3c61876645de021b/panel/panel.py#L227

The only useful method, `current_lineno` returns always `len(to_render)` if `to_render` were a simple list.

https://github.com/SublimeLinter/SublimeLinter/blob/5ee12b9256605ff57115686b3c61876645de021b/panel/panel.py#L222

pykong:
That class is indeed a relic and has lived past its time. I'll remove it.

Issue #804, opened 2018-01-14
[beta] For multiple errors on the same position, jump to the next position, not the next error


kaste:
Here, again my cursor is at 1:1. If I hit F4, I currently 'jump' to the next error in the list, which is still on 1:1, but I expect to go to 1:13.

![image](https://user-images.githubusercontent.com/8558/34918835-060ce610-f95a-11e7-93fe-d2be9d8f9a69.png)

(If I'm at pos 1:13, pressing shift-F4 actually jumps to the first 1:1 in the panel list. So it jumps correctly *over* the D101 error.)

pykong:
@kaste can you tell whether this behavior has been introduced by #797?

kaste:
Sorry, I can't, before your PR I actually never used F4 movement, bc it was just too confusing for me. It always jumped all over the places to whatever file. So I basically used the panel only with the mouse double click.

braver:
We do not control what F4 does in any way and I don't think we can (or should even try) because it would break the pattern. This behaviour therefore hasn't been introduced in any way by us.

pykong:
#803 implies a possible easy fix. If we extend the panel selection to encompass multiple lines this may well do the trick.

kaste:
I hoped for that as well 🐈

Issue #803, opened 2018-01-14
[beta] Highlight all errors under cursor in panel


kaste:
Here I'm on pos 1:1. It would be cool if both lines were highlighted

![image](https://user-images.githubusercontent.com/8558/34918809-966f25d4-f959-11e7-897e-b230a1d67841.png)

pykong:
This suggestion has been raised in the past.

Indeed I think it is a good suggestion.

Maybe related to #804

PR #823, opened 2018-01-16
Basing panel logic on raw errors

pykong:
This PR address a variety of issues. By basing panel logic of our new the flattened data structure `raw_errors` several issues get fixed. In the following they are listed. Unfortunately such change requires a bigger PR, than I initially intended, this inevitable to keep all panel functions alive.

## Region Sorting
Fixes: #802

After degression caused in #798, lines of the panel are now again sorted by starting column then by end column.

~~I see potential merge conflicts with the other panel related PR current open: #817 #822 
As the code living in `panel/panel.py` got moved. I decided to branch from master, as those PRs still got their problems. Hence, I don't know how stable of base any of those PRs really are.~~

## Panel Multiselect

If multiple error regions belong to the current position they will all get highlighted within the panel.
Fixes: #803 
Fixes: #804

### Preventing scroll when current position below last error
Fixes: #813

## Miscellaneous Cleanup

Removing relic class `RenderLines` Fixes: #812

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1092,7 +1092,7 @@ class Linter(metaclass=LinterMeta):
             "linter": self.name,
             "error_type": error_type,
             "code": m.error or m.warning or '',
-            "msg": m.message,
+            "msg": m.message.strip(),
         }
 
     def maybe_fix_tab_width(self, line, col, vv):
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -275,14 +275,15 @@ def update_panel_selection(active_view, we_count, current_pos, **kwargs):
     if panel_lines[0] == panel_lines[1]:
         region = get_panel_region(panel_lines[0], panel, is_full_line)
     else:  # multiple panel lines
+        is_full_line = True
         region_a = get_panel_region(panel_lines[0], panel)
-        region_b = get_panel_region(panel_lines[1], panel, is_full_line=True)
+        region_b = get_panel_region(panel_lines[1], panel, is_full_line)
         region = sublime.Region(region_a.begin(), region_b.end())
 
     update_selection(panel, region)
 
-    # scroll selection into panel
-    if not panel.visible_region().contains(region):
+    # scroll non-empty selection into panel if ot visible
+    if not panel.visible_region().contains(region) and is_full_line:
         panel.show(region)
 
     # simulate scrolling to enforce rerendering of panel,
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -188,8 +188,6 @@ def fill_panel(window, update=False, **panel_filter):
             # insert empty line between views sections
             to_render.append("")
 
-        import json
-        print(json.dumps(buf_lines, indent=2, sort_keys=True))
     run_update_panel_cmd(panel, text="\n".join(to_render))
 
 
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -2,7 +2,6 @@ import os
 import sublime
 import bisect
 
-from ..lint.const import WARN_ERR
 from ..lint import util, persist
 
 PANEL_NAME = "SublimeLinter"
@@ -23,33 +22,6 @@ OUTPUT_PANEL_SETTINGS = {
 }
 
 
-class RenderLines:
-    def __init__(self):
-        self.lines = []
-        self._line_counter = 0
-
-    def append(self, str):
-        self.lines.append(str)
-        self._line_counter += 1
-
-    def current_lineno(self):
-        return self._line_counter
-
-    def render(self):
-        return "\n".join(self.lines)
-
-
-def dedupe_views(errors):
-    if len(errors) == 1:
-        return errors
-    else:
-        return {
-            vid: dic
-            for vid, dic in errors.items()
-            if sublime.View(vid).is_primary()
-        }
-
-
 def get_common_parent(paths):
     """
     Get the common parent directory of multiple paths.
@@ -63,15 +35,23 @@ def get_common_parent(paths):
     return common_path.rstrip('/\\')
 
 
-def create_path_dict(x):
-    abs_dict = {}
+def get_filenames(window, bids):
+    """Return dict of buffer_id: file_name for all views in window.
+
+       Untitled buffers are file names are substituded by:
+       <untitled buffer_id>
+    """
+
+    return {
+        v.buffer_id(): v.file_name() or "<untitled {}>".format(v.buffer_id())
+        for v in window.views()
+        if v.buffer_id() in bids
+    }
+
+
+def create_path_dict(window, bids):
     base_dir = ""
-    for vid in x:
-        view = sublime.View(vid)
-        if view.file_name():
-            abs_dict[vid] = view.file_name()
-        else:
-            abs_dict[vid] = "<untitled " + str(view.buffer_id()) + ">"
+    abs_dict = get_filenames(window, bids)
 
     if len(abs_dict) == 1:
         for vid, path in abs_dict.items():
@@ -118,9 +98,9 @@ def ensure_panel(window: sublime.Window):
     return get_panel(window) or create_panel(window)
 
 
-def filter_errors(window, errors):
-    vids = [v.id() for v in window.views()]
-    return {vid: d for vid, d in errors.items() if vid in vids}
+def get_window_raw_errors(window, errors):
+    bids = {v.buffer_id() for v in window.views()}
+    return {bid: d for bid, d in errors.items() if bid in bids}
 
 
 def format_header(f_path):
@@ -136,68 +116,49 @@ def run_update_panel_cmd(panel, text=None):
     panel.run_command(cmd, {'text': text, 'clear_sel': clear_sel})
 
 
-def format_row(lineno, error_type, dic):
-    lineno = int(lineno) + 1
-    start = dic['start'] + 1
-    msg = dic['msg'].rstrip()
-    tmpl = " {LINENO:>5}:{START:<4} {ERR_TYPE:7} {linter:>12}: {code:12} {MSG}"
-    return tmpl.format(
-        LINENO=lineno, START=start, ERR_TYPE=error_type, MSG=msg, **dic)
+def format_row(item):
+    line = int(item["line"]) + 1
+    start = item['start'] + 1
+    tmpl = " {LINE:>5}:{START:<4} {error_type:7} {linter:>12}: {code:12} {msg}"
+    return tmpl.format(LINE=line, START=start, **item)
 
 
-def filter_ok(check_val, key, panel_filter):
+def passes_listing(buf_d, panel_filter):
     """Check value against white- and blacklisting and return bool."""
     # check whitelisting
     if not panel_filter:
         return True
 
-    if key in panel_filter and check_val not in panel_filter[key]:
-        return False
+    x = {"types": "error_type", "codes": "code", "linter": "linter"}
+    for key, check_val in x.items():
+        # check whitelisting
+        if key in panel_filter:
+            if buf_d[check_val] not in panel_filter[key]:
+                return False
 
-    # check blacklisting
-    exclude_key = "exclude_" + key
-    if exclude_key in panel_filter and check_val in panel_filter[exclude_key]:
-        return False
+        # check blacklisting
+        exclude_key = "exclude_" + key
+        if exclude_key in panel_filter:
+            if buf_d[check_val] in panel_filter[exclude_key]:
+                return False
 
+    # if all checks passed return True
     return True
 
 
-def get_view_lines(view_dict, panel_filter):
-    view_lines = []
-    for lineno, line_dict in sorted(view_dict["line_dicts"].items()):
-        items = []
-        for error_type in WARN_ERR:
-            if not filter_ok(error_type, "types", panel_filter):
-                continue
-
-            err_dict = line_dict.get(error_type)
-            if not err_dict:
-                continue
-
-            items += [
-                (lineno, error_type, item)
-                for item in err_dict
-                if filter_ok(item['linter'], "linter", panel_filter) and
-                filter_ok(item['code'], "codes", panel_filter)
-            ]
-
-        # sort items by start col, then by end col
-        items.sort(key=lambda x: (x[2]['start'], x[2]['end']))
-
-        view_lines.extend(items)
-
-    return view_lines
+def get_buf_lines(buf_dicts, panel_filter):
+    return [d for d in buf_dicts if passes_listing(d, panel_filter)]
 
 
 def fill_panel(window, update=False, **panel_filter):
-    errors = persist.errors.data
+
+    errors = persist.raw_errors
     if not errors:
         return
 
-    errors = filter_errors(window, errors)
-    errors = dedupe_views(errors)
+    errors = get_window_raw_errors(window, errors)
 
-    path_dict, base_dir = create_path_dict(errors)
+    path_dict, base_dir = create_path_dict(window, errors)
     assert window, "missing window!"
 
     panel = ensure_panel(window)
@@ -211,26 +172,25 @@ def fill_panel(window, update=False, **panel_filter):
     else:
         settings.set("sublime_linter_panel_filter", panel_filter)
 
-    to_render = RenderLines()
-    for vid, view_dict in errors.items():
-        if util.is_none_or_zero(view_dict["we_count_view"]):
-            continue
-
-        view_lines = get_view_lines(view_dict, panel_filter)
-        if view_lines:
+    to_render = []
+    for bid, buf_dict in errors.items():
+        buf_lines = get_buf_lines(buf_dict, panel_filter)
+        if buf_lines:
             # append header
-            to_render.append(format_header(path_dict[vid]))
+            to_render.append(format_header(path_dict[bid]))
 
             # append lines
-            for lineno, error_type, item in view_lines:
-                to_render.append(format_row(lineno, error_type, item))
-                item["panel_lineno"] = to_render.current_lineno()
+            base_lineno = len(to_render)
+            for i, item in enumerate(buf_lines):
+                to_render.append(format_row(item))
+                item["panel_lineno"] = base_lineno + i
 
             # insert empty line between views sections
             to_render.append("")
 
-    rendered_text = to_render.render() if to_render else None
-    run_update_panel_cmd(panel, text=rendered_text)
+        import json
+        print(json.dumps(buf_lines, indent=2, sort_keys=True))
+    run_update_panel_cmd(panel, text="\n".join(to_render))
 
 
 # logic for updating panel selection
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -256,8 +256,8 @@ def get_next_lineno(num, interval):
         return neighbours[1]
 
 
-def change_selection(panel_lineno, full_line=False, window=None):
-    panel = get_panel(window or sublime.active_window())
+def change_selection(panel_lineno, full_line=False):
+    panel = get_panel(sublime.active_window())
     if not panel:
         return
 
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -102,8 +102,11 @@ def sort_errors(errors):
 
 
 def get_window_raw_errors(window, errors):
-    bids = {v.buffer_id() for v in window.views()}
-    return {bid: sort_errors(d) for bid, d in errors.items() if bid in bids}
+    return {
+        bid: sort_errors(errors[bid])
+        for bid in {v.buffer_id()
+                    for v in window.views()}
+    }
 
 
 def format_header(f_path):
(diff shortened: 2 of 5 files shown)


braver:
We can probably merge this and then rebase the other branches.

panel/panel.py
```diff
@@ -165,22 +165,27 @@ def filter_ok(check_val, key, panel_filter):
...
             ]
-            for item in items:
-                view_lines.append((lineno, error_type, item))
+
+        # sort items by start col, then by end col
+        items.sort(key=lambda item: (item['start'], item['end']))
+
+        # return a flat list here for easier processing
+        for item in items:
+            view_lines.append((lineno, error_type, item))
```
kaste: Nope. You lose the correct error_type here.

pykong: Indeed! Flat is better than nested. Thanks for making me aware of the `raw_errors`.

kaste:
Ugly, but otherwise does the job 😄

kaste:
Since I'm into code ⛳️ today, I'll send a refactor of your code.

pykong:
@kaste Thanks for #829. Never had a commit dedicated to me.
Your `raw_errors` implementation is simple so great.
I was too unthinking an timid to touch that base data structure of SL.

Incredible how much effort we spent just to pack and unpack that.

`On your requirement to base the panel logic on the new `raw_errors` instead I did exactly that.`
Certainly this will make this PR much bigger again.

We need to check how this fits together with #829 
And also #826 needs to be merged as changes invariably affect the panel selection as well.

panel/panel.py
```diff
@@ -136,64 +118,51 @@ def run_update_panel_cmd(panel, text=None):
...
-                if filter_ok(item['linter'], "linter", panel_filter) and
-                filter_ok(item['code'], "codes", panel_filter)
-            ]
-            for item in items:
-                view_lines.append((lineno, error_type, item))
-
-    return view_lines
+def get_buf_lines(buf_dicts, panel_filter):
+    buf_dicts = [d for d in buf_dicts if passes_listing(d, panel_filter)]
+    return sorted(buf_dicts, key=lambda x: (x["line"], x["start"], x["end"]))
```
pykong: Line redundant? Are `raw_errors` already sorted that way?

FichteFoll:
Resolves #812 btw.

panel/panel.py
kaste: No. Unsorted, arbitrary order.

sublime_linter.py
```diff
@@ -243,6 +243,7 @@ def highlight(self, view, errors, hit_time):
             return
 
         bid = view.buffer_id()
+        errors.sort(key=lambda x: (x["line"], x["start"], x["end"]))
         persist.raw_errors[bid] = errors
```
FichteFoll: With this, you should mention somewhere (in the docstring/comment of `persist.raw_errors` most likely) that its values are supposed to be sorted at all times (an invariant).

kaste: I already mentioned that we don't sort here. The order of the errors is arbitrary as results fly in. Also this particular implementation mutates errors.

pykong: OK. This is why I did not write a docstring. Moving it back and using `sorted( `.

FichteFoll: But sorting was added in https://github.com/SublimeLinter/SublimeLinter/pull/823/commits/c8f5472c233e286862b597b5697bf95bb06bd0a1, unless I'm mistaken?

panel/panel.py
```diff
@@ -118,9 +98,10 @@ def ensure_panel(window: sublime.Window):
     return get_panel(window) or create_panel(window)
 
 
-def filter_errors(window, errors):
-    vids = [v.id() for v in window.views()]
-    return {vid: d for vid, d in errors.items() if vid in vids}
+def get_window_raw_errors(window, errors):
+    bids = {v.buffer_id() for v in window.views()}
+    return {bid: d for bid, d in errors.items() if bid in bids}
+    errors.sort(key=lambda x: (x["line"], x["start"], x["end"]))
```
kaste: flip the lines

kaste: No inline sorting!

panel/panel.py
```diff
@@ -207,107 +175,103 @@ def fill_panel(window, update=False, **panel_filter):
...
+            to_render.append(format_header(path_dict[bid]))
 
             # append lines
-            for lineno, error_type, item in view_lines:
-                to_render.append(format_row(lineno, error_type, item))
-                item["panel_lineno"] = to_render.current_lineno()
+            base_lineno = len(to_render)
+            for i, item in enumerate(buf_lines):
+                to_render.append(format_row(item))
+                item["panel_line"] = base_lineno + i
```
kaste: mutate alert!

kaste: Okay if it's too hard, skip this comment

pykong: We need that for updating the panel selection, that you all adore so much.
It might have escaped your notice that the panel line was written to `persist.errors` also.
If you got a different suggestion, just tell.

pykong:
Check this out! Multi line selection works again! Now fully build on kaste's ingenious `raw_errors`
multi_line_select

panel/panel.py
```diff
@@ -207,107 +175,103 @@ def fill_panel(window, update=False, **panel_filter):
...
 
 
 def update_panel_selection(active_view, we_count, current_pos, **kwargs):
     if current_pos == (-1, -1):
         return
 
-    full_line = False
-    view_dict = persist.errors.get_view_dict(active_view.id())
-    if not view_dict or util.is_none_or_zero(we_count):
+    buf_dicts = persist.raw_errors.get(active_view.buffer_id())
```
pykong: I have noticed a `NoneType exception` in the console once. How could this be with a defaultdict (`{}`).
Shouldn't the default value be an empty dict instead?
Wished that guard immediately below could go.

pykong:
Please test and review. From my side - it is merge ready.

panel/panel.py
kaste: Using `get` you get None's bc otherwise you would add keys to the dict when you would like to check if a key is in the dict. defaultdict replaces exactly what we do with `setdefault(key, {})`

`bar.setdefault('foo', []).append('fox') -- becomes --> bar['foo'].append('fox')`

It is more readable and faster.

panel/panel.py
kaste: I don't tell you yet bc than you freak out and change everything again.

panel/panel.py
```diff
@@ -136,64 +119,51 @@ def run_update_panel_cmd(panel, text=None):
...
 
-def format_row(lineno, error_type, dic):
-    lineno = int(lineno) + 1
-    start = dic['start'] + 1
-    msg = dic['msg'].rstrip()
-    tmpl = " {LINENO:>5}:{START:<4} {ERR_TYPE:7} {linter:>12}: {code:12} {MSG}"
-    return tmpl.format(
-        LINENO=lineno, START=start, ERR_TYPE=error_type, MSG=msg, **dic)
+def format_row(item):
+    line = int(item["line"]) + 1
```
kaste: I actually cast to int in Linter. Did you ever had an error here?

panel/panel.py
```diff
@@ -136,64 +119,51 @@ def run_update_panel_cmd(panel, text=None):
     panel.run_command(cmd, {'text': text, 'clear_sel': clear_sel})
 
 
-def format_row(lineno, error_type, dic):
-    lineno = int(lineno) + 1
-    start = dic['start'] + 1
-    msg = dic['msg'].rstrip()
```
kaste: Where is the `rstrip`?

panel/panel.py
```diff
@@ -118,9 +97,13 @@ def ensure_panel(window: sublime.Window):
...
-def filter_errors(window, errors):
-    vids = [v.id() for v in window.views()]
-    return {vid: d for vid, d in errors.items() if vid in vids}
+def sort_errors(errors):
+    return sorted(errors, key=lambda x: (x["line"], x["start"], x["end"]))
+
+
+def get_window_raw_errors(window, errors):
+    bids = {v.buffer_id() for v in window.views()}
+    return {bid: sort_errors(d) for bid, d in errors.items() if bid in bids}
```
kaste: I always thought about this as 

```
{bid: sorted(errors[bid]) for bid in {v.buffer_id() for v in window.views()}}

```

T.i. you don't iterate over `items()` and filter this via if bid in bids

panel/panel.py
```diff
@@ -207,107 +177,115 @@ def fill_panel(window, update=False, **panel_filter):
...
 
-    to_render = RenderLines()
-    for vid, view_dict in errors.items():
-        if util.is_none_or_zero(view_dict["we_count_view"]):
-            continue
-
-        view_lines = get_view_lines(view_dict, panel_filter)
-        if view_lines:
+    to_render = []
+    for bid, buf_dict in errors.items():
```
kaste: `buf_dict` is actually a list these are the `errors`, no?

what you call `errors`above is probably `all_errors` or `errors_by_bid`

panel/panel.py
```diff
@@ -136,64 +119,51 @@ def run_update_panel_cmd(panel, text=None):
...
-                if filter_ok(item['linter'], "linter", panel_filter) and
-                filter_ok(item['code'], "codes", panel_filter)
-            ]
-            for item in items:
-                view_lines.append((lineno, error_type, item))
-
-    return view_lines
+def get_buf_lines(buf_dicts, panel_filter):
+    """Filter raw_errors through white- and blacklisting, then sort them."""
+    buf_dicts = [d for d in buf_dicts if passes_listing(d, panel_filter)]
```
kaste: This is confusing naming. `d` -> error / buf_dicts -> errors

panel/panel.py
```diff
@@ -136,64 +119,51 @@ def run_update_panel_cmd(panel, text=None):
...
-        LINENO=lineno, START=start, ERR_TYPE=error_type, MSG=msg, **dic)
+def format_row(item):
+    line = int(item["line"]) + 1
+    start = item['start'] + 1
+    tmpl = " {LINE:>5}:{START:<4} {error_type:7} {linter:>12}: {code:12} {msg}"
+    return tmpl.format(LINE=line, START=start, **item)
 
 
-def filter_ok(check_val, key, panel_filter):
+def passes_listing(buf_d, panel_filter):
```
kaste: buf_d -> error

panel/panel.py
```diff
@@ -136,64 +119,51 @@ def run_update_panel_cmd(panel, text=None):
...
     """Check value against white- and blacklisting and return bool."""
     # check whitelisting
     if not panel_filter:
         return True
 
-    if key in panel_filter and check_val not in panel_filter[key]:
-        return False
+    x = {"types": "error_type", "codes": "code", "linter": "linter"}
```
kaste: what is this a map from ? error keys to style setting keys ?

panel/panel.py
```diff
@@ -207,107 +177,115 @@ def fill_panel(window, update=False, **panel_filter):
...
+def get_next_panel_line(buf_line, buf_dict):
+    """Get panel line for next error in buffer.
 
-    if num < interval[0] or interval[-1] < num:
-        return interval[0]
+    If not found this means the current line is past last error's buffer line.
+    In that case return last error's panel line incremented by one, which will
+    place panel selection in empty space between buffer sections.
+    """
+    for d in buf_dict:
```
kaste: again just the name `for error in errors`

panel/panel.py
pykong: No, that casting was not initially introduced by me. I just stuck with it.

panel/panel.py
pykong: In Vegas? Let's place them upstream in the data flow like in `Linter.error` bc status bar messages would profit from it as well. And why not `strip` both sides altogether?

panel/panel.py
pykong: Yeah, makes sense to iterate over buffers in active window = shorter.

panel/panel.py
pykong: Just to ensure I understand you here: your comment is in regards of variable names not logic?

panel/panel.py
pykong: It is for command args you use for white-/ blacklisting to keys in `raw_errors`.

panel/panel.py
kaste: Yes

panel/panel.py
kaste: COMMAND_ARG_TO_ERROR_KEY

panel/panel.py
pykong: I am eager to find out what you got in your sleeves.

panel/panel.py
pykong: Upper case, bc it is a const? I have seldom seen const been declared in function. Usually they are in global scope at top at the file.

kaste: But it *is* a constant mapping regardless where you define it

panel/panel.py
```diff
@@ -136,64 +122,56 @@ def run_update_panel_cmd(panel, text=None):
...
-                filter_ok(item['code'], "codes", panel_filter)
-            ]
-            for item in items:
-                view_lines.append((lineno, error_type, item))
-
-    return view_lines
+def get_buf_lines(buf_errors, panel_filter):
+    """Filter raw_errors through white- and blacklisting, then sort them."""
+    buf_errors = [e for e in buf_errors if passes_listing(e, panel_filter)]
+    return sorted(buf_errors, key=lambda x: (x["line"], x["start"], x["end"]))
```
kaste: I still think you sort twice.

kaste: Otherwise, this is the inlined version of *sort_errors* above.

panel/panel.py
```diff
@@ -136,64 +122,56 @@ def run_update_panel_cmd(panel, text=None):
...
+    errors = persist.raw_errors
     if not errors:
         return
 
-    errors = filter_errors(window, errors)
-    errors = dedupe_views(errors)
+    errors = get_window_raw_errors(window, errors)
 
-    path_dict, base_dir = create_path_dict(errors)
+    path_dict, base_dir = create_path_dict(window, errors)
```
kaste: Here we have not errors but errors_by_bid. errors means a list, errors_by_bid relates naturally to a dict. In create_path_dict you call this arg bids, which is simply not what it is. It is enough though to pass down only the `.keys()`, bc the dict itself is never used.

panel/panel.py
```diff
@@ -207,107 +185,115 @@ def fill_panel(window, update=False, **panel_filter):
...
-    to_render = RenderLines()
-    for vid, view_dict in errors.items():
-        if util.is_none_or_zero(view_dict["we_count_view"]):
-            continue
-
-        view_lines = get_view_lines(view_dict, panel_filter)
-        if view_lines:
+    to_render = []
+    for bid, buf_errors in errors.items():
+        buf_lines = get_buf_lines(buf_errors, panel_filter)
```
kaste: I think you can also filter in get_window_raw_errors in one go, unless I'm mistaken. The point is to get the right data set as soon as possible.  Then the following `if` branch would also be always true, thus unnecessary.

panel/panel.py
```diff
@@ -207,107 +185,115 @@ def fill_panel(window, update=False, **panel_filter):
...
+    If current position is past last error, place empty selection on the panel line following that of last error.
+
+    """
     if current_pos == (-1, -1):
         return
 
-    full_line = False
-    view_dict = persist.errors.get_view_dict(active_view.id())
-    if not view_dict or util.is_none_or_zero(we_count):
+    buf_errorss = persist.raw_errors.get(active_view.buffer_id())
```
kaste: typo, we're not nazis

lint/linter.py
```diff
@@ -1092,7 +1092,7 @@ def process_match(self, m, vv):
             "linter": self.name,
             "error_type": error_type,
             "code": m.error or m.warning or '',
-            "msg": m.message,
+            "msg": m.message.strip(),
```
kaste: I thought about this today, and looked at different code spots where we use this. 

I think this is okay (as I said I looked today). But, I asked you to just file an issue about this. The change is simple, and you could have just let this rstrip here bc right now we don't have a need for this anywhere in the code. And with an issue open everyone gets involved and can raise a hand. 

As I said again, **I** think this is safe and reasonable, and I would have raised my hand in such a issue, but the other contributors have to have a look here as well.

panel/panel.py
pykong: The latter. DRY violation.

kaste:
All the team members @braver, @FichteFoll ! Hi 👋 ! 

@pykong has a wild PR. Join the reviewlution.

panel/panel.py
kaste: No, then the code is not clear. It is very unusal that you have to sort the same data-set on the exact same keys twice. 

`get_window_raw_errors -> {bid: sorted(errors)}`
`get_buf_lines(sorted(errors), filterfn) -> filtered_sorted_errors`

lint/linter.py
pykong: Sorry. Missed that.

pykong:
@braver This could be a super productive weekend. Could we please move this PR forward?

lint/linter.py
braver: Sure, I have nothing against strip here. 

We do a `line.rstrip()` elsewhere... perhaps we should sanitise our values all in one place so everything else can just rely on it being stripped?

braver:
Just the one comment is all I have, looks good. @kaste hit merge if you agree.

lint/linter.py
pykong: Absolutely! I have opened an issue for it #854.
My suggestion let's just leave the `strip` there, because it is already moved out of `panel.py`
and we later on centralize sanitation.

FichteFoll:
I think I'll skip on this one. Haven't done anything with the internal structs yet and am short on time for today.

kaste:
Does not update on activate event. T.i when you switch between views it does currently not scroll. 

Does not seem to update on selection modified event. Only when we sit on an error. See:

image

Here I am on line 25, but I actually have an error on line 23

panel/panel.py
```diff
@@ -63,15 +34,23 @@ def get_common_parent(paths):
...
+
+def create_path_dict(window, bids):
     base_dir = ""
-    for vid in x:
-        view = sublime.View(vid)
-        if view.file_name():
-            abs_dict[vid] = view.file_name()
-        else:
-            abs_dict[vid] = "<untitled " + str(view.buffer_id()) + ">"
+    abs_dict = get_filenames(window, bids)
```
kaste: What does `abs_dir` mean? This is a mapping from bid to file_id or normalized_filename.  Something like this.

panel/panel.py
```diff
@@ -63,15 +34,23 @@ def get_common_parent(paths):
...
 
     if len(abs_dict) == 1:
         for vid, path in abs_dict.items():
```
kaste: vid -> bid

kaste: here and below L66

panel/panel.py
```diff
@@ -136,64 +128,50 @@ def run_update_panel_cmd(panel, text=None):
...
+    errors = persist.raw_errors
     if not errors:
         return
 
-    errors = filter_errors(window, errors)
-    errors = dedupe_views(errors)
+    errors_by_bid = get_window_raw_errors(window, errors, panel_filter)
 
-    path_dict, base_dir = create_path_dict(errors)
+    path_dict, base_dir = create_path_dict(window, errors_by_bid)
```
kaste: create_path_dict expects `bids`as its second argument, so passing down a dict is confusing. Pass down `errors_by_bid.keys()`

panel/panel.py
```diff
@@ -207,107 +185,117 @@ def fill_panel(window, update=False, **panel_filter):
...
-    interval = list(interval)
-    interval.sort()
-
-    if num < interval[0] or interval[-1] < num:
-        return interval[0]
+    If not found this means the current line is past last error's buffer line.
+    In that case return last error's panel line incremented by one, which will
+    place panel selection in empty space between buffer sections.
+    """
+    for d in buf_errors:
```
kaste: d -> error

panel/panel.py
```diff
@@ -207,107 +185,117 @@ def fill_panel(window, update=False, **panel_filter):
...
     if current_pos == (-1, -1):
         return
 
-    full_line = False
-    view_dict = persist.errors.get_view_dict(active_view.id())
-    if not view_dict or util.is_none_or_zero(we_count):
+    buf_errors = persist.raw_errors.get(active_view.buffer_id())
+    if not buf_errors:  # None type error with default dict??
+        return
+    buf_errors = [d for d in buf_errors if "panel_line" in d]
```
kaste: d -> error

panel/panel.py
```diff
@@ -207,107 +185,117 @@ def fill_panel(window, update=False, **panel_filter):
...
+    if not buf_errors:
         return
 
-    (lineno, colno) = current_pos
-    line_dicts = view_dict["line_dicts"]
+    (buf_line, col) = current_pos
+    line_dicts = [d for d in buf_errors if d["line"] == buf_line]
```
kaste: d -> error

panel/panel.py
```diff
@@ -207,107 +185,117 @@ def fill_panel(window, update=False, **panel_filter):
...
-    else:
-        lineno = get_next_lineno(lineno, line_dicts)
+    panel_lines = None
+    is_full_line = False
 
-    lineno = 0 if lineno is None else lineno
+    if line_dicts:
+        # we got line dict, now check if current position has errors
+        region_panel_lines = [
+            d["panel_line"]
```
kaste: d->error

panel/panel.py
```diff
@@ -118,9 +97,22 @@ def ensure_panel(window: sublime.Window):
...
 
 
-def filter_errors(window, errors):
-    vids = [v.id() for v in window.views()]
-    return {vid: d for vid, d in errors.items() if vid in vids}
+def sort_errors(errors):
+    return sorted(errors, key=lambda x: (x["line"], x["start"], x["end"]))
+
+
+def get_filter_and_sort(buf_errors, panel_filter):
```
kaste: get_filter_and_sort -> filter_and_sort otherwise in indicates you return probably a function

panel/panel.py
braver: Historically it’s a dictionary of absolute paths IIRC

panel/panel.py
braver: I should’ve caught this too. Unless a var is short lived, eg. scopes inside a short loop or something, it shouldn’t have a single character for a name.

panel/panel.py
pykong: yes, historically `abs_` referring to absolute file paths. We should come up with more descriptive name though

lint/linter.py
braver: Works for me

panel/panel.py
braver: The docstring for get_filenames needs another look too: “buffers are file names are substituded“ has a typo and doesn’t make sense.

Anyway... How about bid_filenames?

I suck at this at least as much as everyone else. 😐

panel/panel.py
pykong: Absolutely. That function names is bad.

panel/panel.py
pykong: I prefer renaming `buf_errors` -> plain `errors` and `d` -> `e` standing for error. What do you think?

I feel comprehensions are one of few contexts in which single char variable names make sense.
Their shortness provide additional visual structure. Yet, they should be named according to the first letter of the iterable.

panel/panel.py
pykong: `filenames_by_bid`

braver: Go for it 👍🏻👍🏻

kaste:
Holy shit, it works! 😢

kaste:
Kudos, you fixed like 8 issues in one PR, and you adapted to the new error data structure which wasn't easy at all. You approached the problems in a totally chaotic way, but went through the storm. Now the 👑 is yours.

braver:
Awesome job all!

Issue #818, opened 2018-01-15
Naively: the default current working directory should be the project dir


kaste:
This is coming from #581 

Current code for the `cwd` is

https://github.com/SublimeLinter/SublimeLinter/blob/ce7045121f6d32b7189ad9340921e5b285633867/lint/linter.py#L951-L962

which means we use the `dirname`, t.i. the folder, of the current file if we have a saved file. For unsaved files, we 'expand' the dot, which can be anything, e.g. it can be the sublime folder. 

I really don't know, what the rationale here is. For me, it seems unpredictable for unsaved, new files. It is not what I usually do and expect if I'm working on the command line. Everyones, almost all the time works from the root of the current project, to lint, run test commands, `npm build` whatever.

So, please enlighten me. Why is that? Should we change it?

FichteFoll:
As an aside, I so something similar but more automatic for the mypy linter here:

https://github.com/fredcallaway/SublimeLinter-contrib-mypy/blob/cd6fbaf0cd9c8f52b23df621537fa9276a92bae1/linter.py#L134-L142

pykong:
This!

braver:
> So, please enlighten me. Why is that? Should we change it?

I tried a few things, and the result of `os.path.realpath('.')` never made any sense to me. I also don't see an API that gives us a more meaningful directory. So, I can totally get behind your proposal, at least that's predictable.

FichteFoll:
Suggestion: Allow a linter-specific override for this as `cwd` that defaults to `${folder}`, which is either the project folder containing the file or the file's containing folder.

Highly related to #825.

kaste:
`${folder:$file_path}` is exactly what I wanted to implement here. But this is blocked by #825 where I factored that `guess_project_path` fn. 

I don't think we need a fallback here, you proposed in #832, bc if we return `None` and pass it down to `popen`, it will not break anything, it will just not change the cwd, which is good.

We have the linter specific override, but the setting is called `chdir`. Plugin authors can provide a default
here, and end users can override this again in their settings. 

`chdir` would get the new var-expansions stuff from #825

FichteFoll:
I think the complaint in this issue was also about `.` being used as the working dir, which is kind of unpredictable. I thought using the tempdir for that would be deterministic and more desirable. The fallback would only get used for untitled views anyway. I'm pretty much indifferent on this though.

FichteFoll:
[...]

PR #825, opened 2018-01-16
Refactor view settings

### LINKED ISSUE #824: Actually where is the `${project}`
I wanted to refactor the settings-tokens-stuff after #581 esp. https://github.com/SublimeLinter/SublimeLinter/pull/581#issuecomment-357340886

Also related #750 and #818 

So, how the find the project home?

1. If you just open a file in a new window:

> >>> w = sublime.active_window()
>>> w.project_file_name()  # None
>>> w.project_data()  # None

`w.extract_variables` also doesn't contain anything useful. That is okay.

2. If you open or drag a folder into Sublime without a project file:

> >>> w = sublime.active_window()
>>> w.project_data()
{'folders': [{'path': 'C:\\Dev\\Sublime Text 3'}]}
>>> w.project_file_name()  # still None

`w.extract_variables()` will not contain that folder.

3. If you open a folder with a project file:

`w.project_data()` will have all the settings, the first folder in folders has whatever the user typed here, but most probably `'.'`, which we cannot resolve via `os.path.realpath`, but `extract_variables` will have expanded path at ~~`['project_path']`~~ EDIT: it's `['folder']`

Any additional folder is always a realpath.

Any comments here @FichteFoll @braver @pykong, otherwise I'll go from here

kaste:
Confirm to the standards.

**** BREAKING ****

We start with all Sublime variables, but enhance the ${folder} with a better
guess. We use standard python to expand '~'. We also expand all environment 
variables.

- ${folder} instead of ${project}
- ~ instead of ${home}
- ${GEM_HOME} instead of ${env:GEM_HOME}

We use `sublime.expand_variables` under the hood, so we get 'OR' functionality
for free. E.g. `${XDG_CONFIG_HOME:~.config}/blub` will just work. 

Note that our current wanted default working directory (specced in #818) can be expressed just 
like `${folder:$file_path}`. T.i. try the *folder*, fallback to `os.path.dirname(filename)`.

Fixes #824 

Notable: Removed `cls.lint_settings`

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -477,9 +477,8 @@ class Linter(metaclass=LinterMeta):
                 return value
 
         window = self.view.window()
-        variables = os.environ.copy()
-        if window:
-            variables.update(window.extract_variables())
+        variables = ChainMap(
+            {}, window.extract_variables() if window else {}, os.environ)
 
         filename = self.view.file_name()
         project_folder = self._guess_project_path(window, filename)
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -476,11 +476,13 @@ class Linter(metaclass=LinterMeta):
 
 
         """
-        def recursive_replace_value(expressions, value):
+        def recursive_replace(expressions, value):
             if isinstance(value, dict):
-                value = recursive_replace(expressions, value, nested=True)
+                value = {key: recursive_replace(expressions, val)
+                         for key, val in value.items()}
             elif isinstance(value, list):
-                value = [recursive_replace_value(expressions, item) for item in value]
+                value = [recursive_replace(expressions, item)
+                         for item in value]
             elif isinstance(value, str):
                 for exp in expressions:
                     if isinstance(exp['value'], str):
@@ -489,12 +491,6 @@ class Linter(metaclass=LinterMeta):
                         value = exp['token'].sub(exp['value'], value)
 
             return value
-
-        def recursive_replace(expressions, mutable_input, nested=False):
-            for key, value in mutable_input.items():
-                mutable_input[key] = recursive_replace_value(expressions, value)
-            if nested:
-                return mutable_input
 
         # Expressions are evaluated in list order.
         expressions = []
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -459,29 +459,29 @@ class Linter(metaclass=LinterMeta):
         See https://docs.python.org/3/library/os.path.html#os.path.expandvars
 
         """
-        def recursive_replace(vars, value):
+        def recursive_replace(variables, value):
             if isinstance(value, Mapping):
-                return {key: recursive_replace(vars, val)
+                return {key: recursive_replace(variables, val)
                         for key, val in value.items()}
             elif isinstance(value, list):
-                return [recursive_replace(vars, item)
+                return [recursive_replace(variables, item)
                         for item in value]
             elif isinstance(value, str):
                 value = os.path.expanduser(value)
                 value = os.path.expandvars(value)
-                value = sublime.expand_variables(value, vars)
+                value = sublime.expand_variables(value, variables)
                 return value
 
         window = self.view.window()
-        vars = window.extract_variables() if window else {}
+        variables = window.extract_variables() if window else {}
 
         filename = self.view.file_name()
         project_folder = self._guess_project_path(window, filename)
         if project_folder:
-            vars['folder'] = project_folder
+            variables['folder'] = project_folder
 
-        persist.debug('Available variables: {}'.format(vars))
-        return recursive_replace(vars, settings)
+        persist.debug('Available variables: {}'.format(variables))
+        return recursive_replace(variables, settings)
 
     @staticmethod
     def _guess_project_path(window, filename):
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -444,15 +444,23 @@ class Linter(metaclass=LinterMeta):
     def replace_settings_tokens(self, settings):
         """Replace tokens with values in settings.
 
-        Generally, we support all variables Sublime Text provides:
+        Settings can be a string, a mapping or a sequence,
+        and replacement is recursive.
+
+        Utilizes Sublime Text's `expand_variables` API,
+        which uses the `${varname}` syntax
+        and supports placeholders (`${varname:placeholder}`).
+
+        Support all variables Sublime Text provides:
         "packages", "platform", "file", "file_path", file_name",
         "file_base_name", "file_extension, "folder", "project", project_path",
-        "project_name", "project_base_name, "project_extension".
+        "project_name", "project_base_name, "project_extension",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I think we should try to use sublime.expand_variables here. It should be backwards compatible in the general sense but has the additional feature of specifying default values in case a value isn't specified.

It will break environment variable substitutions though, since they use the colon for namespacing.

kaste:
> It will break environment variable substitutions though, since they use the colon for namespacing.

Sure, I would have used `expand_variables`, but the env expansions are probably important. Maybe someone uses `${env:GEM_HOME}`, if that is even a thing. 

Now, for the PythonLinter, we currently actually do this:

https://github.com/SublimeLinter/SublimeLinter/blob/4cb3dd5903aa83e69b2c4a136c093fd2819eee59/lint/base_linter/python_linter.py#L209-L227

which is totally fine bc we don't do it for every setting, only for the ones where we expect a path from the user. But here in SublimeLinter core land, we touch every setting. So, we touch `excludes`, probably fine, as well as e.g. `max-line-length`, probably stupid. 

But I don't want a config to configure which settings SublimeLinter should expand. Is it risky to call `os.path.expanduser` unconditionally? Probably. However, calling `os.path.expandvars` for everything should be ok.

Now, we can extend/enhance the dict we use for `expand_variables`. Should we change our token names, to match the Sublime ones?

@braver PTAL

lint/linter.py
```diff
@@ -578,19 +545,32 @@ def recursive_replace(expressions, mutable_input, nested=False):
...
         })
 
+        persist.debug('Available expressions: {}'.format(expressions))
         recursive_replace(expressions, settings)
 
-    def merge_project_settings(self, view_settings, project_settings):
-        """
-        Return this linter's view settings merged with the current project settings.
+    @staticmethod
+    def _guess_project_path(window, filename):
```
kaste: Extracted functionality probably useful for #818

lint/linter.py
```diff
@@ -578,19 +545,32 @@ def recursive_replace(expressions, mutable_input, nested=False):
...
+        data = window.project_data()
+        if not data:
+            return
 
-        Subclasses may override this if they wish to do something more than
-        replace view settings with inline settings of the same name.
-        The settings object may be changed in place.
+        folders = data.get('folders')
```
kaste: Maybe `window.folders()` is just what we need. Have to check.

kaste: Okay. Confimed `folders()` seems to be exactly what we need here.

kaste:
Okay, PTAL this is the most breaking version, we can have. If we need a transition or so, we can add it probably from here. 

The most problematic thing is the change from ${project} -> ${folder} probably, but ${folder} is the Sublime name here, so it is actually the least surprising thing. ${project}, $project_path} points to the actual project file, which can be anywhere as discussed. Also ${directory} -> ${file_path}. The same. That's the Sublime name we use in the build or templates. On the other side ${home} -> ~, is nice.

kaste:
For testing, just add some settings, see

image

And log print the expanded settings somewhere here in linter.py

image

lint/linter.py
```diff
@@ -438,159 +435,70 @@ def get_view_settings(self):
...
-                        value = exp['token'].sub(exp['value'], value)
-
-            return value
-
-        def recursive_replace(expressions, mutable_input, nested=False):
-            for key, value in mutable_input.items():
-                mutable_input[key] = recursive_replace_value(expressions, value)
-            if nested:
-                return mutable_input
+                value = os.path.expanduser(value)
```
FichteFoll: Since `expanduser` only expands a `~` if it is the first character, I suppose it makes the most sense to have this translation last in the chain.

lint/linter.py
```diff
@@ -438,159 +435,70 @@ def get_view_settings(self):
...
+                value = os.path.expandvars(value)
```
FichteFoll: We *could* also just add `os.environ` to our `vars` dict, while having the ST variables override. `expand_variables` is case-sensitive so it will be *very* unlikely to have collisions with the ST variables.

This would then allow replacements such as `${XDG_CONFIG_HOME:$HOME/.config}/blub`. When moving expanduser to the end of the chain, `${XDG_CONFIG_HOME:~.config}/blub` even. None of this is possible currently.

Note that this won't support `%APPDATA%` replacements for Windows, but this wasn't possible previously either.

kaste:
Pulled @FichteFoll . Now this is really 🌶 

Updated description of the PR.

kaste:
image

More feature, less code. 👏

braver:
@kaste is this ready to merge (i.e. ready for me to test and update our docs) in your opinion?

capripot:
That is a great improvement 👍 Thanks for your work @kaste!

kaste:
@braver Yeah, I removed the '[WIP]'. Test and merge. This is a collab between @FichteFoll and me. So he already reviewed, and we made it 100% better since the initial PR. 

Note that other PRs rely on this feature to be in master. But you review within 7 minutes or faster anyway 😄

kaste:
Hi @capripot 👋 !

You probably noticed, I already remove your `${root}` token bc your PR inspired me to make the whole thing better.  Your `${root}` should be `${folder:$file_path}` if this lands.

lint/linter.py
```diff
@@ -438,159 +435,82 @@ def get_view_settings(self):
...
+        if not filename:
+            return folders[0]
 
-        """
-        view_settings.update(project_settings)
-        return view_settings
+        for folder in folders:
+            # Take the first one; should we take the deepest one? The shortest?
+            if folder in filename:
+                return folder
```
kaste: This 'algorithm' can be improved. But it is not necessary for the first iteration. 

The question here is: if you have a file open, and your *window* has multiple folders open, and more than one open folder contains the open file, which folder should be regarded as the 'project' folder. The deepest, the shortest, or the first?

kaste: E.g. your main aka 'first' folder is the root of your current webapp. The second folder is a npm module/dependency you control. You open and edit a file from your (own) dependency. Which folder is the project folder here?

braver: FWIW SublimeText uses the first folder for "reveal in sidebar". I think it makes sense, or is at least easy to explain, to use the first folder as primary / most important.

braver:
> Can we make sure we update (breaking) version and documentation in this PR too?

Sure, I'll merge and update the docs 👍 Perhaps even push out an upgrade message to beta users.

kaste:
🎆

shulard:
Hello,

With that changes, some of the linter README documentations became invalid because they reference the `${project}` variable.

Example: https://github.com/SublimeLinter/SublimeLinter-phpcs/blob/master/README.md

Maybe a meta issue must be created to help update and clean the corresponding READMEs ?

braver:
> Maybe a meta issue must be created to help update and clean the corresponding READMEs ?

Done #848

PR #826, opened 2018-01-16
First working version of panel multiselect.

pykong:
If multiple error regions belong to the current position they will all get highlighted within the panel.
Addresses: #803 #804

### DIFF
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -236,19 +236,24 @@ def fill_panel(window, update=False, **panel_filter):
 
 # logic for updating panel selection
 
-def get_closest_region_dict(dic, colno):
-    dics = [
-        d
-        for error_dict in dic.values() for d in error_dict
+def get_current_panel_linenos(dic, colno, is_full_line):
+    line_dicts = [
+        d for error_dict in dic.values() for d in error_dict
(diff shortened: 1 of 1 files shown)


pykong:
multi_line_select

Our panel gets more amazing by the day!

Multiselect has arrived. It runs very smooth already. Still, report any trip ups here, so I can fix 'em.

I opened the PR into the (as of now unmerged) branch it is derived from and depends on. I hope this was the right approach here git-wise.

braver:
You can set the PR to merge into master. That way we can merge this after the branch it depends on. All PR’s should merge into master, if we should do one before the other you can indicate that in a comment (as you did).

FichteFoll:
It would make the PRs harder to review though, since they contain more changes. You could for example just change the target branch once the dependant PR has been merged.

braver:
Git is nothing if not rich in options ☺️

pykong:
Yeah, git is a hundred-headed hydra. 
I have already switched target branch to master this morning.
What do we do?

braver:
Let's just focus on panel_region_sort and it'll sort itself out later

pykong:
Code will be superseded by #823

PR #829, opened 2018-01-17
Refactor panel.get_view_lines

kaste:
For @pykong with ❤️

### DIFF
--- a/panel/panel.py
+++ b/panel/panel.py
@@ -2,7 +2,6 @@ import os
 import sublime
 import bisect
 
-from ..lint.const import WARN_ERR
 from ..lint import util, persist
 
 PANEL_NAME = "SublimeLinter"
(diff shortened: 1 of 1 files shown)


panel/panel.py
```diff
@@ -163,30 +162,18 @@ def filter_ok(check_val, key, panel_filter):
...
-    for lineno, line_dict in sorted(view_dict["line_dicts"].items()):
-        items = []
-        for error_type in WARN_ERR:
-            if not filter_ok(error_type, "types", panel_filter):
-                continue
+    return sorted(
+        (item
+         for lineno, line_dict in view_dict["line_dicts"].items()
+         for error_type, items in line_dict.items()
+         for item in items
```
kaste: This block is nice bc if you change the receiving data structure, you just have to mangle the for..in unpacking here. The rest stays as is.

E.g. if you would consume `raw_errors`, you just delete the first two for..ins and basically just do a `for item in raw_errors[buffer_id]`. And you're done.

kaste: I'm done for today. 🥃

pykong: Cheers!

braver:
@pykong do you want to merge this into your branch?

pykong:
As I now started to base the panel logic on kaste's `raw_errors` this PR has become redundant.
We do not even have `panel.get_view_lines` there at all.

@kaste Should we close this PR?

PR #831, opened 2018-01-18
Specify working directory in Popen call

### LINKED ISSUE #828: util.cd is not thread aware
Due to async linting you can see something like this in action:

> Already patched
** chdir C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\User** chdir C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter

SublimeLinter: flake8: wanted executable is 'C:\Dev\python35\Scripts\flake8.exe' 
** chdir C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint
SublimeLinter: flake8: util.py ['C:\\Dev\\python35\\Scripts\\flake8.exe', '-'] 

What do you think the current dir will be after all this? T.i. when https://github.com/SublimeLinter/SublimeLinter/blob/4cb3dd5903aa83e69b2c4a136c093fd2819eee59/lint/util.py#L604-L606 runs 😒 😶

FichteFoll:
Using os.chdir is a bad choice and thread-unsafe, which bites us
when running linters concurrently.
To ensure thread safety, this function must be removed entirely,
which is a breaking change for linters using it.

I omitted tmpdir here because I have no idea how this is used
(if at all).

Fixes #828.

### DIFF
--- a/lint/base_linter/python_linter.py
+++ b/lint/base_linter/python_linter.py
@@ -259,8 +259,7 @@ def ask_pipenv(linter_name, chdir):
 @lru_cache(maxsize=None)
 def _ask_pipenv(linter_name, chdir):
     cmd = ['pipenv', '--venv']
-    with util.cd(chdir):
-        venv = _communicate(cmd).strip().split('\n')[-1]
+    venv = _communicate(cmd, cwd=chdir).strip().split('\n')[-1]
 
     if not venv:
         return
@@ -268,7 +267,7 @@ def _ask_pipenv(linter_name, chdir):
     return find_script_by_python_env(venv, linter_name)
 
 
-def _communicate(cmd):
+def _communicate(cmd, cwd):
     """Short wrapper around subprocess.check_output to eat all errors."""
     env = util.create_environment()
     info = None
@@ -281,7 +280,7 @@ def _communicate(cmd):
 
     try:
         return subprocess.check_output(
-            cmd, env=env, startupinfo=info, universal_newlines=True
+            cmd, env=env, startupinfo=info, universal_newlines=True, cwd=cwd
         )
     except Exception as err:
         persist.debug(
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -953,7 +953,6 @@ class Linter(metaclass=LinterMeta):
         chdir = settings.get('chdir', None)
 
         if chdir and os.path.isdir(chdir):
-            persist.debug('chdir has been set to: {0}'.format(chdir))
             return chdir
         else:
             if self.filename:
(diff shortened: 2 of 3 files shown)


braver:
[Our package control channel](https://github.com/SublimeLinter/package_control_channel) has a python script to clone all linter plugins in one go. It's pretty easy to check if breaking the interface really hits plugins or not. I'm not sure what to look for in this case, perhaps you can take a look yourself?

FichteFoll:
We should certainly do that. Unfortunately there is no way to have a forwards and backwards compatible way to do this without potentially duplicating a lot of code.

I'll see if I can check lintere later today.

kaste:
That's exactly the code I would have written. I'll try later.

We only change signatures in the util module! If plugin authors use these and hence their plugin breaks, it is trivial for them to fix this.

kaste:
How to test this?

Just code reviewing now on the big screen. It's dope 👍 

It is not really breaking, bc it tries to not alter the fn signatures on the linter methods. That is all you can ask for here. Usually, I would like to call `get_view_settings` and with it `get_chdir` one or two levels up and pass it down. But @FichteFoll calls it *just* before leaving `linter.py` and entering `util.py`. `linter.run` (17 times) and `linter.communicate` (4 times) are both used as public API. I don't like it, but that's how it is.

kaste:
My test strategy: I open multiple windows with `.eslintignore` files.

From the eslint docs (emphasis mine):

> When ESLint is run, it looks **in the current working directory** to find an .eslintignore file before determining which files to lint. If this file is found, then those preferences are applied when traversing directories. Only one .eslintignore file can be used at a time, so .eslintignore files **other than the one in the current working directory** will not be used.

Now, I check if eslint ignores the correct files. It does.


LGTM 👍 Merge

braver:
I don't think I can seriously test this any better than you already did. If you want me to I'll try to give it a go tomorrow, but it's also fine by me if you just hit the merge button.

kaste:
I'm not prepared yet to hit that button 💅

FichteFoll:
I initially had the call to `get_chdir` in the same place as previously, but realized that I ended up passing `cwd` around a lot without actually doing anything with it. Since it only depends on the state of the linter, I resorted to just putting it at the least possible point before leaving `linter.py` and the class's state.

I think ideally the cwd would get read together with the command and other settings and store it in an instance attribute, but I didn't feel like figuring out where the best place would be as I don't know about the exact message flow and where polymorphism from linter implementations could bite us. This can trivially be refactored later.

Also note that the only test I did was whether my setup still worked. I also haven't checked yet which linters used `util.cd`, which I removed.

FichteFoll:
The only linter using `util.cd` is flake8 in its master branch, i.e. the SL3 backwards-compat. This is safe to merge,

kaste:
My first merge. 💃 🕺

PR #832, opened 2018-01-18
Use tempdir as working dir fallback

FichteFoll:
Just expanding the "current dir" is a) unnecessary, since it would just be the current directory anyway, and b) unpredictable. If the working dir cannot be specified in any other way, i.e. for unsaved views, we fall back to the tempdir instead.

This is just the "ultimate fallback" version. Ideally, the project folder should be attempted first, but my proposal in #818 would require #825 to be merged first.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -7,6 +7,7 @@ import os
 import re
 import shlex
 import sublime
+import tempfile
 
 from . import highlight, persist, util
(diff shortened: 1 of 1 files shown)


kaste:
I'm 👎 on this one, see https://github.com/SublimeLinter/SublimeLinter/issues/818#issuecomment-358590593

braver:
See  #845

PR #835, opened 2018-01-18
Revise "Setting tokens" docs

FichteFoll:
- Rename them to "Settings Expansion"
- Fix usages of backticks
- Add link to build system docs for variable explanations
- Be more explicit about how things work and what to watch out for

Iterates on https://github.com/SublimeLinter/SublimeLinter/commit/965fba0445a5b7c68b886e5fd229e93c15ae697e.

### DIFF
--- a/docs/linter_settings.rst
+++ b/docs/linter_settings.rst
@@ -42,7 +42,7 @@ The value must be a string, corresponding to a valid directory path.
         "chdir": "${project}"
     }
 
-With the above example, the linter will get invoked from the ``${project}`` directory (see :ref:`Setting Tokens <settings-tokens>` for more info on using tokens).
+With the above example, the linter will get invoked from the ``${project}`` directory (see :ref:`Settings Expansion <settings-expansion>` for more info on using variables).
 
 .. note::
 
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -3,13 +3,11 @@ Settings
 
 Settings stack
 --------------
-SublimeLinter merges settings from several sources to calculate the value.
+SublimeLinter merges settings from several sources to calculate the value. Settings are merged in the following order:
 
-.. code-block:: none
-
-    Default settings
-    User settings
-    Project settings
+1. Default settings
+1. User settings
+1. Project settings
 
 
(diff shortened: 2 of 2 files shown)


braver:
Thanks!

There are more docs that need a once-over. I tried my best to make them less verbose and more to the point, remove duplications and outdated examples, but it may have ended up a bit cryptic in places. If anyone feels up to it... 😁

FichteFoll:
Yes, I'm confident the docs will need a huge pass before the eventual release of 4.0.

Just something to consider: semantic linefeeds are *really* for writing text inside a VCS environment where single line-breaks are collapsed, such as reStructured Text. I use them for the undocs and it's been a bliss ever since we converted most our docs to using those.

https://github.com/guillermooo/sublime-undocs/blob/sublime-text-3/CONTRIBUTING.md#line-widths

braver:
Oh most of that is from the old docs. I just turned wrapping on to deal with it. Most of my negative line count contributions come from weeding out that jungle.

FichteFoll:
Just checking whether I can split lines where I want them for potential future PRs against the docs~ I'll take that as permission to do so.

docs/linter_settings.rst
```diff
@@ -42,7 +42,7 @@ The value must be a string, corresponding to a valid directory path.
         "chdir": "${project}"
     }
 
-With the above example, the linter will get invoked from the ``${project}`` directory (see :ref:`Setting Tokens <settings-tokens>` for more info on using tokens).
+With the above example, the linter will get invoked from the ``${project}`` directory (see :ref:`Settings Expansion <settings-expansion>` for more info on using variables).
```
kaste: project -> folder

FichteFoll: The example still lists `${project}`, so the description is correct. I suspect we'd want this to just show the default, though, don't we?

kaste: Ah well, it's just that end-users get confused bc they relate ${project} to be the base dir of their app. But it's a Sublime name here referring to the sublime project file. End-users usually are looking for ${folder}.

PR #841, opened 2018-01-19
Fix: Abort lint if no executable can be found.

### LINKED ISSUE #836: Funny: if `get_cmd` signals we should abort linting, we lint anyway
Don't run `git blame` 😨 😔

kaste:
Fixes #836

Fixed two comments.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -874,16 +874,14 @@ class Linter(metaclass=LinterMeta):
             return self.default_type
 
     def lint(self, code, hit_time):
-        """
-        Perform the lint, retrieve the results, and add marks to the view.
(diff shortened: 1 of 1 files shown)


kaste:
Only this is relevant:

image

We just clearly express our API at the top. `self.cmd`can be None which has a special meaning. Otherwise we call `get_cmd()` which either returns a command or None. This logic was hidden, and actually the reason for the bug. If you can't see it, bc things which are related are spread across, you get bugs.

braver:
Check, verified. I have a project with a local install of jshint and don't have a global jshint. If I opened a js file outside this project it would report that no jshint could be found and then still go ahead and try to run it on the file, resulting in multiple errors. This fixes that. Awesome 🔥

PR #842, opened 2018-01-19
Merge into shared Sublime thread when calling the callback

### LINKED ISSUE #839: Merge backend threads into the shared Sublime thread when done
We cannot and don't want to guarantee that the lint callback is thread aware or even 'safe'. If we merge into the shared thread, Sublime runs all the queued tasks sequentially.

kaste:
Fixes #839

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -41,7 +41,8 @@ def lint_view(view, hit_time, callback):
 
     all_errors = chain.from_iterable(results)
 
-    callback(view, list(all_errors), hit_time)
+    sublime.set_timeout_async(
+        lambda: callback(view, list(all_errors), hit_time))
 
 
 def run_concurrently(max_workers=2):
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -48,29 +48,6 @@ def lint_view(view, hit_time, callback):
         lambda: callback(view, list(all_errors), hit_time))
 
 
-def run_concurrently(max_workers=2):
-    def run(tasks):
-        with ThreadPoolExecutor(max_workers=max_workers) as executor:
-            work = [executor.submit(task) for task in tasks]
-            yield from await_futures(work)
-
-    return run
-
-
-def await_futures(fs, ordered=False):
-    if ordered:
-        done, _ = wait(fs)
-    else:
-        done = as_completed(fs)
-
-    for future in done:
-        try:
-            yield future.result()
-        except Exception:
-            ...
-            traceback.print_exc()
-
-
 def execute_lint_task(linter, code, offset, hit_time):
     errors = linter.lint(code, hit_time) or []
     translate_lineno_and_column(errors, offset)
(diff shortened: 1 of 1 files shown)


lint/backend.py
```diff
@@ -41,7 +41,8 @@ def lint_view(view, hit_time, callback):
 
     all_errors = chain.from_iterable(results)
 
-    callback(view, list(all_errors), hit_time)
+    sublime.set_timeout_async(
+        lambda: callback(view, list(all_errors), hit_time))
```
FichteFoll: I generally propose using `functools.partial` when a function is only invoked with arguments and no other logic is necessary. Lambdas are evalualted lazily which *can* in some occasions, particularly within loops, produce unexpected results.

FichteFoll:
Maybe add a comment mentioning why `set_timeout_async` is preferred?

kaste:
Added a comment. This 'functional' stuff does not yet match the style of this repo. (And this `lambda` here is totally save.)

You remember my `lrucache` trick? This is subtle. But I actually got a comment from you. And I used dropwhile/takewhile, and was asked why I loop twice. I hope you understand. How I write code matches the style of the code given, and of course transforms it bit-by-bit. It should work, and it should fit.

Ah, we'll see. This is not the last time I hit this file.

braver:
Guys, let's try this. Give me a holler if you think this is ready for merging and I'll do a final check, perhaps update some docs, or do something useful like updating plugins if needed, and hit the merge button. Most of these PR's are not hitting the UI and I totally trust you both with the architecture, refactoring and judging impact. I do try follow along, but I don't have the time to wrap my head around this and participate in a useful way, and so I cannot really see if you expect to add commits or comments to a PR.

So, just request a review, label it some way or mention me and I'll come in and push the green button.

kaste:
LGTM. Merging. 

- It's near weekend
- This exact line will meet me again bc we have another issue here open
- It does work. @FichteFoll addresses a stylistic thing. He says, if we do the `partial` thing it is safer when juniors adopt this. I say, this is more welcoming when juniors read this.
- It fixes a bug.

kaste:
Generally, it is good for my workflow, if somebody/you says LGTM. Then I quickly check again, and either merge or change and request again.

Bc we had these early merges between us, I add '[WIP]' if it's not ready but I need feedback and remove the '[WIP]' somewhen.

braver:
Excellent. I totally think we should keep at least a minimum of review on everything, even if it’s just 7 minutes worth. I’ll check in on stuff that’s not labeled “WIP” and add a LGTM if it is so.

FichteFoll:
My comment was not a functional comment, indeed, and merely stylistic. I still believe it holds value and over the past I have grown to adapt that pattern in order to prevent potential issues, but I won't create a PR just for this. I mean, it works and it's fine, but I personally wouldn't write it that way and I'd like you to know why.

PR #844, opened 2018-01-19
[docs] Change chdir example value to the default

FichteFoll:
`${project}` is most likely not what users want, so we better not suggest to use that.


braver:
Thanks.

How would you feel if I converted all the docs to markdown? Any experience with readthedocs in markdown? I already use too many languages on a given day and rst is breaking my brain.

FichteFoll:
I've only used RTD with reST so far. I know facelessuser is using mkdocs, which is markdown-backed. We talked about that briefly in the #general chat on Discord a while ago.

PR #845, opened 2018-01-19
Set the working dir by default to the project root dir

kaste:
As discussed in #818 

- Does not mangle with names, yet.

This works for me (I think), the rest is name bike-shedding.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -856,16 +856,20 @@ class Linter(metaclass=LinterMeta):
                     options[name] = value
 
     def get_chdir(self, settings):
-        """Find the chdir to use with the linter."""
+        """Return the working dir for this lint."""
         chdir = settings.get('chdir', None)
 
         if chdir and os.path.isdir(chdir):
-            return chdir
-        else:
-            if self.filename:
-                return os.path.dirname(self.filename)
+            if os.path.isdir(chdir):
+                return chdir
             else:
-                return os.path.realpath('.')
+                persist.printf(
+                    "{}: WARNING: wanted working_dir '{}' is not a dir"
+                    "".format(self.name, chdir)
+                )
+                return None
+
+        return self._guess_project_path(self.view.window(), self.view.file_name())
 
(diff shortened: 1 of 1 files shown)


braver:
@kaste and @FichteFoll should we close #832 in favor of this?

kaste:
Yes, I think #832 can be closed. We never need to chdir to a tempdir for such cases here, bc the linter will not find any config files in tempdirs and its parents. This was just a bit confusing in the working and thinking about this. 

Our fallback is to **not** chdir at all if we can't guess anything reasonable. E.g. when the user opens a file in a fresh sublime window. No folders, no project, nothing. Then we don't chdir to tempdir, or anything we do nothing. And actually, it depends how you started sublime, this could be `.`.

FichteFoll:
Yes, we can close #832 over this. As I mentioned somewhere, the idea to use a tempdir was just to provide a predictable working dir as opposed to none at all and to have potential cache files be created in that predictable folder, but I don't know how many linters actually do this and for mypy I already use a different method.

kaste:
Hm, this is immediately useful bc it fixes eslint bugs about eslint ignores not being processed correctly. 

Currently, this is like a fix. A good one. The name-bikeshedding stuff. Should this setting be called `working_dir`, should we change the method name to `get_working_dir`, we can decide this later. 

WDYT @braver @FichteFoll

braver:
We can bike shed till the end times and it won’t make this thing work better. We shouldn’t treat every PR as an opportunity for massive refactoring, let’s just merge stuff if it improves things. If we find an opportunity for bike shedding let’s not block a fix for it but just put it in an issue and return to it later.
Merge at will IMO

FichteFoll:
Merge now, track renaming of the setting (and method) in #818. This is an immediate change that doesn't depend on a rename since the setting already exists.

kaste:
No, I'll do this tomorrow bc tomorrow is Monday and right now it's still weekend.

PR #846, opened 2018-01-19
Implement `executable` setting in core Linter

### LINKED ISSUE #781: [beta] Enable configuring specific executable for NodeLinter
I think we now have this ability for the PythonLinter, does that also work for the node based linters? Would solve https://github.com/SublimeLinter/SublimeLinter-eslint/issues/147 and similar use cases.

kaste:
The executable setting will be recognized by *all* linters. It is implemented in `context_sensitive_executable_path` so the behavior is pluggable. However, it also means that plugins which override that method must be patched to see the new functionality.
    
Fixes #781

### DIFF
--- a/lint/base_linter/composer_linter.py
+++ b/lint/base_linter/composer_linter.py
@@ -43,6 +43,12 @@ class ComposerLinter(linter.Linter):
 
         Return a tuple of (have_path, path).
         """
+        # The default implementation will look for a user defined `executable`
+        # setting.
+        success, executable = super().context_sensitive_executable_path(cmd)
+        if success:
+            return success, executable
+
         local_cmd = None
         global_cmd = util.which(cmd[0])
 
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -678,17 +678,38 @@ class Linter(metaclass=LinterMeta):
         return self.insert_args(cmd)
 
     def context_sensitive_executable_path(self, cmd):
-        """
-        Calculate the context-sensitive executable path, return a tuple of (have_path, path).
+        """Calculate the context-sensitive executable path.
 
-        Subclasses may override this to return a special path.
+        Subclasses may override this to return a special path. The default
+        implementation looks for a setting `executable` and if set will use
+        that.
 
         Return (True, '<path>') if you can resolve the executable given at cmd[0]
         Return (True, None) if you want to skip the linter
         Return (False, None) if you want to kick in the default implementation
             of SublimeLinter
 
+        Notable: `<path>` can be a list/tuple or str
+
         """
+        settings = self.get_view_settings()
+        executable = settings.get('executable', None)
+        if executable:
+            persist.debug(
+                "{}: wanted executable is '{}'".format(self.name, executable)
+            )
+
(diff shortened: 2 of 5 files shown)


braver:
- SublimeLinter-flow seems to have overridden this exactly to get this feature. We should update this one after SL4 is released.
- SublimeLinter-contrib-epcomp is less clear to me, but seems the same and we should open an issue there.

kaste:
- SublimeLinter-flow, well, they have it under a different name *flow-bin*. This change doesn't break them. 💅 

- SublimeLinter-contrib-epcomp (I actually guided this implementation last year 😆 ) uses *epbin* and it works

I would ask if others already use this exact name 'executable'. Note, when they implement `context_sensitive...` (what shit name is that for an API?) they are automatically safe here *unless* they call super which none of the plugins do. That was part of the implementation strategy here. 

We actually have a problem if they **don't** override the method with the name I always copy paste around bc nobody can write it **and** also have a setting with the name we use namely **executable** **and** it has a different meaning. If it has the same meaning t.i. in code speak to replace `cmd[0]` we're still safe we just do the same work twice.

lint/base_linter/composer_linter.py
```diff
@@ -43,6 +43,12 @@ def context_sensitive_executable_path(self, cmd):
 
         Return a tuple of (have_path, path).
         """
+        # The default implementation will look for a user defined `executable`
+        # setting.
+        success, executable = super().context_sensitive_executable_path(cmd)
+        if success:
+            return success, executable
+
```
kaste: This kind of duplication is not super nice but I think this is called OOP. 😉

lint/linter.py
```diff
@@ -678,17 +678,38 @@ def build_cmd(self, cmd):
...
+        implementation looks for a setting `executable` and if set will use
+        that.
 
         Return (True, '<path>') if you can resolve the executable given at cmd[0]
         Return (True, None) if you want to skip the linter
         Return (False, None) if you want to kick in the default implementation
             of SublimeLinter
 
+        Notable: `<path>` can be a list/tuple or str
+
```
kaste: This is a tricky, nobody uses this yet, feature. This is a bit I tried to see the future 🎱  You can set `executable` to `/bin/flake8`, boring stuff. But you can also set it to `['pipenv', 'run', 'flake8']`. 

Unfortunately, I never had the time to test this.

braver: > Unfortunately, I never had the time to test this.

Me neither. I guess as soon as someone tries we'll find out if it's successful?

lint/base_linter/composer_linter.py
braver: I don't know what the kids call it these days, but I actually like a healthy amount of duplication.

braver:
> I would ask if others already use this exact name 'executable'.

SublimeLinter-contrib-gcc and SublimeLinter-flake8 do but it has the same meaning, ie. "don't look for anything just run this thing".

I guess we're good then. 🎉

kaste:
Ahem, you say LGTM, I merge. Otherwise often I 😰

braver:
Haha, ok 👍🏻

PR #852, opened 2018-01-19
Clean up for status bar indicator code.

pykong:
- Enhanced readability.
- Cleaner, more pythonic code.

### DIFF
--- a/busy_indicator_view.py
+++ b/busy_indicator_view.py
@@ -56,13 +56,10 @@ class UpdateState(sublime_plugin.EventListener):
         draw(**State)
 
 
-indicators = [
-    'Linting.  ',
-    'Linting.. ',
-    'Linting. .',
-    'Linting ..',
-    'Linting  .',
-]
+def tick():
+    while True:
+        for phase in ('.  ', '.. ', '. .' ' ..', '  .'):
+            yield "Linting" + phase
 
 
 def draw(active_view, running, **kwargs):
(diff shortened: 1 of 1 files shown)


pykong:
Generators for the rescue!

braver:
Eh, we must've improved performance again because I can't get the darn thing to show up anymore. Code looks fine to me.

busy_indicator_view.py
```diff
@@ -56,23 +56,18 @@ def on_activated_async(self, active_view):
...
-    'Linting.  ',
-    'Linting.. ',
-    'Linting. .',
-    'Linting ..',
-    'Linting  .',
-]
+def tick():
+    while True:
+        for phase in ('.  ', '.. ', '. .' ' ..', '  .'):
+            yield "Linting" + phase
```
kaste: In python you actually do this with `itertools.cycle`

```
phaser = cycle(['.  ', '.. ', '. .' ' ..', '  .'])
```

And then you call next(phaser)

busy_indicator_view.py
```diff
@@ -56,23 +56,18 @@ def on_activated_async(self, active_view):
...
 
 def draw(active_view, running, **kwargs):
     buffer_id = active_view.buffer_id()
     start_time = running.get(buffer_id, None)
     now = time.time()
     if start_time and (INITIAL_DELAY <= (now - start_time) < TIMEOUT):
-        num = len(indicators)
-        text = indicators[int((now - start_time) * 1000 / CYCLE_TIME) % num]
-        active_view.set_status(STATUS_BUSY_KEY, text)
+        active_view.set_status(STATUS_BUSY_KEY, next(tick()))
```
kaste: This doesn't work bc every time you draw you start a new generator `tick()` and take one value out of it via `next()`.

busy_indicator_view.py
```diff
@@ -56,23 +56,18 @@ def on_activated_async(self, active_view):
...
         sublime.set_timeout_async(lambda: draw(**State), CYCLE_TIME)
```
kaste: If you do animations, probably in JS or so, you always reference a timer. `now` `then` etc.

Bc you actually never know when you're callback runs. So you actually don't need a cycler that gives you a new frame per call. You have to write a function that gives you the correct frame at given time.

busy_indicator_view.py
pykong: @kaste +1 for cycle My previous indicator was implemented exactly that way ()!
And you ripped it out. ;-)

pykong:
OK. Was a suggestion. Let's move on to other PRs.

PR #851, opened 2018-01-19
Status bar messages sourced from persist.raw_errors.

pykong:
Earliest working version

For @kaste with <3

PR #853, opened 2018-01-19
[WIP] - Migrating to new style errors

pykong:
**Currently WIP**

Requires prior merging of: #823 #851
- removed persist.errors everywhere else except status_bar_view and `panel`
- early work on switching to raw_errors in `sublime_linter.open_tooltip`

Fixes:  #734


sublime_linter.py
```diff
@@ -398,31 +383,27 @@ def open_tooltip(self, active_view=None, point=None, is_inline=False):
...
-            show_count = True
+        if not is_gutter:  # do not show tooltip on hovering empty gutter
+            errors = [e for e in errors if e["start"] <= col <= e["end"]]
 
-        if not line_dict:
+        if not errors:
             return
 
-        we_count = persist.errors.get_we_count_line(vid, lineno)
+        we_count =  # where to get those from?
```
kaste: That's actually relatively simple. But I go to bed 🥃 and start my weekend.

pykong: You owe it!

kaste:
After a rebase you have to git push -f otherwise GitHub hates you

pykong:
Thanks. I'll put that into my notebook.

What about refactoring the popup logic? And put it into its own controller, like we did with panel and status bar? In a different PR though.

kaste:
Yes, that would be nice

lint/linter.py
```diff
@@ -1111,7 +1127,9 @@ def clear_view(view):
 
         view.erase_status(STATUS_KEY)
         highlight.clear_view(view)
-        persist.errors.pop(view.id(), None)
+        vid = view.buffer_id()
+        if vid in persist.raw_errors:
+            del persist.raw_errors[vid]
```
pykong: should be bid

kaste: Actually this whole function should go away. Just popping bid doesn't work, bc you can have multiple views into the same bid. You can delete this only if it is the last view.

But otoh we just don't need this here.

lint/linter.py
```diff
@@ -569,17 +587,6 @@ def remove(cls, vid):
...
-        for vid, linters in persist.view_linters.items():
-            for linter in linters:
-                linter.clear()
-                persist.view_linters[vid].remove(linter)
-                linter_class = persist.linter_classes[linter.name]
-                linter = linter_class(linter.view, linter.syntax)
-                persist.view_linters[vid].add(linter)
-
     @classmethod
     def clear_all(cls):
```
pykong: @kaste Can we remove this?

kaste: You need to check the callers

pykong: Better said should we clear `raw_errors` here instead?
It is only called after the syntax map changes. But I also need to go through all linter plugins to see whether they harbor any callers.

pykong: I have left the `clear` statement in, even after renaming `raw_errors` to `errors`.

pykong:
LGTM. What do you guys think?

kaste:
You need to work on your git skills, @pykong. You need it for your career anyway. To ask for a review, when there are 90 commits and 50 files changed, doesn't work if you have co-workers, colleagues.

braver:
Yeah this is impossible to review right now. Just look at the files changed tab yourself: is all of that your change set? Did you rebase and then merge again, or what happened? I can do some first aid on your branch if you want, but you definitely need to practice syncing up with master.

pykong:
Sorry guys. I indeed should have made clearer that I needed some more instructions on how to fix that mess, after doing my first rebasing. **LGTM** meant regarding the actual changes made. 
Would #853 work for you?

PR #856, opened 2018-01-22
Cleanup outdated PythonLinter documentation

kaste:
I'm not good at `rst` but all this documentation is outdated.


braver:
I'm even worse at it. Intend to migrate to markdown this week, perhaps even today. 

Thanks for this anyway 👍

PR #858, opened 2018-01-22
Hotfix: `util.tmpdir` should not use `os.chdir` on its own

kaste:
Funfact: it didn't even try to switch back to the original dir.

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -475,8 +475,7 @@ def tmpdir(cmd, files, filename, code, output_stream=STREAM_STDOUT, env=None):
             else:
                 shutil.copyfile(f, target)
 
-        os.chdir(d)
-        out = communicate(cmd, output_stream=output_stream, env=env)
+        out = communicate(cmd, output_stream=output_stream, env=env, cwd=d)
 
         if out:
             # filter results from build to just this filename


braver:
I'm not sure how to verify that this works, but it looks simple enough to just merge?

braver:
Oh, I get this now:

```py
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/backend.py", line 158, in await_futures
    yield future.result()
  File "./python3.3/concurrent/futures/_base.py", line 394, in result
  File "./python3.3/concurrent/futures/_base.py", line 353, in __get_result
  File "./python3.3/concurrent/futures/thread.py", line 54, in run
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/backend.py", line 39, in <lambda>
    lambda: execute_lint_task(*task, hit_time=hit_time)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/backend.py", line 52, in execute_lint_task
    errors = linter.lint(code, hit_time) or []
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 922, in lint
    output = self.run(cmd, code)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1282, in run
    return self.tmpfile(cmd, code)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1352, in tmpfile
    cwd=cwd)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/util.py", line 441, in tmpfile
    os.remove(path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/cp/659wm5bs1pl2kpz8j9jvckrw0000gn/T/SublimeLinter3-koenlageveen/util.py'
```

That's including this hot fix.

kaste:
Okidoki. I don't see how this is related? I changed `tmpdir` and you throw on `tmpfile`.

kaste:
Well, the code in `tmpfile` is also sloppy? https://github.com/SublimeLinter/SublimeLinter/blob/8b46320aca3d61bc76f88b315e1f1cab78c548f1/lint/util.py#L422

tempdir is a constant, so how do you guarantee that multiple linters will not trash this?

Which linter is that btw?

kaste:
Sure, but do you get the error all the time or sometimes?

> I’m running pylint pycodestyle pydocstyle and Flake8.

And which one doesn't work with stdin?

braver:
My guess would be pylint, but I’d have to check and am not at my computer right now

FichteFoll:
The only linter that uses this function is SublimeLinter-contrib-gometalinter.
I'm not even sure when this function is useful as opposed to just using `tmpfile` since it only copies the single file anyway, and not the entire project for example (which would be pretty imperformant). Maybe we don't need it at all?

kaste:
@FichteFoll No, it copies `files`

FichteFoll:
Oh nvm, I missed that the function takes a list of files.

braver:
Ok, sorry for the hijack. Moved the bug report to #864

PR #859, opened 2018-01-22
Rename setting `chdir` to `working_dir`

kaste:
Follow up to #845 

** BREAKING **

**This needs a banner on install.** 

```
Renamed setting `chdir` to `working_dir`. The new default is `${folder:$file_path}`, which means it chooses the project root if available, and falls back to use the directory of the file you're currently editing.  
```

### DIFF
--- a/docs/linter_settings.rst
+++ b/docs/linter_settings.rst
@@ -34,8 +34,7 @@ working_dir
 -----
 This setting specifies the linter working directory.
 
-The value must be a string, corresponding to a valid directory path. Variable
-substitutions are supported.
+The value must be a string, corresponding to a valid directory path.
 
 .. code-block:: json
 
--- a/docs/linter_settings.rst
+++ b/docs/linter_settings.rst
@@ -30,16 +30,17 @@ The value may be a string or an array. If it is a string, it will be parsed as i
 The default value is an empty array.
 
 
-chdir
+working_dir
 -----
 This setting specifies the linter working directory.
 
-The value must be a string, corresponding to a valid directory path.
+The value must be a string, corresponding to a valid directory path. Variable
+substitutions are supported.
 
 .. code-block:: json
 
     {
-        "chdir": "${folder:$file_path}"
+        "working_dir": "${folder:$file_path}"
     }
 
 With the above example,
@@ -49,11 +50,7 @@ or the file's directory if it is not contained within a project folder
 
 .. note::
 
-     If the value of ``chdir`` is unspecified (or inaccessible), then:
-
-     - If linting a saved file, the directory is set to that of the linted file
-
-     - If linting an unsaved file, the directory is unchanged
+     The above example is the default.
 
 
 excludes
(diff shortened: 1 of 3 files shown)


kaste:
@FichteFoll PTAL

braver:
There is a 4.0.0.rc1.txt in messages that should contain (short) description of all important changes between SL3 and SL4. We haven't done messages for each beta release, although perhaps this risks a bit more breakage than some others and also warrants a separate message.

docs/linter_settings.rst
```diff
@@ -30,16 +30,17 @@ The value may be a string or an array. If it is a string, it will be parsed as i
...
 
 
-chdir
+working_dir
 -----
 This setting specifies the linter working directory.
 
-The value must be a string, corresponding to a valid directory path.
+The value must be a string, corresponding to a valid directory path. Variable
+substitutions are supported.
```
FichteFoll: Does this need an explicit mention? We replace substitutions in all settings.

docs/linter_settings.rst
```diff
@@ -49,11 +50,7 @@ or the file's directory if it is not contained within a project folder
 
 .. note::
 
-     If the value of ``chdir`` is unspecified (or inaccessible), then:
-
-     - If linting a saved file, the directory is set to that of the linted file
-
-     - If linting an unsaved file, the directory is unchanged
+     The above example is the default.
```
FichteFoll: Imo it should still be mentioned what happens if the resulting `working_dir` ends up being empty, i.e. it won't be changed. With the default setting, this is only relevant for unsaved views, but if the setting is modified and `$file_path` is removed, for example, this could occur more frequently.

lint/linter.py
```diff
@@ -874,17 +874,17 @@ def build_options(self, options, type_map, transform=None):
...
 
-        if chdir:
-            if os.path.isdir(chdir):
-                return chdir
+        if cwd:
+            if os.path.isdir(cwd):
+                return cwd
             else:
                 persist.printf(
                     "{}: WARNING: wanted working_dir '{}' is not a dir"
```
FichteFoll: Not changed in this PR, but I suggest `is not a directory`.

FichteFoll:
Regarding the banner: `${folder, $file_path}` should be `${folder:$file_path}`.

docs/linter_settings.rst
kaste: Done

lint/linter.py
kaste: Done

docs/linter_settings.rst
kaste: Here, I dunno. If you set `working_dir` to a falsy value, the default kicks in. Any truthy value passes the first if check. If `working_dir` is set to something which is not a directory we will err and just not use this as `cwd`. We pass `None`.

Generally, unsaved files are not a problem. Only editing an unsaved file in a window without any open folder.

braver:
Ok, looks good. @kaste wanna merge this? I'm going to move the docs to markdown today, probably easier to have this merged before.

PR #860, opened 2018-01-22
Implement `env` setting

### LINKED ISSUE #507: Setting to set environment variables for linters
For example for `mypy` I would need to set the `MYPYPATH` environment variable.

### LINKED ISSUE #847: Fishy code in RubyLinter
In RubyLinter we do:

https://github.com/SublimeLinter/SublimeLinter/blob/527b8c56c3d3f43b05e301e99a51c3d667d04e4c/lint/base_linter/ruby_linter.py#L119-L128

But, when we use popen we pass in (an enhanced) `os.environ` as `env` anyway. Here we take GEM_HOME and pass it unmodified back. What am I missing?

kaste:
Fixes #507 

Fixes #847 

Related discussion #847

### DIFF
--- a/lint/base_linter/ruby_linter.py
+++ b/lint/base_linter/ruby_linter.py
@@ -119,18 +119,4 @@ class RubyLinter(linter.Linter):
         else:
             ruby_cmd = [ruby]
 
-        # Attention readers! All self mutations can have surprising
-        # side-effects for concurrent/async linting.
-
-        # Don't use GEM_HOME with rbenv, it prevents it from using gem shims
-        if rbenv:
-            self.env = {}
-        else:
-            gem_home = os.environ.get('GEM_HOME', None)
-
-            if gem_home:
-                self.env = {'GEM_HOME': gem_home}
-            else:
-                self.env = {}
-
         return True, ruby_cmd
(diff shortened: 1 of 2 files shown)


FichteFoll:
What if a linter wants to set an environment variable on its own? For example, PythonLinters would like to have `PYTHONIOENCODING` set (for Windows mostly) so they can reliably parse UTF-8-encoded output.

Imo `self.env` is useful for these situation and doesn't need to be deprecated. It can be overridden by the user's settings anyway.

kaste:
Plugin authors can - just like with `working_dir` - set a default: 

```
 defaults: {
	'env': {'GEM_HOME': '~/foo/bar'}
 }
```

or override `get_environment` which is the OOP standard here.  `self` is not safe.

FichteFoll:
Good point, I forgot about those. 👍 
I suppose, all Python linters should include `PYTHONIOENCODING` in their defaults then.

braver:
Nice!

kaste:
Okay, rebased on master. Did I hear a LGTM from everyone?

braver:
Yeah, LGTM here 🤘🏻

PR #861, opened 2018-01-22
Remove python paths

kaste:
This is unused code. `git blame` for yourself.


kaste:
> git blame for yourself.

sounds a bit negative here, but this is really funny, if you git blame, you actually find that this breakage is a collaborative effort. Everyone currently involved shows up. 😄

braver:
Go team!

PR #865, opened 2018-01-22
Remove lru_cache from which

### LINKED ISSUE #863: `util.which` shouldn't use a dumb lrucache
We already had to remove lrucaches bc they can cache the wrong results. Here is another one:

https://github.com/SublimeLinter/SublimeLinter/blob/8b46320aca3d61bc76f88b315e1f1cab78c548f1/lint/util.py#L242-L243

kaste:
Generally, lru_cache is safe for pure functions and otherwise not.

Fixes #863

### DIFF
--- a/lint/settings.py
+++ b/lint/settings.py
@@ -148,10 +148,6 @@ class Settings:
             from . import style
             style.StyleParser()()
 
-        # Clear the path-related caches if the paths list has changed
-        if "paths" in self.changeset:
-            util.clear_path_caches()
-
         # If the syntax map changed, reassign linters to all views
         from .linter import Linter
 
--- a/lint/util.py
+++ b/lint/util.py
@@ -239,7 +239,6 @@ def can_exec(path):
     return os.path.isfile(path) and os.access(path, os.X_OK)
 
 
-@lru_cache(maxsize=None)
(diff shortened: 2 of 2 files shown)


braver:
I’d be very careful about caching anything unless we’re really sure performance is unacceptable without. You can assume very little about what happens outside SL and even inside SL. Cache invalidation and naming things...

braver:
So that’s LGTM from me

kaste:
You always can run into the problem that you have Sublime open, it tells you that *eslint* is not installed locally. You quickly install it, and then SublimeLinter will not catch up ... and actually you have to restart Sublime.

FichteFoll:
The last remaining usage of `lru_cache` in that file is `find_file`.

https://github.com/kaste/SublimeLinter3/blob/e7712619df4d824971c2e8591ec3c1ec746f56b1/lint/util.py#L162

This is probably going to be called rather frequently and could take a longer time depending on directory size, so I guess caching there seems okay, but unlimited caches are rarely a good idea and it does't seem like this cache is ever cleared either.

braver:
If blame doesn't provide info about why the caching was added, we can remove it until we find out why not caching that causes problems?

PR #866, opened 2018-01-22
Prepare complete runtime environment within `linter.get_environment`

kaste:
Note that `util.create_environment()` has our enhanced PATH and this
augmented environment is not used for lint execution anymore.

The augmented PATH is only used for `which` calls.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -15,6 +15,20 @@ ARG_RE = re.compile(r'(?P<prefix>@|--?)?(?P<name>[@\w][\w\-]*)(?:(?P<joiner>[=:]
 NEAR_RE_TEMPLATE = r'(?<!"){}({}){}(?!")'
 BASE_CLASSES = ('PythonLinter',)
 
+# Many linters use stdin, and we convert text to utf-8
+# before sending to stdin, so we have to make sure stdin
+# in the target executable is looking for utf-8. Some
+# linters (like ruby) need to have LANG and/or LC_CTYPE
+# set as well.
+UTF8_ENV_VARS = {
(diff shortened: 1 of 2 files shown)


lint/linter.py
```diff
@@ -15,6 +15,20 @@
...
+# in the target executable is looking for utf-8. Some
+# linters (like ruby) need to have LANG and/or LC_CTYPE
+# set as well.
+UTF8_ENV_VARS = {
+    'PYTHONIOENCODING': 'utf8',
+    'LANG': 'en_US.UTF-8',
+    'LC_CTYPE': 'en_US.UTF-8',
+}
+
+BASE_LINT_ENVIRONMENT = ChainMap(UTF8_ENV_VARS, os.environ)
```
kaste: Generally, this is nice to read code.

In this version, the environment during **runtime** would be `os.environ` plus a few *utf8* settings. Of course, we could also augment our PATH here. I just don't know if that's necessary. 

We usually use the augmented path for our `which` calls. And this is used if we actually have a real executable. *The runtime environment.*

lint/linter.py
```diff
@@ -895,7 +909,7 @@ def get_working_dir(self, settings):
 
     def get_environment(self, settings):
         """Return runtime environment for this lint."""
-        return ChainMap({}, settings.get('env', {}), self.env)
+        return ChainMap({}, settings.get('env', {}), self.env, BASE_LINT_ENVIRONMENT)
```
kaste: This is also nice to read. But note, that if you subclass, and use super and augment that, you can easily get the wrong shadowing.

braver:
LGTM 👍

PR #867, opened 2018-01-23
Avoid scroll if region is already visible on goto error

gerardroche:
If the region we're moving to is already visible, then we don't want
the view to suddenly scroll. If the region is not visible, then we want
the surrounding area of the region to be visible.

### DIFF
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -83,11 +83,31 @@ def goto(view, direction, count, wrap):
     move_to(view, line, start)
 
 
+class _sublime_linter_goto_line(sublime_plugin.TextCommand):
+
+    def run(self, edit, line, col):
+        pt = self.view.text_point(line, col)
+        self.view.sel().clear()
+        self.view.sel().add(pt)
+        self.view.show(pt)
+
+
 def move_to(view, line, col):
     window = view.window()
-    filename = view.file_name() or "<untitled {}>".format(view.buffer_id())
-    target = "{}:{}:{}".format(filename, line + 1, col + 1)
-    window.open_file(target, sublime.ENCODED_POSITION)
+    if view == window.active_view():
+        # If the region we're moving to is already visible, then we don't want
+        # the view to suddenly scroll. If the region is not visible, then we
+        # want the surrounding area of the region to be visible.
+        # We need to a use a custom goto line command for several reasons:
+        # * ST's goto line command doesn't accept a col argument.
+        # * SL requires that on_selection_modified events MUST be triggered for
+        #   each move.
+        # See https://github.com/SublimeLinter/SublimeLinter/pull/867.
+        view.run_command('_sublime_linter_goto_line', {'line': line, 'col': col})
+    else:
+        filename = view.file_name() or "<untitled {}>".format(view.buffer_id())
+        target = "{}:{}:{}".format(filename, line + 1, col + 1)
+        window.open_file(target, sublime.ENCODED_POSITION)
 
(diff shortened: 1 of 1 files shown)


kaste:
The requirement here is that `on_selection_modified` events MUST be triggered by Sublime for each move. Can you check if your method does fulfill that?

gerardroche:
You're right. It works for NeoVintageous but not if not if you just use the keybindings.  

For example these won't work:

```
    { "keys": ["ctrl+n"], "command": "sublime_linter_goto_error",
      "args": { "direction": "next" }
    },
    { "keys": ["ctrl+p"], "command": "sublime_linter_goto_error",
      "args": { "direction": "previous" }
    },
```

They kinda work: the cursor doesn't move until you press an arrow key and then the cursor suddenly jumps to the correct position.

I was going to use the goto line command `view.run_command("goto_line", {"line": line})` but that doesn't accept a column argument. One way is to add a command that does the same thing but also accepts a column argument:

```
class _sublime_linter_goto_line_col(sublime_plugin.TextCommand):
    def run(self, edit, line, col):
        pt = self.view.text_point(line, col)
        self.view.sel().clear()
        self.view.sel().add(pt)
        self.view.show(pt)
```

So the `move_to` command would look like this:

```
def move_to(view, line, col):
    window = view.window()
    if view == window.active_view():
        # If the region we're moving to is already visible, then we don't  want
        # the view to suddenly scroll. If the region is not visible, then we
        # want the surrounding area of the region to be visible.
        # pt = view.text_point(line, col)
        # view.sel().clear()
        # view.sel().add(pt)
        # view.show(pt)
        view.run_command('_sublime_linter_goto_line_col', {'line': line, 'col': col})
    else:
        filename = view.file_name() or "<untitled {}>".format(view.buffer_id())
        target = "{}:{}:{}".format(filename, line + 1, col + 1)
        window.open_file(target, sublime.ENCODED_POSITION)
```

It can obviously be refactored. It does mean adding that extra command though, which I don't like. Any ideas on how best do it, I don't know enough about the internals of SL?

kaste:
You did not update the PR with your new proposed solution. I'm also a bit confused if the proposed changes from your comment, actually work or not. Requirements: We must actually scroll the point into the viewport and *on selection modified* must trigger.

FichteFoll:
I suspect that the introduction of a TextCommand to perform the selection modifications will cause `on_selection_modified` to be triggered.

gerardroche:
Yeah I didn't update the pr. I wanted to hear what the opinion of the suggested solution was first. I've updated the pr. 

The introduction of the TextCommand does indeed seem to trigger the `on_selection_modified` event. I've tested it and it works. I still don't properly understand why the previous code doesn't work, but like I said I don't know enough about the internals. The *only* thing I don't like is the additional command. It's no big deal really, it just feels less than ideal. It's a pity ST's goto line command doesn't accept a column argument.

kaste:
Jesus, @gerardroche , that was probably a lucky guess, but now we know Sublime is the new InternetExplorer and you deserve the 👑 

(That's exactly the functionality I/we wanted to implement, but failed at doing so.)

kaste:
That's a perfect example why I would like to have dependencies for plugins. This MoveTo command is useful for everyone. We would publish it, declare it as a dependency, everybody wins.

FichteFoll:
See https://github.com/SublimeTextIssues/Core/issues/485#issuecomment-337480388 for an explanation.

PR #868, opened 2018-01-23
[WIP] Markdown docs

braver:
I think more people are familiar with markdown (I certainly am), it being the default for readme's etc., than reStructuredText. It would mean we lose some features, but the docs could be a bit simpler and easier to maintain anyway. 

This is an awful diff to read, but just to let people know it's happening here's the PR.


braver:
Well, damn, the markdown parser is super limited and it looks like it's a dead project for them (last change in May). *This* close to just putting everything in the wiki here.

kaste:
Then don't do it. I mean `rst` is more capable, the better format, and it's not that hard. I used it before, it's just obviously not that wide spread.

braver:
Yeah, you're right. Gotta go lear another language then.

FichteFoll:
>Dang, the highlighting is bad though.

One of these days I'll rewrite it. Soon. Eventually. Probably when I do some serious work on the undocs again.

As I mentioned somewhere, instead of only switching one page to markdown, imo the entire docs should be converted to markdown and then `mkdocs` could be used to generate the html representation. I know that rtd supports these too, since facelessuser does that for his packages.

That said, I'm perfectly fine with rst and don't see a need to replace it. The only reason I *could* see was if extending sphinx with custom roles was too complicated and there were better alternatives, but this is certainly not needed for this project.

braver:
Sure, the intire thing would be markdown, and I would like that because I don’t enjoy markup languages at all and I definitely don’t see why the world needs more than one.. However, I don’t control any of the hosting or whatever and don’t feel like touching that. It’s not bad to have it in rst, but I hoped for a somewhat easy migration and rtd just has too lame md support. So whatever, this is also good. Not like there is anything fancy going on.

PR #870, opened 2018-01-23
Fix temporary files, remove temporary folders

### LINKED ISSUE #864: SL-pylint doesn't work on unsaved buffers, throws errors
```py
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/backend.py", line 158, in await_futures
    yield future.result()
  File "./python3.3/concurrent/futures/_base.py", line 394, in result
  File "./python3.3/concurrent/futures/_base.py", line 353, in __get_result
  File "./python3.3/concurrent/futures/thread.py", line 54, in run
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/backend.py", line 39, in <lambda>
    lambda: execute_lint_task(*task, hit_time=hit_time)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/backend.py", line 52, in execute_lint_task
    errors = linter.lint(code, hit_time) or []
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 922, in lint
    output = self.run(cmd, code)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1282, in run
    return self.tmpfile(cmd, code)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1352, in tmpfile
    cwd=cwd)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/util.py", line 415, in tmpfile
    os.remove(path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/cp/659wm5bs1pl2kpz8j9jvckrw0000gn/T/SublimeLinter3-koenlageveen/linter.py'
```

The temp dir exists, but the tempfile doesn't and hence cannot be removed. I only get this with SublimeLinter-pylint, and always as soon as I edit a file (linters running in "background").

braver:
Fixes #864 by creating a random temp file using `tempfile.NamedTemporaryFile`, so that every linter gets his own file to work on an delete when done. 

The same problem could occur in `tmpdir` if multiple linters create and then remove the same temporary directory. The only linter to use this feature is SublimeLinter-contrib-gometalinter, but only for go < 1.5 (it apparently doesn't even try to support background linting mode for newer versions). Leaving it in a broken state is not an option IMO. I therefore propose to drop the feature altogether instead trying to fix it.

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -335,11 +335,12 @@ def tmpfile(cmd, code, filename, suffix='', output_stream=STREAM_STDOUT, env=Non
     if suffix:
         filename = os.path.splitext(filename)[0] + suffix
 
+    file = tempfile.NamedTemporaryFile(delete=False)
+    path = file.name
+
     try:
-        f = tempfile.NamedTemporaryFile(delete=False)
-        f.write(bytes(code, 'UTF-8'))
-        f.close()
-        path = f.name
+        file.write(bytes(code, 'UTF-8'))
+        file.close()
 
         cmd = list(cmd)
 
--- a/lint/util.py
+++ b/lint/util.py
@@ -383,15 +383,11 @@ def tmpfile(cmd, code, filename, suffix='', output_stream=STREAM_STDOUT, env=Non
     if suffix:
         filename = os.path.splitext(filename)[0] + suffix
 
-    path = os.path.join(tempdir, filename)
-
     try:
-        with open(path, mode='wb') as f:
-            if isinstance(code, str):
-                code = code.encode('utf-8')
-
-            f.write(code)
-            f.flush()
+        f = tempfile.NamedTemporaryFile(delete=False)
+        f.write(bytes(code, 'UTF-8'))
+        f.close()
+        path = f.name
 
         cmd = list(cmd)
 
--- a/lint/util.py
+++ b/lint/util.py
@@ -4,34 +4,20 @@ from functools import lru_cache
 import locale
 from numbers import Number
 import os
-import getpass
 import re
-import shutil
-import stat
 import sublime
 import subprocess
 import tempfile
 
-
 from copy import deepcopy
 
 from .const import WARNING, ERROR
-
-if sublime.platform() != 'windows':
-    import pwd
-
 
 STREAM_STDOUT = 1
 STREAM_STDERR = 2
 STREAM_BOTH = STREAM_STDOUT + STREAM_STDERR
 
 ANSI_COLOR_RE = re.compile(r'\033\[[0-9;]*m')
-
-# Temp directory used to store temp files for linting
-tempdir = os.path.join(
-    tempfile.gettempdir(),
-    "SublimeLinter3-" + getpass.getuser()
-)
 
 
 def printf(*args):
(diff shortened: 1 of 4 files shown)


FichteFoll:
>The same problem could occur in tmpdir if multiple linters create and then remove the same temporary directory. 

Directories created by `tempfile.TemporaryDirectory` are guaranteed to be unique, so unless the entire temp st-specific folder is cleared once linting is finished (why would it?), this shouldn't result in errors. Linter runners (like `tmpfile`) should just remove their files once linting is done. They should also get a unique file.

Generally, I like grouping of SL's temporary files in its own directory. It is going to potentially create many of these files (and folders), so being able to remove them all at once sounds like an advantage. Currently this doesn't really occur, but it is possible once linters start managing their own linting directories (like I do on mypy for its cache) and they could then use SL's own temporary directory for that, exposed in its API.

braver:
Oh, sure, having a tempdir-like thing in the API is a good idea. Keeping this one around just isn't very interesting because nobody uses it. Currently the only linter we know that uses temporary directories (mypy) isn't using ours. If there is demand for it we can make a new one. Right now I really feel like removing something Aparajita built 5 years ago that nobody uses, even if it *does work (potentially)*.

lint/util.py
```diff
@@ -335,12 +335,11 @@ def tmpfile(cmd, code, filename, suffix='', output_stream=STREAM_STDOUT, env=Non
     if suffix:
         filename = os.path.splitext(filename)[0] + suffix
 
-    try:
-        f = tempfile.NamedTemporaryFile(delete=False)
-        f.write(bytes(code, 'UTF-8'))
-        f.close()
-        path = f.name
+    with tempfile.NamedTemporaryFile(delete=False) as file:
+        file.write(bytes(code, 'UTF-8'))
```
braver: @FichteFoll like so you mean?

lint/util.py
```diff
@@ -383,15 +383,11 @@ def tmpfile(cmd, code, filename, suffix='', output_stream=STREAM_STDOUT, env=Non
...
-    path = os.path.join(tempdir, filename)
-
     try:
-        with open(path, mode='wb') as f:
-            if isinstance(code, str):
-                code = code.encode('utf-8')
-
-            f.write(code)
-            f.flush()
+        f = tempfile.NamedTemporaryFile(delete=False)
```
kaste: @FichteFoll is trolling you. Just move the `f = ` line outside of the `try` and probably the `path = `at the top of the `try` block, and you're good to go.

kaste:
The `tmpdir` was added just for *Go* and nobody else uses it. I don't think we should maintain something that hasn't paid off. But, I would file an issue 

```
Intent to remove tmpdir in SL4

You're the only linter using `tmpdir`. We do not believe that this is a very performant solution to linting, and don't want to maintain this functionality in the future. We're about to drop this for SL4. 

You're free to copy-paste all the relevant code into this repo, or drop the feature as well.

Sorry
```

Generally, I feel the linter authors should implement something interesting on their own, and then we can decide if we pull it in. But after reading a lot of linter code, actually the only relevant contribution came from my *flake8* implementation. (E.g. there is no state-of-the-art ruby linter, but ours sucked.)

braver:
> I would file an issue

Will do.

> Generally, I feel the linter authors should implement something interesting on their own, and then we can decide if we pull it in

I think that's a wise approach indeed.

kaste:
So I would squash the 1st and 3rd commit. And leave the drop tmpdir one open. You can point them to the open PR, so they actually know what things they should copy-paste. Then move on after whatever 1 week.

lint/util.py
```diff
@@ -335,11 +335,12 @@ def tmpfile(cmd, code, filename, suffix='', output_stream=STREAM_STDOUT, env=Non
     if suffix:
         filename = os.path.splitext(filename)[0] + suffix
 
+    file = tempfile.NamedTemporaryFile(delete=False)
+    path = file.name
```
braver: After fixing stupid bugs in bash scripts today I am now fully confident I don't know what I am doing more ofter than I do, even if I think I do.

braver:
I referenced the removal in [the issue](https://github.com/alecthomas/SublimeLinter-contrib-gometalinter/issues/11). The time between now and SL4's final release is enough prior notice I think.

I personally don't really care about the order of commits, but I get your point. It's small enough of a change to try to keep it clean. I'll take that as a LGTM and will push the squashed commits outside this PR.

FichteFoll:
Uhm, has this been superseded?

I support @kaste's take on this. I'm fine with removing the `tmpdir`, but a rationale for that would have been nice in the commit.

The only thing I'm more or less unhappy about is the removal of the "shared" tempdir, which, when exposed in an API, I could have used in mypy and get rid of all my custom handling there. Or I just move  that to the core, if there's interest?

braver:
The commits are now here:

https://github.com/SublimeLinter/SublimeLinter/commit/a3db729367c0ceb423a3f617ab3fa8ab4b739409
https://github.com/SublimeLinter/SublimeLinter/commit/e61b935c391df53ac5f3fe104c095d58eb99e11a

Perhaps there should have been more info in the commit, but the sole affected party has been notified. But again, even there it did not work for recent versions of the language it intends to lint. So a non-working, non-used feature has been removed, along with more than I can count. 🤷🏼‍♂️

I’m not against restoring a shared tempdir feature if there is a case like mypy for it. Why not move your work there into core? That sounds like a great idea. If the basic concept is useful, and > 0 linters use it, and it’s implemented well, I’m all for it.

FichteFoll:
Noted.

kaste:
@FichteFoll Your cleanup code is like 5 lines.

FichteFoll:
It's not long, but it feels dirty.

kaste:
*You* create those directories in your code using the raw python `tempfile` API.  It is dirty if someone else does the cleanup, I could argue as well.

PR #871, opened 2018-01-23
provide version info via package metadata

### LINKED ISSUE #855: Log SL version if debug mode is enabled
When the user enables debug mode, SL's version should be logged for better debuggability.
I suggest basing this on a `package-metadata.json` file, if available, or the git revision, if `.git` is available. I don't exactly recommend hard-coding a version string in the code somewhere, but that *could* be used as well.

braver:
Fixes #855 by creating `util.get_sl_version()`. Version number, if known, is printed in debug mode. I guess this also makes the version available to plugins, but unlike the ST build number it isn't a number and cannot be converted to one so you can't really target e.g. > 4.0.0 without shenanigans.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -46,7 +46,7 @@ def plugin_loaded():
     persist.plugin_is_loaded = True
     persist.settings.load()
     persist.debug("debug mode: on")
-
+    persist.debug(util.get_sl_version())
     style.load_gutter_icons()
     style.StyleParser()()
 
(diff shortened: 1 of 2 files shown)


sublime_linter.py
```diff
@@ -46,7 +46,7 @@ def plugin_loaded():
     persist.plugin_is_loaded = True
     persist.settings.load()
     persist.debug("debug mode: on")
-
+    persist.debug(util.get_sl_version())
```
FichteFoll: `persist.debug("version: " + util.get_sl_version())`?

kaste: Do this, then ship.

braver: 👌🏻

braver:
83adac2d728c9b81e7ced942813414b845dcf406

PR #822, opened 2018-01-16
Autonomous, self-contained panel

kaste:
For @pykong 

Follow up to #817 

Fixes #827

PR #872, opened 2018-01-23
Fix paths in panel

### LINKED ISSUE #727: [beta] enable path shortening with untitled views
Now, path shortening is skipped if there are untitled views. Previously it tried something but failed (at least on macOS) and produced this:

screen shot 2017-12-20 at 10 19 23

(if `panel.py` line 41 is `common_path = os.path.commonprefix([path + '/' for path in paths if path and not path.startswith("<untitled")])`).

Ideally path shortening works all the time and doesn't mangle the "path" of untitled views.

kaste:
Fixes regression where paths were not shortened regardless of untilted views.

Fixes #727

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -225,12 +225,19 @@ def filter_and_sort(buf_errors, panel_filter):
     return sort_errors(buf_errors)
 
 
-def get_window_raw_errors(window, errors, panel_filter):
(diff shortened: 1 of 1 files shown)


kaste:
blocked by #822

FichteFoll:
👍 I can remove this from my todo then.

PR #875, opened 2018-01-23
Allow opening the panel early on start up

### LINKED ISSUE #811: [beta] Cannot open panel at start up
I don't know why but you cannot open the panel on startup. You have to wait until the initial active_view() has been linted.

kaste:
Fixes #811


kaste:
blocked by #822

Issue #792, opened 2018-01-12
Render lint errors as they arrive 


kaste:
The new async backend runs linters in parallel. E.g. it runs `flake8` and `mypy` concurrently even for the same file/view. However, it waits for both to finish (fan-in), before it passes the errors back to our views which then update, e.g. draw gutter marks.

We could instead pass the errors as they arrive. This would result in multiple renders, so it could be that we put too much pressure on Sublime.

kaste:
Oh, you know that I don't get notifications for 👍 

So, shut up. I want to leave my notes here.

The implementation here is not completely trivial. Currently the data structure is almost flat, we only group by `buffer_id`. If we don't fan-in, we must merge the data in our callback. For (view, linter) pairs this is simple bc they translate to (buffer_id, linter_name). However, this doesn't work for our selector based, virtual view based, lintings. E.g. we a html file with two script tags. We run eslint twice in parallel, but from outside we cannot have a differentiator here, we'd need (bid, linter_name, x) here, but I don't see it. 

So currently it seems that we need a group by (linter) operation, where we still fan-in in the backend, and a merge operation in the callback. E.g. the html file above, we start *three* `lint`s concurrently, html once, and eslint twice. But we call the callback only two times, once for html, once for eslint.

FichteFoll:
`x` could be the point at which the region to be linted starts. For whole-file lints, it would just be 0.

PR #884, opened 2018-01-24
Prevent lint races in lint_view

FichteFoll:
I was running into an issue where I had `yamllint` executed up to 15 times for a single file for no apparent reason. I could re-lint only the affected file (`ctrl+k, l`) and would receive a variable amounts of errors for the same lines. In the console, I could confirm that the linter was indeed run multiple times and all the results were combined. This hints at a race condition.

Then I realized I had this "fix" up in my stashes, which I made when I noticed the usage of `lambda` in this situation, which I know from PackageDev can turn out detrimental. Turns out that this seems to fix the described problem, as I'm unable to reproduce it anymore after mutliple attempts.

See commit message for a technical explanation.

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -3,6 +3,7 @@ import sublime
 from concurrent.futures import ThreadPoolExecutor, as_completed, wait
 from fnmatch import fnmatch
 from itertools import chain
+from functools import partial
 import os
 import traceback
 
@@ -34,9 +35,8 @@ def lint_view(view, hit_time, callback):
     linters = get_linters(view)
     lint_tasks = get_lint_tasks(linters, view)
 
-    run = run_concurrently(max_workers=5)
-    results = run(
-        lambda: execute_lint_task(*task, hit_time=hit_time)
+    results = run_concurrently(
+        partial(execute_lint_task, *task, hit_time=hit_time)
         for task in lint_tasks)
 
     all_errors = chain.from_iterable(results)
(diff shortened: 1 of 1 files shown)


kaste:
Isn't this the correct test

```
    run = run_concurrently(max_workers=5)
    test = run(
        lambda: print('**', i)
        for i in range(5))
    list(test)
```

FichteFoll:
You need more items than workers to trigger the race.

```py
from concurrent.futures import ThreadPoolExecutor, as_completed, wait


def run_concurrently(tasks, max_workers=5):
    with ThreadPoolExecutor(max_workers=max_workers) as executor:
        work = [executor.submit(task) for task in tasks]
        yield from await_futures(work)


def await_futures(fs, ordered=False):
    if ordered:
        done, _ = wait(fs)
    else:
        done = as_completed(fs)

    for future in done:
        try:
            yield future.result()
        except Exception:
            import traceback
            traceback.print_exc()


results = run_concurrently(
    lambda: print('**', i)
    for i in range(15)
)
list(results)
```

```
** 0
** 1
** 2
** 3
** 4
** 14
** 14
** 14
** 14
** 14
** 14
** 14
** 14
** 14
** 14
```

kaste:
Yeah, okay, I'm interested in the fix, not in the stylistic stuff.

kaste:
Okay, I changed my mind while thinking about #792. I need `run_concurrently` anyway as one function, not as a factory.

All or nothing. 👍

braver:
✨

PR #885, opened 2018-01-24
Prototype/idea: Implement `toast` messages

kaste:
🌶 

This PR implements toast messages. For demo purposes, the Goto commands `ctrl-k, n` flashes toast messages, if there are no more problems to jump to etc. Also, here and then, we simulate a big red config error.

Implementation details are not important. I did this somewhen, and just rebased it now, to show it. 😄

### DIFF
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -2,7 +2,7 @@ import sublime
 import sublime_plugin
 
 from itertools import dropwhile, takewhile
-from .lint import persist
+from .lint import persist, events
 
 
 """
@@ -100,3 +100,5 @@ def get_current_pos(view):
 def flash(view, msg):
     window = view.window() or sublime.active_window()
     window.status_message(msg)
+
+    events.broadcast(events.TOAST, {'message': msg})
--- a/lint/events.py
+++ b/lint/events.py
@@ -4,6 +4,7 @@ import traceback
 
 BEGIN_LINTING = 'BEGIN_LINTING'
 FINISHED_LINTING = 'FINISHED_LINTING'
+TOAST = 'SHOW_TOAST'
 
 
 listeners = defaultdict(set)
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -678,6 +678,14 @@ class Linter(metaclass=LinterMeta):
             util.printf('WARNING: {} cannot locate \'{}\''.format(self.name, which))
             return None
 
+        # TOAST error example
+        from . import events
+        import random
+        if random.randrange(0, 3, 1) == 0:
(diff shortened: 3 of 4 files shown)


braver:
🔥 !

I'm in general no fan of adding UI patterns like this, but I must say this is really nifty.

kaste:
I was surprised as well.

FichteFoll:
Why go through events instead of exeuting `show_toast` directly?

kaste:
@FichteFoll Bc then I have to import `toast_view` into `linter`

braver:
FYI, while this is certainly very cool (very very cool), I don't intend to merge. Not sure if that was even the goal. I definitely think Sublime Text could use some more fancy feedback UI options than flashing some text on the statusbar. And I also think creativity from package authors could help push the envelope. But I don't like having such creative usage of the popup API in SL.

So perhaps we could put this behind an "experimental" flag if it's nicely self-contained. Or we could post something on the forums every week about how cool this is and how much ST needs an api for 🍞➡️🔥 messages.

pykong:
I hurts to miss so much awesomeness. But I am with braver here.

Albeit for other reasons. I rather like to keep SublimeLinter lean after we removed so much code.
And like it to ship it with fewer lines of code rather than more.

While I understand @braver's conservative stance on UI design, generally we should not shy away from new patterns  if they are well implemented and make sense. That is still the case with the toast messages. Keep in mind we have been innovating in Sublime's UI all the way. In minor things like the look of the status bar indicator or the two way synchronization of view and panel selection.

Another idea: How difficult would it be to open SublimeLinter for plugins?
Maybe we could expose `Events` and `errors` as a minimal API so other plugins could hook into that.
Thereby the toast messages could move to their own plugin as well.

kaste:
@pykong That's actually the road we go here. We try to make all views as independent, self-contained as possible, And then we look at their imports to see what they actually need, what API is useful for the views. And we can always take a view and make a real plugin out of it.

For the toasts here, the real implementation would have been of course to put it in a separate plugin. But we still need cooperation from SublimeLinter bc it has to send events the toast plugin will react to.

braver:
@kaste can you store this idea in a personal repo somewhere, and open issues here if you need changes to make this work as a plugin? I'm closing this because the separate plugin is definitely the way to go here IMO, not because it isn't awesome (because it is 🔥 )

kaste:
I just hope Github doesn't delete this bc it is my backup here until I get some time to do this.

braver:
The PR itself will never be removed. The PR doesn’t contain the code though, it just references it, so you should store the commits somewhere safe.

PR #887, opened 2018-01-25
Render lint errors as soon as possible

### LINKED ISSUE #792: Render lint errors as they arrive

kaste:
Fixes #792

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -33,11 +33,9 @@ def lint_view(view, hit_time, next):
     the corresponding section is linted as embedded code.
     """
     linters = get_linters(view)
-    lint_tasks = get_lint_tasks(linters, view)
+    lint_tasks = get_lint_tasks(linters, view, hit_time)
 
-    results = run_concurrently(
-        partial(execute_lint_task, *task, hit_time=hit_time)
-        for task in lint_tasks)
+    results = run_concurrently(lint_tasks)
 
     all_errors = list(chain.from_iterable(results))
 
@@ -45,6 +43,13 @@ def lint_view(view, hit_time, next):
     # So we merge here into Sublime's shared worker thread. Sublime guarantees
     # here to execute all scheduled tasks ordered and sequentially.
     sublime.set_timeout_async(lambda: next(all_errors))
+
+
+def get_lint_tasks(linters, view, hit_time):
+    for (linter, region) in get_lint_regions(linters, view):
+        code = view.substr(region)
+        offset = view.rowcol(region.begin())
+        yield partial(execute_lint_task, linter, code, offset, hit_time)
 
 
 def execute_lint_task(linter, code, offset, hit_time):
@@ -69,13 +74,6 @@ def translate_lineno_and_column(errors, offset):
                 'start': error['start'] + col_offset,
                 'end': error['end'] + col_offset
             })
-
-
-def get_lint_tasks(linters, view):
-    for (linter, region) in get_lint_regions(linters, view):
-        code = view.substr(region)
-        offset = view.rowcol(region.begin())
-        yield (linter, code, offset)
 
 
 def get_lint_regions(linters, view):
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -1,6 +1,5 @@
 import sublime
 
-from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor, as_completed, wait
 from fnmatch import fnmatch
 from itertools import chain
@@ -36,13 +35,9 @@ def lint_view(view, hit_time, next):
     linters = get_linters(view)
     lint_tasks = get_lint_tasks(linters, view, hit_time)
 
-    tasks_by_linter = defaultdict(list)
-    for linter, task in lint_tasks:
-        tasks_by_linter[linter].append(task)
-
     run_concurrently(
         partial(run_tasks, tasks, next=partial(next, linter))
-        for linter, tasks in tasks_by_linter.items()
+        for linter, tasks in lint_tasks
     )
 
 
@@ -57,10 +52,14 @@ def run_tasks(tasks, next):
 
 
 def get_lint_tasks(linters, view, hit_time):
-    for (linter, region) in get_lint_regions(linters, view):
-        code = view.substr(region)
-        offset = view.rowcol(region.begin())
-        yield linter, partial(execute_lint_task, linter, code, offset, hit_time)
+    for (linter, regions) in get_lint_regions(linters, view):
+
+        def make_task(region):
+            code = view.substr(region)
+            offset = view.rowcol(region.begin())
+            return partial(execute_lint_task, linter, code, offset, hit_time)
+
+        yield linter, map(make_task, regions)
 
 
 def execute_lint_task(linter, code, offset, hit_time):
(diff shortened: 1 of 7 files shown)


lint/backend.py
```diff
@@ -33,18 +34,33 @@ def lint_view(view, hit_time, callback):
...
 
-    all_errors = chain.from_iterable(results)
+    run_concurrently(
+        partial(run_tasks, tasks, next=partial(next, linter))
+        for linter, tasks in tasks_by_linter.items()
+    )
+
+
+def run_tasks(tasks, next):
+    results = run_concurrently(tasks)
```
FichteFoll: Just a note that this means we could have up to 25 threads linting concurrently. That's probably fine, but seems excessive.

Also, why do we need to group tasks by linter again?

lint/backend.py
```diff
@@ -33,18 +34,33 @@ def lint_view(view, hit_time, callback):
     the corresponding section is linted as embedded code.
     """
     linters = get_linters(view)
-    lint_tasks = get_lint_tasks(linters, view)
+    lint_tasks = get_lint_tasks(linters, view, hit_time)
 
-    results = run_concurrently(
-        partial(execute_lint_task, *task, hit_time=hit_time)
-        for task in lint_tasks)
+    tasks_by_linter = defaultdict(list)
```
FichteFoll: `itertools.groupby`

Or just do that in `get_lint_tasks`.

lint/backend.py
```diff
@@ -13,7 +14,7 @@
 WILDCARD_SYNTAX = '*'
 
 
-def lint_view(view, hit_time, callback):
+def lint_view(view, hit_time, next):
```
FichteFoll: `next` is a built-in function.

kaste: I know that. But it's also the canonical name for the continuation function.

FichteFoll:
Well, I guess I was a bit slow with the review here.

kaste:
So I lost you all with the `partial`s I guess. 😄 

> Just a note that this means we could have up to 25 threads linting concurrently. That's probably fine, but seems excessive.

The theoretical max here is 5*5 threads. For that, you need 5 different linters each having at least 5 tasks, on a single view/file. So, that's at least not your typical *js* file. I don't know if you can reach that limit if you maybe do a ipython notebook/literate programming.

> Also, why do we need to group tasks by linter again?

The pair (buffer_id, linter_name) identifies the ongoing processes here. The concurrent processing of different buffers is done in `queue`. Here, in `lint_view(buffer_id)`, we differentiate by *linter(_name)*. 

The better id would be (buffer_id, linter_name, code), but we don't have it. The code would need to transform into that direction.

FichteFoll:
I'd suggest Region instead of code for the identifier here. Might be worth to stuff that in a data structure class (named tuple? Mable mutable?) for easier data flow. Then we can continue to use the initial thread worker pool of 5 and optionally increase that number. 

This isn't exactly about the number of threads rather than the redundancy of an additional layer imho.

FichteFoll:
*maybe

Editing comments on the mobile website is just too much effort ~

kaste:
The region is not a stable identifier. After each edit, you likely have different regions/offsets. (It is only stable *within* `lint_view` itself.) Now if you draw a line from ideal *uuid* to *rnd*, I see it as follows:

```
   uuid() ---------------------------------------- random()
                                            ^ region
                        ^ buffer_id   
                  ^ linter_name 
                    ^ file_name
```

FichteFoll:
Maybe I'm not at the top of the linting process lifecycle, but I thought that outdated linter processes (i.e. those that had their underlying buffer changed) were invalidated, so I considered the region to be immutable for the purposes of identification.

If the PR does things satisfactorily, I'm fine with merging it as is though. My comments were merely aimed at thinking about a possibly better implementation, but that shouldn't affect any outside API and can thus be an isolated change made at a later point in time.

kaste:
Ah, cool. No, I don't invalidate here.

braver:
I'm running an extreme setup here with a handful of python linters with overlapping reports, but the multiple re-renders of highlights and gutter icons is really annoying.

blink

Reducing the number of linters I'm running to something sensible makes improves matters a lot. I think I'm getting more redraws that on master, but I'm just fooling around and not sure how much it matters in actual use.

braver:
Testing this branch on a big (3k lines) typescript (tslint isn't super fast) file I get the following behaviour. Note how the highlight jumps around trying to catch up with the movement of the code.

high-kaste

The same sequence of events, but now on master. Note how the highlight sticks to its correct position.

high-master

kaste:
Then, works as intended. Ship it 😄

braver:
Haha, I don’t doubt it 😄

braver:
So uhh, where do you want to go with this?

kaste:
Maybe I should open a feature branch for this.

Anyway, here you should only test if this implementation burns your CPU, forgets some linters etc. 

Also of interest would be if the panel can handle the load.

braver:
> burns your CPU

Seems to burn a few more cycles, but from my limited time with it it doesn't seem to be a problem (on my machine)

> forgets some linters

Doesn't seem like it

> if the panel can handle the load

Seems to be able to keep up just fine. Because it's just secondary information the multiple paints aren't a problem so much as with the highlights.

kaste:
PTAL again.

I think I messed up the protected regions. But this wasn't always right before as well, so I'm not completely sure. Otherwise, I don't see the flicker and region dances.

kaste:
Redid the protected regions thing. Now I at least know why it wasn't working very reliable before.

braver:
No flickers, no dances, seems to update well, doesn't seem to thrash my CPU. What's not to like?

2 things I see now:
- Disabling a linter doesn't remove the highlights and gutter icons. Enabling a linter doesn't run it until you hit save again. On master disabling/enabling is effective instantaneously.
- Style updates require a restart of SublimeLinter.

So reacting to settings changes seems to require some restoration, but the core behaviour seems to be pretty solid.

kaste:
Cannot reproduce the style updates thing: I update from underline to outline, or icon from dot to circle and it actually updates everything correct within a second or so.

*Disabling* indeed does not work, and I don't know where to fix that. Is that an edge case, so we should tackle it in settings.on_update? Or should the backend fire events even though it didn't do anything (which is the new behavior)?

kaste:
No, it's an edge case in *backend*. I thought doing less is more, but that's bullshit we can have say script tags within html. Then we delete the scripts, and of course the backend must fire 'hello, we have updates, there are no more lint errors.'

braver:
I gave it another whirl. Looks really solid now. 

One thing happend, will have to check if it's a regression here. I close the settings window and it threw me this:

```
reloading /Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 530, in on_activated_async
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 106, in on_activated_async
    self.hit(view)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 255, in hit
    self.check_syntax(view)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 280, in check_syntax
    Linter.assign(view, reset=True)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 534, in assign
    persist.debug("detected syntax: " + syntax)
TypeError: Can't convert 'NoneType' object to str implicitly
```

kaste:
I don't get that on master nor here.

braver:
> I don't get that on master nor here.

braver:
I get it on master to. If I open my ST user settings and quickly close it again a bunch of times it'll eventually bork out. Nothing to do with this PR then. Which is LGTM really. 🚢 it?

kaste:
no risk, no fun

braver:
💪

PR #898, opened 2018-01-27
New style errors migration

pykong:
Successor of: #853


braver:
Much better 👍🏻

sublime_linter.py
```diff
@@ -332,22 +316,25 @@ def join_msgs(cls, line_dict, we_count, show_count=False):
                 <div>{messages}</div>
             '''
 
-        template = "{linter}: {code} - {escaped_msg}"
-        template_no_code = "{linter}: {escaped_msg}"
+        tmpl_with_code = "{linter}: {code} - {escaped_msg}"
+        tmpl_sans_code = "{linter}: {escaped_msg}"
 
         all_msgs = ""
         for error_type in WARN_ERR:
```
kaste: The nested for-loop is probably not necessary.just `for error in errors`?

sublime_linter.py
```diff
@@ -138,7 +134,6 @@ def on_pre_close(self, view):
             self.loaded_views,
             self.linted_views,
             self.view_syntax,
-            persist.errors,
```
kaste: We need to add some logic to remove `bid` from `errors` when it's not used anymore.

lint/linter.py
```diff
@@ -1127,7 +1127,6 @@ def clear_view(view):
 
         view.erase_status(STATUS_KEY)
         highlight.clear_view(view)
-        persist.errors.pop(view.id(), None)
```
kaste: `clear_view`, `clear_all` and `clear` should go somewhere, prob to hell?. 

But we can have a different PR for this stuff. Probably an issue so we don't forget.

sublime_linter.py
kaste: This can be tricky. Generally we can multiple views into the same buffer. So we can remove only if this view here is the last one. But one has to check/test how we can iterate here over window.views() for example. Is this view here still in window.views(). What if the user closes the window?

So, that's probably tricky, then you should make an issue

braver:
👍🏻I’ll put it through its paces tomorrow, (unless this flue thing really catches on)

kaste:
@pykong There are 4 merge commits here, you can get rid of them too.

FichteFoll:
You can actually rebase and squash the whole thing into one commit. The commits don't seem to be atomic (`WIP` in message) and some even repeated with a random "stash" commit in there as well.

braver:
Yeah, just focus on the code, I was intending to squash-merge this anyway.

sublime_linter.py
```diff
@@ -332,22 +316,25 @@ def join_msgs(cls, line_dict, we_count, show_count=False):
...
                 continue
+
+            msg_list = sorted(msg_list, key=lambda x: (x["start"], x["end"]))
+            count = len(msg_list)
+
             for item in msg_list:
-                item["escaped_msg"] = html.escape(item["msg"], quote=False)
-                template_to_fill = template if item.get('code') else template_no_code
-                error_type_msgs.append(template_to_fill.format(**item))
+                msg = html.escape(item["msg"], quote=False)
```
pykong: Message sanitation here again. Maybe we should put that near the `rstrip` we moved previously.
(In another PR.)

pykong:
Just for my understanding: We can squash, "the whole thing", even if there are merges in its history?

braver:
Yes, squashing will reduce everything down to one commit with the actual changes.

pykong:
OK, I'll try that.

braver:
Well, you don't really have to do anything about it right now. The important thing for a PR is that people can go to the "Files changed" tab and see just your changes and not a whole history of merges. The squash thing is really a bit of a "rescue" to prevent merging a lot of merge commits into the master, not something you want to do a lot of.

pykong:
Ah, this makes sense.

lint/linter.py
pykong: #899

sublime_linter.py
pykong: #900

sublime_linter.py
pykong: We need to separate error types here. Do you see a cleaner solution than iterating over `("warning", "error")`?

sublime_linter.py
```diff
@@ -332,22 +316,25 @@ def join_msgs(cls, line_dict, we_count, show_count=False):
                 <div>{messages}</div>
             '''
 
-        template = "{linter}: {code} - {escaped_msg}"
-        template_no_code = "{linter}: {escaped_msg}"
+        tmpl_with_code = "{linter}: {code} - {escaped_msg}"
+        tmpl_sans_code = "{linter}: {escaped_msg}"
```
pykong: I also did not find a better way to conditionally include ` - ` based upon noneness of `code`.
AFAIK the string formatting mini lang does not support conditions.

FichteFoll: None of Python's default templating engines support this functionality. A more advanced template engine like `jinja2` would be necessary for this, but this just seems overkill here.

FichteFoll: Alternatively, `sublime.expand_variables` could be abused with replacements on the `code` variable for the `" - "` segment, though that would make things less clear imo.

(i.e. `${code}${code/.+/ - }`, assuming an empty `code` is an empty string)

braver: The templates are only used in one place, and now they’re clearly named. I wouldn’t put more thought into it at this point.

sublime_linter.py
braver: I think the html.escape should be done just-in-time. It’s not just sanitation of the data, it’s a transformation of the string specifically for this situation. It’s 👍🏻 as is IMO.

sublime_linter.py
pykong: Thanks for the input guys. We'll just leave it the way it is.

braver:
Ok, when testing I did find bugs (#901 and #902), but they were not introduced here. So this is LGTM for me.

PR #903, opened 2018-01-29
re-parse styles on linter settings change

braver:


kaste:
But we can at least be that efficient that if styles and linters change at the same time, we only parse once. 😁

braver:
Oh jeez, this is going to be a loooong week if I stay this fuzzy. I'm off to get more ☕️

pykong:
LGTM.

Issue #897, opened 2018-01-27
Provide API major version for plugins


braver:
SL3 has no means to report its version.

SL4 has util.get_sl_version, which reports a semver version string when installed through package control and “unknown” otherwise.

Currently the version is set manually when creating a release. I don’t like hardcoding a version number and trying to keep it in sync with the version in PC. Manually managing a semver versioning scheme is already broken enough if you do it in one place. 
However, we could have a different version number internally for the api to plugins. I propose exposing an integer via sublimelinter.version, hardcoded as a constant, and we should remember to bump it when we hit the api. It would by nature be None on SL3 or manual installs of old versions, which plugins need to guard for anyway.

braver:
cc @kaste

FichteFoll:
Hardcoding a semi-constant value such as `(major, minor)` sounds like a good compromise here, imho. Minor releases aren't as frequent but could still be useful for plugins in case a new feature was introduced.
That will also work for `master` checkouts.

A CI test for tag branches could be added to ensure that the hardcoded version tuple doesn't go out of sync.

braver:
fixed in e19efc326f8d29c0c7d4ebedd27a2fef667d27e1 and 72e66ebe97fa47f7bc2aad3367808a0113776a65

PR #906, opened 2018-01-29
A version number for plugins

braver:
addresses #897 

This is not really the most excellent way to go about this, but perhaps gets the ball rolling. With this, a plugin can do

```py
from SublimeLinter.lint import PythonLinter, util

try:
    v = util.get_sl_version().split(".")
    api_version = v[0]
except Exception:
    # SL3 doesn't have this feature
    api_version = 3
```

I really don't want to introduce different versions for plugins vs. end users, that needlessly complicates matters. New releases with niceties for end users: bump the patch number. New feature in the api: bump the minor number. Removal of features from the api: bump the major number. We can let the hardcoded patch number at 0 as long as we remember to bump it for the api changes, which shouldn't happen that often either. 

I don't think we should have all sorts of different versions out there, it's not the goal to have multiple implementations in every linter. It's also not a goal to create a very formal contract with each plugin about the API features they can expect. End of the day it's all just a single environment full of python code anyway. We're just looking for a, mostly temporary, hook to not have flake8 break on everybody right now, and to enable api breakage down the road.

This PR exposes the version we want to communicate with end users. When deployed via package control `get_sl_version` returns the exact same semver string as we tagged the repo with. The hardcoded version is a fallback for development usage. Because older version of SL cannot report their version plugins cannot be sure that `get_sl_version` exists, but can reasonably assume that if it doesn't it's probably SL3.

Opinions very welcome @FichteFoll @kaste @pykong

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -454,8 +454,10 @@ def load_json(*segments, from_sl_dir=False):
 
 
 def get_sl_version():
+    import SublimeLinter
+    version = str(getattr(SublimeLinter, 'VERSION'))
     try:
         metadata = load_json("package-metadata.json", from_sl_dir=True)
-        return metadata.get("version")
+        return version + ", " + metadata.get("version")
     except Exception:
-        return "unknown"
+        return version + ", dev"
--- a/lint/util.py
+++ b/lint/util.py
@@ -11,7 +11,7 @@ import tempfile
 
 from copy import deepcopy
 
-from .const import WARNING, ERROR
+from .const import WARNING, ERROR, VERSION
 
 STREAM_STDOUT = 1
 STREAM_STDERR = 2
@@ -458,4 +458,4 @@ def get_sl_version():
         metadata = load_json("package-metadata.json", from_sl_dir=True)
         return metadata.get("version")
     except Exception:
-        return "unknown"
+        return VERSION
--- a/lint/util.py
+++ b/lint/util.py
@@ -11,7 +11,7 @@ import tempfile
 
 from copy import deepcopy
 
-from .const import WARNING, ERROR, VERSION
+from .const import WARNING, ERROR
 
 STREAM_STDOUT = 1
 STREAM_STDERR = 2
@@ -458,4 +458,4 @@ def get_sl_version():
         metadata = load_json("package-metadata.json", from_sl_dir=True)
         return metadata.get("version")
     except Exception:
-        return VERSION
+        return "unknown"
--- a/__init__.py
+++ b/__init__.py
@@ -1,0 +1,2 @@
+"""Provides an API version for plugins."""
+VERSION = 1
--- a/lint/const.py
+++ b/lint/const.py
@@ -1,5 +1,3 @@
-VERSION = "4.0.0"
-
 PROTECTED_REGIONS_KEY = "sublime_linter.protected_regions"
 STATUS_KEY = "sublime_linter_status"
 STATUS_BUSY_KEY = "sublime_linter_status_busy"
--- a/lint/const.py
+++ b/lint/const.py
@@ -1,3 +1,5 @@
+VERSION = "4.0.0"
+
 PROTECTED_REGIONS_KEY = "sublime_linter.protected_regions"
 STATUS_KEY = "sublime_linter_status"
 STATUS_BUSY_KEY = "sublime_linter_status_busy"


pykong:
Minor point:
You may consider declaring a custom exception.
Makes it more explicit.

```py
class SuperBadVersionException(Exception):
    pass
```

FichteFoll:
Any comment on my suggestion to make it `(major, minor)`?

lint/util.py
```diff
@@ -458,4 +458,4 @@ def get_sl_version():
         metadata = load_json("package-metadata.json", from_sl_dir=True)
         return metadata.get("version")
     except Exception:
-        return "unknown"
+        return VERSION
```
FichteFoll: Since we're talking semver here, you could make this `VERSION + "-dev"`, for example.

braver:
Well, that's an option, but then do we report the tagged version to end users, or do we report the hardcoded version and build a check into CI that tells us if we tagged it wrong? In this PR, we only hardcode something for SL developers and people who want to hack on it, and so it doesn't carry any guarantees.

FichteFoll:
The idea behind `(major, minor)` is to allow plugins more fine-grained control over what to do under certain SL versions, like e.g. branching the entire module behind a version check. This can be done with the version string too, but requires more attention to detail since the version string may be a semantic version with pre-releases and metadata that needs to be stripped, potentially. Furthermore, but not including the patch number, it could be reserved for bugfixes and UI changes only. Usually you shouldn't depend on a certain patch version anyway with semver, so major and minor are sufficient. 

Checking for version compliance would then be
```
sl_version = util.version() if hasattr(util, 'version') else (3, 0)
if sl_version > (3,): # stuff in 4.x
elif sl_version >= (4,0): # same as above
elif sl_version >=(4, 2) and sl_version <= (4, 4): # do some hotfixing for versions 4.2.x to 4.4.x
```

CI checking would be entirely separate and only check on the `VERSION` field of `util.py`.

---

However, I believe that hard-coding the version is probably the way to go here, in which case that could just become a tuple of `(major, minor)` and optionally even patch, in case someone would need to go this far. The reason why I proposed using `package-metadata.json` was for debugging purposes only inside a print, but if the version is to be exposed to plugins, this is way too fragile, in which case a hard-coded fallback needs to be specified. However, if we're doing a hard-coded fallback, we might as well not care about PC's metadata at all and use that directly. This has the benefit of being reliable, static and an attribute (allows `getattr(module, field, default)` as opposed to the inline-if in my example.

Thus, I propose the following:

- Hard-code the current version inside `__init__.py` (inside root) as `VERSION = (4, 0, 0)` as an API for plugins. Plugins can use it as `getattr(SublimeLinter, 'VERSION', (3,))` for example.
- Keep `util.get_sl_version` for debugging (as it is used currently). Improve debugging output to show information about being inside a `.sublime-package`, whether it's managed by PC and what the latest git commit and active branch is, if available.
- (less important) Add a CI check for tagged releases on whether the hard-coded `VERSION` is correct (or use a script to push releases that updates it automatically)

braver:
Since only major and minor are the only versions plugins should care about, extracting them from a semver string is trivial too.

How is reading package metadata fragile? I don’t see how that can break really. It will work for everybody but the 5 people who hack on SL.

I’d still rather parse the version string ourselves and return a tuple from that. I really don’t like the idea of manually keeping 2 version numbers in sync, or explaining to plugin devs why they’re different.

lint/util.py
braver: Why would a plugin care if it's "-dev"? Is that really something you want a plugin to be able to switch on? Either their API breaks or it doesn't, beyond the minor version plugins shouldn't care.

braver: Sorry, misunderstood.

braver:
> This has the benefit of being reliable, static and an attribute (allows getattr(module, field, default)
> Hard-code the current version inside __init__.py (inside root) 

Well, ok, that does sound like a good idea. I'd like to completely decouple the numbering from the package control versioning though. All plugins should care about is the internal API version. There is no need for semantic versioning here, if the API breaks we just bump the integer. 

> Keep util.get_sl_version for debugging (as it is used currently). Improve debugging output to show information about being inside a .sublime-package, whether it's managed by PC 

"dev" indicates as much in the debug print

> and what the latest git commit and active branch is, if available.

Well that's too much. If you're developing on SL you better know about that stuff and how to report issues in a way that collaborators can work with. That's not what the debug print is for. 

> (less important) Add a CI check for tagged releases on whether the hard-coded VERSION is correct (or use a script to push releases that updates it automatically)

I'll maybe do something like incrementing the integer after each release. Not the first thing on my list.

@kaste have an opinion to share before I merge?

kaste:
That's almost too much here, really. The current version is '4'. As a farewell commit and release the SL3 gets the same function with '3'. This is the only relevant thing here IMHO, SL3 needs that function, so we don't have to hack here.

The debugging thing stays untouched here. You just print the package metadata, and maybe '4-on-master' otherwise.

So in __init__.py VERSION = 4, for SL3 a similar change. and done. After the change, this should be possible for a plugin.

```
import SublimeLinter

if SublimeLinter.VERSION == 3:
	import ...
    # the old plugin
else:
	import ....
    # the new plugin
```
The above is the actual goal here of what we want to achieve. Not fancy versioning.

braver:
> The above is the actual goal here of what we want to achieve. Not fancy versioning.

Excellent, totally agree.

braver:
see e19efc326f8d29c0c7d4ebedd27a2fef667d27e1 and 72e66ebe97fa47f7bc2aad3367808a0113776a65

FichteFoll:
All right, that's pretty much what I had in mind as well. Well, except that I think the minor version would have benefits, but we could also just bump the major version for the next breaking change in that regard and introduce it later, if we somehow find a plugin in need of that.

Issue #907, opened 2018-01-30
'Available variables' debug print is noisy


FichteFoll:
It can be useful at times, but since the print includes all my environment variables, which count 36 in total right now, each printed on a new line and not ever changing, I don't thing this is useful. We could print the environment variables once on startup and `window.extract_variables` on each lint, as that seems useful for context on the file to be linted, but is also excessive since most of the values are deterministic aside from `file`, `folder` and `project` (and `packages` potentially).

PS: Since `window.extract_variables` only accounts for the window's active view, is linting only ever run for the active view? It does seem like it, though.

FichteFoll:
I just noticed that the *entire* console is filled with debug prints for this when I save a linter plugin (which probably causes a reload of SL). Yes, everything from the top to the bottom. I didn't even know there was a limit for this.

This definitely needs to be optimized. I have 44 views open currently.
```py
>>> sum(len(list(w.views())) for w in sublime.windows())
44
```

FichteFoll:
Looking into this, the entire debug print of the variables to substitute is kind of questionable since linters are run asynchronously. You have no way to associate which prints are coming from which lint task and which linter is being run on which view since they all come in at random.

We're either looking at a logger abstraction to have logging/debug events attributed to certain linters/tasks by their associated logger instance attribute, or removing this basically useless debug statement.

kaste:
> PS: Since window.extract_variables only accounts for the window's active view, is linting only ever run for the active view? It does seem like it, though.

No that's a real bug. I never came to file this.

Otherwise, printing env variables is not only noisy as hell, but often this data contains stuff you shouldn't post here in issues. 

I wanted to change that so we only log changed settings. E.g. Changed 'executable' from '..' to '..'. Or: executable now is '...'

PR #909, opened 2018-01-30
Refactor logging

FichteFoll:
It was really hard to associate some debug prints with a certain linter (not even considering which view it was executed on) if they originated from outside of the `linter.Linter` class, since no context information could be printed.
Furthermore, added context in logging calls were quite redundant as the linter name had to be added to every call within a lint task.

This PR adds a stateful logging module that maintains a per-thread active logger cache that linters can set when they do the bulk of their work and delegate stuff to the other modules, such as process execution or settings expansion.

Example console output:
```
[SublimeLinter] [flake8] (linter.py) Selected variables:
{   'file': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter/lint/linter.py',
    'folder': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter',
    'project': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter/sublime_linter.sublime-project'}
[SublimeLinter] [flake8] (linter.py) cmd: ['/usr/bin/flake8', '--format', 'default', '--max-line-length=99', '--ignore=E201,E221,E222,E241,E722,W191,W291,W292,W293,W503,D1,D211,D413,D', '-']
[SublimeLinter] [flake8] (linter.py) cwd: /home/fichte/.config/sublime-text-3/Packages/SublimeLinter
[SublimeLinter] [flake8] (linter.py) output:
    stdin:14:100: E501 line too long (117 > 99 characters)
    stdin:439:100: E501 line too long (100 > 99 characters)
```

Partially addresses #907. Settings are still expanded way too frequently.

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -124,10 +124,7 @@ def get_linters(view):
 
                 for pattern in excludes:
                     if fnmatch(filename, pattern):
-                        persist.debug(
-                            '{} skipped \'{}\', excluded by \'{}\''
-                            .format(linter.name, filename, pattern)
-                        )
+                        linter.logger.debug('skipping; excluded by \'{}\'', pattern)
                         matched = True
                         break
 
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -124,7 +124,7 @@ def get_linters(view):
 
                 for pattern in excludes:
                     if fnmatch(filename, pattern):
-                        linter.logger.debug('skipping; excluded by \'{}\'', pattern)
+                        linter.logger.debug('skipping; excluded by %r', pattern)
                         matched = True
                         break
 
--- a/lint/base_linter/python_linter.py
+++ b/lint/base_linter/python_linter.py
@@ -68,7 +68,7 @@ class PythonLinter(linter.Linter):
         # point to a python environment, NOT a python binary.
         python = settings.get('python', None)
 
-        self.logger.debug("wanted python is '{}'", python)
+        self.logger.debug("wanted python is '%s'", python)
 
         cmd_name = cmd[0] if isinstance(cmd, (list, tuple)) else cmd
 
@@ -78,9 +78,9 @@ class PythonLinter(linter.Linter):
                     python, cmd_name
                 )
                 if not executable:
-                    self.logger.print(
-                        "WARNING: deactivated; cannot locate '{}' "
-                        "for given python '{}'",
+                    self.logger.warning(
+                        "WARNING: deactivated; cannot locate %r "
+                        "for given python %r",
                         cmd_name, python
                     )
                     # Do not fallback, user specified something we didn't find
@@ -94,14 +94,14 @@ class PythonLinter(linter.Linter):
                 )
 
                 if executable is None:
-                    self.logger.print(
-                        "WARNING: {} deactivated; cannot locate '{}' "
-                        "for given python '{}'",
+                    self.logger.warning(
+                        "WARNING: deactivated; cannot locate %r "
+                        "for given python %r",
                         cmd_name, python
                     )
                     return True, None
 
-                self.logger.debug("Using '{}' for given python '{}'", executable, python)
+                self.logger.debug("Using '%s' for given python %r", executable, python)
                 return True, executable
 
         # If we're here the user didn't specify anything. This is the default
@@ -109,18 +109,18 @@ class PythonLinter(linter.Linter):
         cwd = self.get_working_dir(settings)
         executable = ask_pipenv(cmd[0], cwd)
         if executable:
-            self.logger.debug("Using '{}' according to pipenv", executable)
+            self.logger.debug("Using '%s' according to pipenv", executable)
             return True, executable
 
         # Should we try a `pyenv which` as well? Problem: I don't have it,
         # it's MacOS only.
 
-        self.logger.debug("trying to use globally installed {}", cmd_name)
+        self.logger.debug("trying to use globally installed %r", cmd_name)
         # fallback, similiar to a which(cmd)
         executable = util.which(cmd_name)
         if executable is None:
-            self.logger.print("WARNING: cannot locate excecutable."
-                              " Fill in the 'python' or 'executable' setting.")
+            self.logger.warning("WARNING: cannot locate excecutable."
+                                " Fill in the 'python' or 'executable' setting.")
         return True, executable
 
 
@@ -159,7 +159,7 @@ def find_script_by_python_env(python_env_path, script):
     else:
         full_path = os.path.join(python_env_path, 'Scripts', script + '.exe')
 
-    logging.print("trying {}".format(full_path))
+    logging.info("trying '%s'".format(full_path))
     if os.path.exists(full_path):
         return full_path
 
@@ -222,7 +222,7 @@ def _communicate(cmd, cwd):
             cmd, env=env, startupinfo=info, universal_newlines=True, cwd=cwd
         )
     except Exception as err:
-        logging.debug("executing {} failed. reason: {!s}", str(err))
+        logging.debug("executing %s failed. reason: %s", cmd, err)
         return ''
 
 
@@ -239,9 +239,9 @@ def get_python_version(path):
         # 'python -V' returns 'Python <version>', extract the version number
         return extract_major_minor_version(output.split(' ')[1])
     except Exception as ex:
-        logging.print(
-            'ERROR: an error occurred retrieving the version for {}:\n{}',
-            path, str(ex)
+        logging.error(
+            'ERROR: an error occurred retrieving the version for %s:\n%s',
+            path, ex
         )
 
         return {'major': None, 'minor': None}
--- a/lint/base_linter/python_linter.py
+++ b/lint/base_linter/python_linter.py
@@ -6,7 +6,7 @@ import re
 import subprocess
 
 import sublime
-from .. import linter, persist, util
+from .. import linter, util, logging
 
 
 class PythonLinter(linter.Linter):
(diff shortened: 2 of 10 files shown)


kaste:
If we do a drastic change here, it should be based on pythons logging module. It is thread safe, and configurable. 

Notes: For the LinterLogger, we can't assume the filename is static. (You assign here everything on *self*, but don't use it, do you?)

Why do we have both: A context manager *and* specific Logger instances on *cls* as well as on *self*?

FichteFoll:
> it should be based on pythons logging module IMO.

I tried to have the API consistent and similar to the logging module so that a potential switch could be performed later. I intended to introduce `warning` and `error` variants to, for which we currently use `print` directly and specify the level in the message, but I wanted to get the initial refactor out first.

The problem with the standard logging module is that it is inherently hierarchic (based on a logger's name, where a dot introduces a new level) (not so much of a problem) and that conveying the extra information about the linter's name and the affected file cannot be reasonably stored within Linter (or even a subclass). It *is* possible, but it would affect global state and that's just not something I'm willing to do in a shared environment like ST's. It also starts getting quite hacky rather soon.
You also cannot use `{}` formatting with the default logger in our environment, but that could be worked around by just using `%s`.

I *could*, however, use the logging module within our custom Logger classes for thread-safe printing instead of using `print` directly.

> For the LinterLogger, we can't assume the filename is static.

Fair point, I shall request the file name on each invocation instead of caching it.

>Why do we have both: A context manager and specific Logger instances on cls as well as on self?

The context manager is needed for logging events that do not originate directly from methods within the Linter class, i.e. `util`. I could pass the logger around as an argument to all functions that would emit debug prints, but I considered this solution to be less intrusive.

Logger instances for self should be self-explanatory (they have context on the linted view, i.e. the file name).

Logger instances for the class are needed for meta-operations without an associated view, i.e. when evaluating a linter plugin inside the meta class or calling any of its classmethods.

kaste:
I don't think so, @FichteFoll. As soon as you're 'in' a logging context `self.logger.print` and `logging.print` should be identical.

FichteFoll:
> As soon as you're 'in' a logging context self.logger.print and logging.print should be identical.

That's true, but I prefer to address the logger explicitly where I can rather than sticking context managers everywhere (and increasing indentation level).

>I don't think so

I'm actually not sure what exactly this refers to.

kaste:
Hm, everywhere you're currently using self.logger you're actually 'in' a context.

FichteFoll:
I found a decent way to utiize the standard `logging` library under the hood. Currently rewriting everything for that. Will continue tomorrow, though.

Unfortunately, I still get some interleaving for individual log events, but it feels much better to rely on the stdlib module, even though I am doing things a little unconventionally.

```
[SublimeLinter] [flake8] (util.py) Selected variables:
{   'file': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter/lint/util.py',
    'folder': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter',
    'project': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter/sublime_linter.sublime-project'}
[SublimeLinter] [annotations] (util.py) Selected variables:
{   'file': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter/lint/util.py',
    'folder': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter',
    'project': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter/sublime_linter.sublime-project'}[SublimeLinter] [flake8] (util.py) wanted python is 'None'

[SublimeLinter] [annotations] (util.py) Selected variables:
{   'file': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter/lint/util.py',
    'folder': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter',
    'project': '/home/fichte/.config/sublime-text-3/Packages/SublimeLinter/sublime_linter.sublime-project'}
```

kaste:
Again, the sheer amount of work you did here I appreciate. I think we can have a simpler solution though, but I'm not completely sure here. 

Generally, I think we first need a consensus about what we want to achieve here. What the main goals are.

I'm actually committed to make #887 work. And I thought you were doing the syntax_map stuff. That's just me probably, I need this workflow where I first see issues/things a team wants to do, then someone picks it, *thinks* about a solution, and then code. Really, that's me, *I* want to know which issues/problems a PR tries to solve, so when I do the review I think about these goals and actually re-read them, and see if the solution and the 'issue' match.

braver:
> What the main goals are.

Perhaps I can provide some focus here. The debug logging exists because users tend to run into the same kinds of problems repeatedly, but don’t always describe them in ways that make it clear what is going on. People have problems with the env vars not being loaded correctly, they tell us “it’s broken dunno why”, and we add a debug dump that lets us see that it’s broken because env vars.

It should be trivial to add logging that reveals something about the state of SL, to help identify specific categories of problems. 

We don’t need full on debugging via the logger for development purposes, although I can see the value in being able to hook up something temporarily. 

The logs shouldn’t contain noise we don’t need because that could hide information we do need.

I think we can scrap some of the logging we do now. We can always add some back if it helps hunting down common issues. I don’t think we need fancy logging, prints should do fine. Unless the logging doesn’t report truth reliably, then we might need something more fancy. I personally haven’t experienced anything lacking in our existing logging, except that it could be reduced a bit.

FichteFoll:
I think I should provide some context here. So, my main goal was to fix #907, which was *excessive* printing on the console in debug mode. So much in fact that it slowed down the entire lint process and made the console basically unusable. In order to attempt a fix, I first started adding some context information to the log events, so I could understand where they are coming from (that's why this PR exists). While doing this I realized that *many* log events have to supply the linter's name to the message manually, which seemed redundant and which the PR makes unnecessary. The rest of the work was done to utilize the stdlib logging module.

I wanted to show a video recording of #907 with a console log for inspection, but for some reason I am unable to reproduce it even with the branch I initially used before starting to hack on it. Let me just put it this way: The console was practically unusable.

FichteFoll:
Ah, I understand now. The reason why I was getting so many prints was because I removed the lru cache in https://github.com/SublimeLinter/SublimeLinter/pull/910/commits/9b0679dd421b3fe9791f570117f99be6885d6a99, which didn't really do what it was supposed to as it was vulnerable to races invalidating the cache (the cache size was 1, but a different view would have different variables).

I'll post the details in #910.

braver:
I don't think this is worth pursuing further. Correct me if I'm wrong, because then we should re-state the goals of this PR.

FichteFoll:
I thing I laid out my goals sufficiently, yet you never commented on them.

From op:

>All logging calls were quite redundant, especialls those including the linter name. Furthermore, it was really hard to associate some debug prints with a certain linter (not even considering which view it was executed on) if they originated from outside of the linter.Linter class, since no context information could be printed.

From my second to last comment:

>In order to attempt a fix, I first started adding some context information to the log events, so I could understand where they are coming from (that's why this PR exists). While doing this I realized that many log events have to supply the linter's name to the message manually, which seemed redundant and which the PR makes unnecessary.

I can understand if you want to reject the pr if you think it does too much work for something that should be simple (logging?), but the fact that the backend is threaded plays a significant role here. 

Either way, closing because you don't see the goals even though I provided my goal and the motivation behind it without even asking questions about it if you had misunderstandings just feels like a slap in the face. Maybe I miscommunicated, but I can't see that just on my own.

braver:
Sorry, I probably just didn't parse the goals then.

> All logging calls were quite redundant, especialls those including the linter name.

I think this calls for reduction in debug prints then. The debug mode has a specific purpose, we only need to have debug prints for things that really fit this purpose and we can remove anything we think we don't need. 

> Furthermore, it was really hard to associate some debug prints with a certain linter (not even considering which view it was executed on) if they originated from outside of the linter.Linter class, since no context information could be printed.

Ok, I understand how that could be improved. However, I think during debugging it's best to only run one linter and circumvent this issue. 

> In order to attempt a fix, I first started adding some context information to the log events, so I could understand where they are coming from (that's why this PR exists).

This I get too, sounds like a good idea. 

> which seemed redundant and which the PR makes unnecessary.

So, is the PR necessary now or not? I don't follow. I've read the thread a couple of times and there is a lot of chatter like this that make me really confused about the problem, why it needs this particular solution, and if it does actually still need this solution. 

I also really think a few prints here and there is all we should have. If you say we really need the logging  from this PR because crucial context information is missing from the prints, I'll happily merge this.

FichteFoll:
>> All logging calls were quite redundant, especialls those including the linter name.
>
> I think this calls for reduction in debug prints then.

In this instance, I was only referring to the linter's name having to be added to every debug print call, for instance as follows:
```diff
-                util.printf(
-                    'WARNING: {} deactivated, cannot locate the gem \'{}\''
-                    .format(self.name, gem)
-                )
+                self.logger.warning('WARNING: deactivated; cannot locate the gem %r', gem)
```

The linter name is now always displayed in the prefix after `[SublimeLinter]` (as `[{linter.name}]`).

>Ok, I understand how that could be improved. However, I think during debugging it's best to only run one linter and circumvent this issue.

There might be issues where a single linter and a single file do not reproduce the issue, such as most likley what I was experiencing in #907 and the high number of lint_tasks as shown in #910, because that seems to involve either some race or a different bug. I won't investigate into that in the next few days at least, though, as I have no time to spare.
Dismissing an improvment for debugging under the premise of "ideally this should never be needed" can bite you real bad when it turns out the ideal couldn't be reached.

By the way, I'm sure one or the other user would like to check the actual commands run and the resutling outputs of the linters and it would help immensely if they had a way to associate a `{linter.name} output:` log with a certain file.

>>which seemed redundant and which the PR makes unnecessary.
>
>So, is the PR necessary now or not?

This was again referring to my first point here (the linter name). I was just repeating myself with different phrasing. The PR makes manual specification of `linter.name` in the log message unnecessary and removes it.

Happy to address any further questions.

FichteFoll:
Translating #930 into actions for this PR, I arrive at the following checklist:

- [ ] Remove `self.logger.*` and replace with `logging.*
- [ ] Set active logger in `backend.execute_lint_task` and include line number for the lint in message, if not for entire view

The rest remains to be decided. Will edit post.

kaste:
Superseded by #980

PR #913, opened 2018-01-31
Fixes #912 where a file extension is required for some linters

### LINKED ISSUE #912: Some linters require a file extension to run properly
I was editing some PHP files the other day and then noticed that the _"SublimeLinter-phpcs"_ was not reporting any errors nor warnings over a file that clearly contained multiple coding style errors suggested by PSR2. After some time debugging the code I found the reason:

* SublimeText is completely off _(Cmd + Q on macOS)_
* Open SublimeText and then load a PHP file with PSR2 errors
* `SublimeLinter: detected syntax: plain text` is printed to the console
* Use the command prompt to change the syntax to PHP
* `SublimeLinter: detected syntax: html` is printed to the console
* Use the command prompt to change the syntax to PHP _(again)_
* `SublimeLinter: detected syntax: plain text` is printed to the console

This really doesn't matter much considering that the [PHP CodeSniffer Linter](https://github.com/SublimeLinter/SublimeLinter-phpcs/blob/a99189a/linter.py#L19) supports PHP, HTML and HTML5 syntax, however, after some thinking I realized what the problem was. PHP CodeSniffer — itself — doesn't runs on non-PHP files:

### Sample File

```
<?php

if
	(TRUE)
{
	echo "hello";
}

else

{echo "world";}
```

### PHP CodeSniffer Against A File With PHP Extension

```
$ phpcs --report=checkstyle --standard=PSR2 untitled.php 
  <?xml version="1.0" encoding="UTF-8"?>
  <checkstyle version="3.2.2">
  <file name="/Users/cixtor/Desktop/untitled.php">
    <error line="3" column="1" severity="error" message="[…]"/>
    <error line="4" column="1" severity="error" message="[…]"/>
    <error line="4" column="6" severity="error" message="[…]"/>
    <error line="4" column="10" severity="error" message="[…]"/>
    <error line="6" column="1" severity="error" message="[…]"/>
    <error line="7" column="1" severity="error" message="[…]"/>
    <error line="9" column="1" severity="error" message="[…]"/>
    <error line="11" column="1" severity="error" message="[…]"/>
    <error line="11" column="15" severity="error" message="[…]"/>
    <error line="11" column="15" severity="error" message="[…]"/>
  </file>
  </checkstyle>
```

### PHP CodeSniffer Against A File Without PHP Extension

```
$ phpcs --report=checkstyle --standard=PSR2 untitled
  <?xml version="1.0" encoding="UTF-8"?>
  <checkstyle version="3.2.2">
  </checkstyle>
```

When a linter runs, SublimeLinter creates a temporary file with the content of the file that is currently being edited. On macOS the files are usually stored under `$TMPDIR` and the names follow this pattern `tmp[a-z]{6}`. When the SublimeLinter-phpcs passes the command through the `cmd` variable, it ends up [here](https://github.com/SublimeLinter/SublimeLinter/blob/dde0412/lint/util.py#L299-L306) with the following:

```
# cmd = [
#     '/usr/local/bin/phpcs',
#     '--report=checkstyle',
#     '--standard=PSR2',
#     '/var/folders/gy/scj0933963ggbfdxcn5c91vw0000gn/T/tmpbilqso'
# ]
out = popen(cmd, stdout=stdout, stderr=stderr,
    output_stream=output_stream, env=env, cwd=cwd)
```

Because of this, SublimeLinter is unable to obtain the correct output and so SublimeLinter-phpcs cannot tag the errors.

### Solution

When SublimeLinter detects the file syntax, this information must be passed down to `util.communicate()`. Once the content of the original file is copied to the temporary file, rename the later to include the file extension, then SublimeLinter can execute the command via `process.Popen`.

cixtor:
I found that some linters like [SublimeLinter-phpcs](https://github.com/SublimeLinter/SublimeLinter-phpcs) stopped working after a recent update. After some debugging of the code I found that most recent versions of PHP CodeSniffer do not work when the file has a different extension than `.php` or has no extension at all.

Before SublimeLinter executes the command specified by the variable `cmd` it creates a temporary file, which on macOS is placed at `$TMPDIR` and then copies the content of the original file there, then it executes the command using `process.Popen`; the problem here is that the files are being created using `tempfile.NamedTemporaryFile` which by default creates file names with this format `/tmp[a-z_]{6}`.

A solution is to pass the `suffix=".ext"` parameter to the method call.

Fixes #912

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1298,22 +1298,18 @@ class Linter(metaclass=LinterMeta):
 
     def get_tempfile_suffix(self):
         """Return the mapped tempfile_suffix."""
-        suffix = ''
+        if self.tempfile_suffix and not self.view.file_name():
+            if isinstance(self.tempfile_suffix, dict):
+                suffix = self.tempfile_suffix.get(util.get_syntax(self.view), self.syntax)
+            else:
+                suffix = self.tempfile_suffix
 
-        if isinstance(self.tempfile_suffix, str):
-            suffix = self.tempfile_suffix
-        elif isinstance(self.tempfile_suffix, dict):
-            suffix = self.tempfile_suffix.get(util.get_syntax(self.view), self.syntax)
+            if not suffix.startswith('.'):
+                suffix = '.' + suffix
 
-        # default to plain text
-        if not suffix:
-            suffix = '.txt'
-
-        # prepend period if necessary
-        if not suffix.startswith('.'):
-            suffix = '.' + suffix
-
-        return suffix
+            return suffix
+        else:
+            return ''
 
     # popen wrappers
 
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1298,18 +1298,22 @@ class Linter(metaclass=LinterMeta):
 
     def get_tempfile_suffix(self):
         """Return the mapped tempfile_suffix."""
-        if self.tempfile_suffix and not self.view.file_name():
-            if isinstance(self.tempfile_suffix, dict):
-                suffix = self.tempfile_suffix.get(util.get_syntax(self.view), self.syntax)
-            else:
-                suffix = self.tempfile_suffix
+        suffix = ''
 
-            if not suffix.startswith('.'):
-                suffix = '.' + suffix
+        if isinstance(self.tempfile_suffix, str):
+            suffix = self.tempfile_suffix
+        elif isinstance(self.tempfile_suffix, dict):
+            suffix = self.tempfile_suffix.get(util.get_syntax(self.view), self.syntax)
 
-            return suffix
-        else:
-            return ''
+        # default to plain text
+        if not suffix:
+            suffix = '.txt'
+
+        # prepend period if necessary
+        if not suffix.startswith('.'):
+            suffix = '.' + suffix
+
+        return suffix
 
     # popen wrappers
 
--- a/lint/util.py
+++ b/lint/util.py
@@ -328,14 +328,6 @@ def tmpfile(cmd, code, filename, suffix='', output_stream=STREAM_STDOUT, env=Non
     The result is a string combination of stdout and stderr.
     If env is None, the result of create_environment is used.
     """
-    if not filename:
-        filename = "untitled"
-    else:
-        filename = os.path.basename(filename)
-
-    if suffix:
-        filename = os.path.splitext(filename)[0] + suffix
-
     file = tempfile.NamedTemporaryFile(delete=False, suffix=suffix)
     path = file.name
 
--- a/lint/util.py
+++ b/lint/util.py
@@ -336,7 +336,12 @@ def tmpfile(cmd, code, filename, suffix='', output_stream=STREAM_STDOUT, env=Non
     if suffix:
         filename = os.path.splitext(filename)[0] + suffix
 
-    file = tempfile.NamedTemporaryFile(delete=False)
+    if not suffix:
+        # Some linters require a file extension to run properly.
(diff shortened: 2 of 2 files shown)


FichteFoll:
I would rather fix `lint.Linter.get_tempfile_suffix` to not return an empty string when a view has a file name. That seems weird in the first place.

lint/linter.py
```diff
@@ -1298,18 +1298,22 @@ def run(self, cmd, code):
...
-                suffix = '.' + suffix
+        if isinstance(self.tempfile_suffix, str):
+            suffix = self.tempfile_suffix
+        elif isinstance(self.tempfile_suffix, dict):
+            suffix = self.tempfile_suffix.get(util.get_syntax(self.view), self.syntax)
 
-            return suffix
-        else:
-            return ''
+        # default to plain text
```
FichteFoll: I don't think we *need* an extension. Just leave it empty if none was specified. 
Alternatively, you could try inferring it from the view's file name, if that exists.

cixtor: Is this the logic that you are suggesting for that method?

1. Extract the extension from `self.view.file_name()` and return if available
2. Infer the extension from the currently selected syntax and return if possible
3. Use `self.tempfile_suffix` in any other cases
4. Never return empty

Please confirm so I can commit these changes.

DaveStein:
@cixtor My files have a `.php` extension so not sure it relates to your fix

lint/util.py
```diff
@@ -336,7 +336,7 @@ def tmpfile(cmd, code, filename, suffix='', output_stream=STREAM_STDOUT, env=Non
     if suffix:
         filename = os.path.splitext(filename)[0] + suffix
 
-    file = tempfile.NamedTemporaryFile(delete=False)
+    file = tempfile.NamedTemporaryFile(delete=False, suffix=suffix)
```
kaste: Bc this is a regression from https://github.com/SublimeLinter/SublimeLinter/commit/a3db729367c0ceb423a3f617ab3fa8ab4b739409

(we just forgot about the suffix completely) This line alone should restore the old SublimeLinter behavior. So if you do just that and it works again. I could merge immediately bc I don't have to think through this whole thing. 

I don't remember we touched `get_tempfile_suffix` recently (git blame?). So I don't want to start a change here for what appears to be hotfixable. WDYT?

Of course the initial lines about `filename`in that function are no-ops. `filename` is not used anymore. Say, we had a bad day here.

braver: Damn, who tests these things. And why does git `blame` me for these lines? 😒

The hotfix (and removal of those lines about "filename") seems like the way to go to me. Can someone confirm the fix for the php thing?

cixtor:
@DaveStein my PHP files also have a `.php` extension.

If you read again my description of issue #913 you will see that the problem is not the file itself but the temporary file that SublimeLinter is creating at `$TMPDIR` which is the file is that actually passed to `phpcs` for linting, because PHP CodeSniffer excepts the file to have a `.php` extension to work, it fails and returns no warnings or errors.

As @kaste said, this pull-request will _(probably)_ fix both GitHub issues as they are the same thing.

@kaste — @braver  I have updated the pull-request with the deletion of the unused `filename` variable and associated logic. And added the missing `suffix=suffix` to the `tempfile.NamedTemporaryFile()` method call. Let me know if there is anything else missing.

braver:
Yes, this should fix #913 and #912. I assume this means it fixed your #912, so I don't have to set up a phpcs test environment for myself. LGTM, so I'll merge and push a release so people get to enjoy php code sniffing again.

PR #917, opened 2018-01-31
Modify Linter.get_tempfile_suffix to extract file extension

cixtor:
Right now, the `Linter.get_tempfile_suffix` method is returning an empty string as a fallback mechanism when the extension for the temporary file is not found. This commit adds an attempt to extract the extension from the filename (if available) and defaults to an empty string if the file has no name.

Notice that this doesn't breaks the current functionality, it merely improves the current code by adding an additional step to try to obtain the information that the method is already trying to get from the View. This change also complements the hotfix applied [here](https://github.com/SublimeLinter/SublimeLinter/pull/913) where a file extension in the temporary file is required by some linters.


kaste:
LGTM! 👌 

@braver This is actually a hotfix addendum to restore the previous behavior.

braver:
🎉

PR #922, opened 2018-02-01
Do not print substitution variables

### LINKED ISSUE #907: 'Available variables' debug print is noisy

kaste:
Fixes #907

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -494,15 +494,7 @@ class Linter(metaclass=LinterMeta):
         if project_folder:
             variables['folder'] = project_folder
 
-        if persist.debug_mode():
-            import pprint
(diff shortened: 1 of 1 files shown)


braver:
That’s a good solution IMO, LGTM and 🚢

PR #926, opened 2018-02-01
fix line report

braver:
Calling the line report command only displays the error at your cursor. This is incorrect behaviour. This PR changes that:

- show line report when cursor not at error
- don't filter errors for cursor position when requesting line report
- line report really is just the gutter report but not at the gutter

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -39,4 +39,4 @@ class SublimeLinterLintCommand(sublime_plugin.TextCommand):
 class SublimeLinterLineReportCommand(sublime_plugin.WindowCommand):
     def run(self):
         from .sublime_linter import SublimeLinter
-        SublimeLinter.shared_plugin().open_tooltip()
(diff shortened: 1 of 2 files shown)


braver:
Thanks, that's a good point. There is actually no need to set the col to 0 if someone hovers over the gutter, and if the line report is called manually it makes more sense to show it at the cursor. In fact, it wouldn't show if you've scrolled horizontally. So this fixes that too by being less smart.

kaste:
LGTM 🚢

FichteFoll:
Uh, mind explaining again why the column shouldn't be zero for the on_hover gutter popup?

braver:
It already is zero and doesn’t have to be set like that. As far as my tests reveal anyway.

FichteFoll:
Ah okay, I see the point kwarg now. Can't check the full diff on mobile.

PR #934, opened 2018-02-02
remove bid from errors when closing its last view

### LINKED ISSUE #900: Method for removing unused errors.
After merging #898, errors are based solely on `buffer_id`. This necessitates a new way to remove unused errors.

Quoting @kaste:
> We need to add some logic to remove bid from errors when it's not used anymore.

> This can be tricky. Generally we can multiple views into the same buffer. So we can remove only if this view here is the last one. But one has to check/test how we can iterate here over window.views() for example. Is this view here still in window.views(). What if the user closes the window?

braver:
`sublime.windows()` and `window.views()` always return lists, we can iterate over them to find all buffer_ids. If the view to be closed is attached to exactly 1 buffer, we also remove that buffer from `persist.errors`.

This seems to work correctly, but I see some bids in `persist.errors` that I don't have views for. They seem to be created when I open e.g. the find panel. Can it be that we create entries in `persist.errors` for such buffers? I couldn't find out if we did. 

We also have the a nuclear option of wiping out `persist.errors` completely if no views remain...

----

fixes #900

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -147,6 +147,16 @@ class Listener:
 
         queue.cleanup(vid)
 
+        # Remove bid from persis.errors if it's the last view on the buffer
+        bid = view.buffer_id()
+        buffers = []
+        for w in sublime.windows():
+            for v in w.views():
+                buffers.append(v.buffer_id())
+
+        if buffers.count(bid) is 1:
+            del persist.errors[bid]
+
     def on_hover(self, view, point, hover_zone):
         """On mouse hover event hook.
 
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -147,7 +147,7 @@ class Listener:
 
         queue.cleanup(vid)
 
-        # Remove bid from persis.errors if it's the last view on the buffer
+        # Remove bid from persist.errors if it's the last view on the buffer
         bid = view.buffer_id()
         buffers = []
         for w in sublime.windows():
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -147,15 +147,20 @@ class Listener:
 
         queue.cleanup(vid)
 
-        # Remove bid from persist.errors if it's the last view on the buffer
         bid = view.buffer_id()
         buffers = []
+        view_count = 0
         for w in sublime.windows():
             for v in w.views():
+                view_count = view_count + 1
                 buffers.append(v.buffer_id())
 
-        if buffers.count(bid) is 1:
-            del persist.errors[bid]
+        if view_count <= 1:
+            # Wipe clean after last view closes
+            persist.errors.clear()
+        elif buffers.count(bid) <= 1:
+            # Remove bid from persist.errors if it's the last view on the buffer
+            persist.errors.pop(bid, None)
 
     def on_hover(self, view, point, hover_zone):
         """On mouse hover event hook.


kaste:
Just to be clear here: in the *on_pre_close* hook, 

- [ ] we call sublime.windows() and the about to be closed *window* is still there.
- [ ] we call window.views() and the about to closed *view* is still there

sublime_linter.py
```diff
@@ -146,6 +146,16 @@ def on_pre_close(self, view):
...
         queue.cleanup(vid)
 
+        # Remove bid from persist.errors if it's the last view on the buffer
+        bid = view.buffer_id()
+        buffers = []
+        for w in sublime.windows():
+            for v in w.views():
+                buffers.append(v.buffer_id())
+
+        if buffers.count(bid) is 1:
```
kaste: just make it `<= 1` bc shit happens

kaste:
> but I see some bids in persist.errors that I don't have views for

That's a convenience for *consumers* of `persist.errors`. persist.errors is a defaultdict(list). So a consumer (a view) can just ask `persist.errors[bid]` and this will never throw but return an empty list. (Otherwise, a bit more correct maybe, you would always have to do `persist.errors.get(bid, [])`.) Well, I'm just a nice guy, ... or stupid.

sublime_linter.py
braver: Well no, it can also be 0 for some reason (shit happens) and we don't want that because `del persist.errors[bid]` will fail.

kaste: And what shit happened?

kaste: Since this is the only place where we call del[bid] it means we call it twice, the first time when buffers.count is 1 and then a second time when buffers.count is 0

braver:
> we call sublime.windows() and the about to be closed window is still there
we call window.views() and the about to closed view is still there

Yes in both cases. Although one should never trust the [docs](https://www.sublimetext.com/docs/3/api_reference.html#sublime_plugin.ViewEventListener), it says so here:

> Called when a view is about to be closed. The view will still be in the window at this point.

It's a synchronous event handler and it will stall closure of the view and it's containing window until we're done cleaning up. We better be fast, but this shouldn't take long.

> Well, I'm just a nice guy, ... or stupid.

Good, can't blame you for either. Let's clean up the `bids` after the last view closes and be done with it.

sublime_linter.py
braver: Good question. Will have a look.

kaste: Okay it probably comes from views we never linted before. I thought we have a precheck here. But now, if you look above, we usually use `pop(vid, None)` so that we never throw (bc shit happens)

kaste: So if you're more into defense, you probably do the `<= 1` check with a `pop(bid, None)`. And otherwise if you're really sure, you just commit what you got.

braver:
🤔

```py
on_pre_close(): 
        print(bid)
        print(buffers)
        print(buffers.count(bid))
```

prints:
```
53
[56]
0
```

... and amazingly, it prints that when I *open* a view. I need some more time with this to check its behaviour it seems.

sublime_linter.py
braver: Probably a good place to be defensive. Cleaning up is a "best effort" thing anyway considering we run in a kinda messy environment.

braver:
Aha, apparently, if you have zero views open there is actually an anonymous view ready for you to type in and that one gets closed when you open your first actual view. This anonymous view is also the reason `buffers.count(bid)` may be zero. So, we should be defensive when doing that.

braver:
Ok, should be good now.

kaste:
👍

PR #946, opened 2018-02-03
Manually compute 'file' variables

### LINKED ISSUE #925: `window.extract_variables` depends on the active view
`window.extract_variables` depends on the active view so we probably need to replace the variables around the current file/filename with custom definitions.

kaste:
bc `window.extract_variables()` depends on `window.active_view()`

Fixes #925

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -494,6 +494,20 @@ class Linter(metaclass=LinterMeta):
         if project_folder:
             variables['folder'] = project_folder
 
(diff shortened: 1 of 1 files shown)


braver:
Ok, sounds good and looks good. Want me to test something? Otherwise just 🚢 .

kaste:
Just 🚢 it. It did this carefully.

braver:
👍🏻

PR #947, opened 2018-02-03
Fixup working dir

### LINKED ISSUE #927: Default working_dir should be equivalent to `${folder:$file_path}`
We previously talked about how the default for `working_dir` should be `${folder:$file_path}`, but in #918 I was made aware that the actual default is just `${folder}` (compare `Linter.get_working_dir`).

https://github.com/SublimeLinter/SublimeLinter/blob/329b5d78ca6d6146cd475a7ead307a265d861481/lint/linter.py#L885-L899

To add this, we either have the option to have settings expansion do the work for us, which would be easy on the surface but require a way to specify a setting default *before* expansion, which currently does not exist, or just enhance `get_working_dir` to check for `self.view.file_name()` and return its dirname if `self._guess_project_path` returned `None`.

kaste:
Fixes #927

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -510,7 +510,7 @@ class Linter(metaclass=LinterMeta):
 
         for folder in folders:
             # Take the first one; should we take the deepest one? The shortest?
-            if folder in filename:
+            if filename.startswith(folder + os.path.sep):
                 return folder
 
     @classmethod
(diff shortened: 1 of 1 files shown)


braver:
I don’t know how to test this actually.

kaste:
If you have project open say SublimeLinter, open a file from a completely different dir. Look that `cwd` is now dirname(filename). Before the change the foreign file would have been linted with settings/cwd from SublimeLinter.

braver:
@kaste feel free to merge this 👍

PR #948, opened 2018-02-03
Fix busy indicator

### LINKED ISSUE #935: Status bar busy indicator stops on first response from backend
We only fire 'BEGIN_LINTING' and 'FINISHED_LINTING'. But since #887 the responses fly in as soon as possible. Currently all responses fire 'FINISHED_LINTING'. 

We need three events here

'START_LINTING(bid)' -> 'LINT_RESULT(bid, linter, errors)' -> 'STOP_LINTING(bid [,errors])'

Feel free to bike-shed the naming here.

kaste:
Fixes #935 

Set initial delay to 2 seconds bc 1 second is just too annoying.

### DIFF
--- a/busy_indicator_view.py
+++ b/busy_indicator_view.py
@@ -38,7 +38,7 @@ def on_begin_linting(buffer_id):
         sublime.set_timeout_async(lambda: draw(**State), INITIAL_DELAY * 1000)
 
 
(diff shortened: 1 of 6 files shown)


braver:
Yep, all good with me.

PR #949, opened 2018-02-03
Ensure highlights on whitespace only regions

### LINKED ISSUE #940: Implement workaround for whitespace 'underline' highlights
Proposal:

If the user chooses one of the underline highlights (which I personally would like to do), we check if the actual text the error is about *only* contains white-space. In such a case, due to a bug in Sublime, the highights would not show up at all. So we switch to 'outline'/'fill' style for these regions.

kaste:
Sublime cannot draw 'underlines' on spaces, which makes choosing such
a mark style not a good choice for the user. (Bc she misses some errors.)

We work around here, and just use the 'outline' style in such cases.

Fixes #940

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -1,5 +1,7 @@
 from collections import defaultdict, ChainMap
 from itertools import chain
+import re
+
 import sublime
 
 from .lint import persist, events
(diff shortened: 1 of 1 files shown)


kaste:
This is 🌶

braver:
This is super 🌶 Merge it!

kaste:
One of the better 20 lines you can write

PR #950, opened 2018-02-03
Small iteration for highlights_view

### LINKED ISSUE #939: Problems/questions for the highlights
As a by-product of refactoring highlight.py (now highlights_view.py). 

1. We have the following filtering:

https://github.com/SublimeLinter/SublimeLinter/blob/fd6f8a8e9d8db3daf634044d27b52a7a5319baec/highlight_view.py#L90-L93

which basically means we *throw away errors* if we can't find a style. (Note I don't know anything about this *style* stuff per se. I just did the refactoring.) Generally, I don't think we should throw away errors, we don't even log here. So the question is, when exactly will we throw away errors? Can't we just rely on a fallback default to be always there? If we really need to discard an error, then we should log about it. Bc it will confuse users of course.

2. We have the following sort for the gutter icons:

https://github.com/SublimeLinter/SublimeLinter/blob/fd6f8a8e9d8db3daf634044d27b52a7a5319baec/highlight_view.py#L118-L121

Shouldn't this be rather `-e['priority'], e['error_type']`? T.i if two errors have the same priority, the 'ERROR' wins. So we could demote/promote errors freely.

Also: Why isn't this code parallel to the sorting for the underline/outline highlights:

https://github.com/SublimeLinter/SublimeLinter/blob/fd6f8a8e9d8db3daf634044d27b52a7a5319baec/highlight_view.py#L134-L138

Here we just take the first 'ERROR' not looking at the priority at all.

kaste:
Fixes #939

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -115,7 +115,7 @@ def filter_errors_for_gutter(errors):
     filtered_errors = []
     for line, errors in errors_by_line.items():
         # We're lucky here that 'error' comes before 'warning'
-        head = sorted(errors, key=lambda e: (e['error_type'], -e['priority']))[0]
+        head = sorted(errors, key=lambda e: (-e['priority'], e['error_type']))[0]
         filtered_errors.append(head)
 
     return filtered_errors
@@ -132,7 +132,7 @@ def filter_errors_for_highlights(errors):
     for pos, errors in by_position.items():
         # If we have multiple 'problems' here, 'error' takes precedence over
         # 'warning'. We're lucky again that 'error' comes before 'warning'.
-        head = sorted(errors, key=lambda e: e['error_type'])[0]
+        head = sorted(errors, key=lambda e: (-e['priority'], e['error_type']))[0]
         filtered_errors.append(head)
 
     return filtered_errors
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -89,8 +89,6 @@ def prepare_data(errors):
     errors_augmented = []
     for error in errors:
         style = get_base_error_style(**error)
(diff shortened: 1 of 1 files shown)


highlight_view.py
```diff
@@ -89,16 +89,17 @@ def prepare_data(errors):
     errors_augmented = []
     for error in errors:
         style = get_base_error_style(**error)
-        if not highlight_store.has_style(style):  # really?
-            continue
```
kaste: That's important. Just remove it for now, and see if we actually get tracebacks from someone

highlight_view.py
```diff
@@ -107,37 +108,25 @@ def get_base_error_style(linter, code, error_type, **kwargs):
...
+        grouped[group_fn(error)].append(error)
 
-    # Since, there can only be one gutter mark per line, we have to select
-    # one 'winning' error from all the errors on that line
     filtered_errors = []
-    for line, errors in errors_by_line.items():
-        # We're lucky here that 'error' comes before 'warning'
-        head = sorted(errors, key=lambda e: (e['error_type'], -e['priority']))[0]
+    for errors in grouped.values():
+        head = sorted(errors, key=lambda e: (-e['priority'], e['error_type']))[0]
```
kaste: Here, we just flip priority and error_type

kaste: For me, this is kind of a bug fix, bc I actually tried to 'promote' a warning over an error, which never worked. So now I know why.

highlight_view.py
braver: Agreed

braver:
Works good and looks good 👍

braver:
@kaste wanna merge this?

PR #910, opened 2018-01-30
Only determine view's settings once per lint

FichteFoll:
We just call get_view_settings in lint and cache it as an instance variable. Too many methods were passing around or re-determining the settings during a single lint for no apparent reason, since it will not
have changed in the meantime.

Fixes #907.

PR #952, opened 2018-02-04
Compute view settings once, and expose them via a thread local ctx ob…

### LINKED ISSUE #951: Evaluate `get_view_settings` once per lint task
Computing `get_view_settings()` multiple times could result in different return values over time (Hej, it's a getter). So we should evaluate it once, to have a consistent view here. It is also probably costly to re-compute, so this would result in a performance improvement.

kaste:
…ject

Fixes #951

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -58,18 +58,20 @@ def run_tasks(tasks, next):
 
 
 def get_lint_tasks(linters, view, hit_time):
-    for (linter, regions) in get_lint_regions(linters, view):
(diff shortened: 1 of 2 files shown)


braver:
I don't know how to test or review this. Sorry. I do trust your judgement here though, and the change set doesn't look weird, so feel free to merge.

kaste:
This is basically the replacement for #910

braver:
Ok, LGTM

PR #956, opened 2018-02-04
Inline make_task

FichteFoll:
Most variables were carried over to `execute_lint_task` anyway,
so we might as well just do everything inline. This way we don't create
a new function object for every iteration.

(After 22f6bc5dd10cf26989f5c6d12286b126a6f0f3c6 and ae4648ccc31da65a5b6f8e826b1450920e5b7e22, the partial call got longer and longer, meaning the benefit from a function was pretty diminishing here anyway.)


kaste:
LGTM 👍 

GitHub did hate me, so I close this and instead merge the other PR, duh.

Issue #958, opened 2018-02-05
If settings are invalid at start, they're still loaded, leading to breakage


braver:
When you update settings they have to pass validation before being applied. If you have invalid settings and quit ST, the invalid settings are applied at startup without passing validation. This may lead to breakage, e.g. with missing mark_style:

```
Traceback (most recent call last):
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/events.py", line 28, in broadcast
    fn(**payload)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/highlight_view.py", line 78, in on_lint_result
    view, linter_name, errors_for_the_highlights)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/highlight_view.py", line 190, in prepare_highlights_data
    flags = MARK_STYLES[mark_style]
KeyError: None
```

kaste:
A typical LOL bug. Sorry.

braver:
#959 is all we're going to do about it

PR #959, opened 2018-02-05
show errors in a panel

### LINKED ISSUE #895: Open an output panel for configuration errors
We should have more visual feedback on errors in settings, or a linter executable that can't be found, etc. It looks like a good idea to open an output panel for such feedback, that also gives us space for explanation and tips on how to fix. It's also probably not that hard to do.

braver:
fixes #895, #958

Creates an output panel to have more visual feedback that settings are invalid, and also include information about what's wrong.
Removes the panel when settings are valid (i.e. removes it before checking again).
The increased feedback also helps notifying the user about broken settings during load.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -692,7 +692,9 @@ class Linter(metaclass=LinterMeta):
             path = self.which(which)
 
         if not path:
-            util.printf('WARNING: {} cannot locate \'{}\''.format(self.name, which))
+            msg = 'WARNING: {} cannot locate \'{}\''.format(self.name, which)
+            util.printf(msg)
+            util.alert(msg)
             return None
 
         cmd[0:1] = util.convert_type(path, [])
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -694,7 +694,7 @@ class Linter(metaclass=LinterMeta):
         if not path:
             msg = 'WARNING: {} cannot locate \'{}\''.format(self.name, which)
             util.printf(msg)
-            util.alert(msg)
+            util.message(msg)
             return None
 
         cmd[0:1] = util.convert_type(path, [])
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1268,10 +1268,10 @@ class Linter(metaclass=LinterMeta):
                 )
                 return True
             else:
-                util.printf(
-                    'WARNING: {} deactivated, version requirement ({}) not fulfilled by {}'
-                    .format(cls.name, cls.version_requirement, cls.executable_version)
-                )
+                warning = 'WARNING: {} deactivated, version requirement ({}) not fulfilled by {}'
+                msg = warning.format(cls.name, cls.version_requirement, cls.executable_version)
+                util.printf(msg)
+                util.message(msg)
 
         return False
 
--- a/lint/settings.py
+++ b/lint/settings.py
@@ -91,16 +91,17 @@ def validate_settings():
     status_msg = "SublimeLinter - Settings invalid!"
     schema_file = "resources/settings-schema.json"
     schema = util.load_json(schema_file, from_sl_dir=True)
-
+    window = sublime.active_window()
+    window.active_view().run_command("sublime_linter_remove_panel")
     good = True
+
     for name, settings in get_settings_objects():
         try:
             validate(settings, schema)
-        except ValidationError as ve:
+        except ValidationError as error:
             good = False
-            ve_msg = ve.message.split("\n")[0]  # reduce verbosity
-            full_msg = "Invalid settings in '{}':\n{}".format(name, ve_msg)
-            window = sublime.active_window()
+            error_msg = error.message.split("\n")[0]  # reduce verbosity
+            full_msg = "Invalid settings in '{}':\n{}".format(name, error_msg)
 
             util.printf(full_msg)
             window.status_message(status_msg)
--- a/lint/settings.py
+++ b/lint/settings.py
@@ -81,9 +81,9 @@ def get_settings_objects():
     for name in sublime.find_resources("SublimeLinter.sublime-settings"):
         try:
             yield name, util.load_json(name, from_sl_dir=False)
-        except IOError as ie:
+        except IOError:
             util.printf("Settings file not found: {}".format(name))
-        except ValueError as ve:
+        except ValueError:
             util.printf("Settings file corrupt: {}".format(name))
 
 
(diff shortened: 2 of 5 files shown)


braver:
@kaste any more cases of configuration error where we would like to use a panel instead of just `printf`?

lint/linter.py
```diff
@@ -692,7 +692,9 @@ def build_cmd(self, cmd):
             path = self.which(which)
 
         if not path:
-            util.printf('WARNING: {} cannot locate \'{}\''.format(self.name, which))
+            msg = 'WARNING: {} cannot locate \'{}\''.format(self.name, which)
+            util.printf(msg)
+            util.alert(msg)
```
FichteFoll: `util.alert` does not exist.

message_view.py
```diff
@@ -0,0 +1,20 @@
...
+        panel_view.erase(edit, sublime.Region(0, panel_view.size()))
+        panel_view.insert(edit, 0, msg)
+        panel_view.set_read_only(True)
+        panel_view.show(0)
+        sublime.active_window().run_command("show_panel", {"panel": "output.{}".format(PANEL_NAME)})
+
+
+class SublimeLinterRemovePanelCommand(sublime_plugin.TextCommand):
+    def run(self, edit):
+        sublime.active_window().destroy_output_panel(PANEL_NAME)
```
FichteFoll: Why does this need its own command? For symmetry with `sublime_linter_display_panel`? 

And why does it not use `self.view.window()` (and why is it not a `WindowCommand`).

message_view.py
```diff
@@ -0,0 +1,20 @@
+import sublime
+import sublime_plugin
+
+PANEL_NAME = "SublimeLinter Messages"
+
+
+class SublimeLinterDisplayPanelCommand(sublime_plugin.TextCommand):
```
FichteFoll: This should be a `WindowCommand`.

message_view.py
braver: For symmetry indeed. Otherwise the caller would have to know the PANEL_NAME, we'd have to share that in some way, etc. I thought this was a bit more elegant.

The creation command needs to be a TextCommand I think. I borrowed it from [GitSavvy](https://github.com/divmain/GitSavvy/blob/bee6b548b5e66bb85b3bde7f7d740f7be318c5f2/common/commands/log.py). It does an erase and insert, which I don't think you can do via a WindowCommand?

The removal command could perhaps be a WindowCommand. I just set them up in a similar way and it worked. Your proposals don't work so I must be missing something.

FichteFoll: Iirc output panels are cleared when "create_output_panel" is called and the insertion could easily be done wit the append command, but I'm.not entirely sure on the first one.

braver: Ok, sounds good. I'll look into it some more.

kaste:
The most prominent error is probably in `util.popen`

braver:
PTAL @kaste @FichteFoll

lint/settings.py
```diff
@@ -81,25 +81,29 @@ def get_settings_objects():
     for name in sublime.find_resources("SublimeLinter.sublime-settings"):
         try:
             yield name, util.load_json(name, from_sl_dir=False)
-        except IOError as ie:
+        except IOError:
             util.printf("Settings file not found: {}".format(name))
-        except ValueError as ve:
+        except ValueError:
```
FichteFoll: Could there be any useful information in this exception?

message_view.py
```diff
@@ -0,0 +1,18 @@
+import sublime
+import sublime_plugin
+
+PANEL_NAME = "SublimeLinter Messages"
+
+
+class SublimeLinterDisplayPanelCommand(sublime_plugin.WindowCommand):
+    def run(self, msg=""):
+        panel_view = sublime.active_window().create_output_panel(PANEL_NAME, True)
```
FichteFoll: `self.window`

message_view.py
```diff
@@ -0,0 +1,18 @@
...
+        panel_view.run_command('append', {'characters': msg})
+        panel_view.set_read_only(True)
+        panel_view.show(0)
+        sublime.active_window().run_command("show_panel", {"panel": "output.{}".format(PANEL_NAME)})
+
+
+class SublimeLinterRemovePanelCommand(sublime_plugin.WindowCommand):
+    def run(self):
+        sublime.active_window().destroy_output_panel(PANEL_NAME)
```
FichteFoll: `self.window`

message_view.py
```diff
@@ -0,0 +1,18 @@
...
+PANEL_NAME = "SublimeLinter Messages"
+
+
+class SublimeLinterDisplayPanelCommand(sublime_plugin.WindowCommand):
+    def run(self, msg=""):
+        panel_view = sublime.active_window().create_output_panel(PANEL_NAME, True)
+        panel_view.run_command('append', {'characters': msg})
+        panel_view.set_read_only(True)
+        panel_view.show(0)
+        sublime.active_window().run_command("show_panel", {"panel": "output.{}".format(PANEL_NAME)})
```
FichteFoll: `self.window`

lint/settings.py
```diff
@@ -81,25 +81,29 @@ def get_settings_objects():
...
 
 
 def validate_settings():
-    status_msg = "SublimeLinter - Settings invalid. Details in console."
+    status_msg = "SublimeLinter - Settings invalid!"
     schema_file = "resources/settings-schema.json"
     schema = util.load_json(schema_file, from_sl_dir=True)
-
+    window = sublime.active_window()
+    util.clear_message()
```
kaste: Is this necessary? We don't do this for the other errors, and it would be tricky to decide when a previous error has been resolved.

lint/settings.py
braver: The `IOError` is a super rare edge case. It would mean that either between looking for files at line 81 and reading one at line 83 the file has become unavailable. 

`ValueError` definitely contains valuable info, e.g. if you're missing a comma. However, if you make such an error in your settings file, SublimeText already throws you a popup explaining what's wrong. 

We should only concern ourselves with validity of user settings in terms of it containing values we understand. Being able to load and parse user settings is ST's job and it handles all of that already.

lint/settings.py
braver: What this accomplishes, is that the message disappears when you're settings are valid again. This is the only case right now where we accurately detect that a problem has disappeared. I think it's important to provide this positive feedback in this particular case.

kaste: 👍

braver:
Ok, good to go then?

lint/settings.py
FichteFoll: Good point. Didn't think of ST already raising errors for invalid settings files.

PR #960, opened 2018-02-05
Hotifx! Downsize NodeLinter

kaste:
NodeLinter used to subclass `lint` for fun. We now do these checks in `context_sensitive_executable_path`.
Since I forgot to adjust its lint method in #952, you cannot currently use any NodeLinter on master. LOL.

- Do not subclass `lint`
- Remove non-working setting `enable_if_dependency`
- Faster `context_sensitive_executable_path`


braver:
Confirmed broken on master and not broken with this fix. 🛥 it.

braver:
And released. 👍 We're running out of version numbers 😁

Issue #930, opened 2018-02-02
Improve logging


kaste:
Goals:

- [x] No interleaved printing in a multi-threaded environment
- [x] Associate a log line with the caller context which is the thread, the file, and its line 
- [x] Configurable, so we can have more or less noise in the console. The default for our current `debug: true` would probably be 'INFO', for `debug: false` it would still be either 'ERROR' or 'WARNING'
- [x] Maybe a hook ('Handler') that wires up 'ERROR' messages with a toast plugin or an (yet to come) error panel, or a just status bar message to keep it simple 🌶 
- [x] If the user says `debug: false` (the current default) we do not format the messages fancy with thread, filename, line; we just print the actual *message* 🎆 
- [x] A plugin author justs import the std python *logging* library and inits it to e.g. `getLogger('SublimeLinter.flake8')` and then logs away, and it will cooperate with our Logger. T.i. the author will not use `persist.debug` anymore ❤️ 

Related #909

FichteFoll:
If the default for `debug` is supposed to be 'INFO', where does 'DEBUG' come in?
In my packages, I allow configuring the level pretty directly through the settings. No need to abstract over that imho.

https://github.com/SublimeText/PackageDev/blob/cdc6d33f6de672d577790f87e9fb207bd7b23131/Package/PackageDev.sublime-settings#L2-L4

And I actually think that the best default for SublimeLinter currently would be INFO, since there is quite some useful information there such as the extracted PATH from the user's shell or similar.

Like the idea for having a handler to fire toast messages (if implemented).

braver:
I think noise control is a bit much here. End users will not want to have anything printed except serious errors, or debug info to help with configurations (i.e. "on" or "off"). Developers will want fine grained control, i.e. logging of the thing they're working on in addition to the errors. 

I think 1 and 2

braver:
Sorry, wrong button

kaste:
3 is a builtin functionality of the std logging library. We must translate our *true|false* to the appropriate so-called log levels. But then we also accept `debug: 'debug'` bc that's the name of such a level. Reducing that functionality would be more work than utilizing it.

braver:
👍🏻 

So, DEBUG equals INFO then.

kaste:
Added two more goals. 

[...]

PR #961, opened 2018-02-05
Use thread_name for additional info in debug messages

kaste:
Code to explain #930 @FichteFoll

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -57,20 +57,30 @@ def run_tasks(tasks, next):
     sublime.set_timeout_async(lambda: next(errors))
 
 
+task_counter = 0
+
+
 def get_lint_tasks(linters, view, hit_time):
+    global task_counter
+
     for (linter, settings, regions) in get_lint_regions(linters, view):
         tasks = []
         for region in regions:
             code = view.substr(region)
             offset = view.rowcol(region.begin())
+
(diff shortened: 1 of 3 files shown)


FichteFoll:
I have a better idea. Just a moment.

FichteFoll:
This is my proposal: https://github.com/FichteForks/SublimeLinter3/commit/afeffec32923641b8ebd1bcf360e4d972b9cb2eb

FichteFoll:
Note that this assumes that `next(task_counter)` is thread-safe, which it is probably not, so I guess I need a mutex here.

FichteFoll:
Actually, is `lint_view` ever called concurrently? Because the lint tasks are created in order (even through `run_concurrently`) on the thread that `lint` was called on and before the executor comes in.

kaste:
It's basically concurrency hell here bc `queue.queue_lint` (or something similar named) already is on a separate thread. 😃

FichteFoll:
I'll take that as a yes.

kaste:
Btw your base backend.py is 2 iterations behind master.

FichteFoll:
Yeah I need to merge master, but I'm not confident it would be pulled anyway so I'm not spending time on it until I get a positive sign on that.

PR #965, opened 2018-02-06
do something about invalid settings on load

braver:
I'm not sure it's actually feasible to check the user settings on load. We depend on the ST api which merges the settings immediately, so we can't really use our own defaults if the settings fail. I think. Perhaps someone else has a better idea. The least we can however, is message to the user, so this PR offers that. In combination with #959, at least the user will be aware of his mistake which is essentially what this is about: help the user fix his configuration. If the user repairs his settings, SL should recover and continue working again, so there may not even be a reason to make this more complicated.

### DIFF
--- a/lint/settings.py
+++ b/lint/settings.py
@@ -12,6 +12,7 @@ class Settings:
 
     def load(self):
         """Load the plugin settings."""
+        validate_settings()
         self.observe()
         self.on_update()
 


kaste:
Esp. I doubt that aborting in `on_update()` if the settings are invalid does help the user. The whole system (SL) will run anyway. With or without 'valid' settings. The validation therefore is *only* to inform the user.

braver:
Oh, you're right. Because we don't use the panel yet it just gets lost in all the other updates during load. We don't need changes here, we need feedback. As you say, it'll run anyway. Closing this and the related issue, #959 is the real deal.

PR #966, opened 2018-02-07
Tackle clear methods

### LINKED ISSUE #899: Factor out "clear methods" in linter

kaste:
Fixes #899

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -622,9 +622,6 @@ class Linter(metaclass=LinterMeta):
         view.erase_status(STATUS_KEY)
         highlight_view.clear_view(view)
 
-    def clear(self):
-        self.clear_view(self.view)
-
     @classmethod
     def text(cls, view):
         """Return the entire text of a view."""
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -608,11 +608,6 @@ class Linter(metaclass=LinterMeta):
 
             del persist.view_linters[vid]
 
-    @classmethod
-    def clear_all(cls):
-        """Clear highlights and errors in all views."""
-        persist.errors.clear()
-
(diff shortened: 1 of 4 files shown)


lint/linter.py
```diff
@@ -559,12 +558,12 @@ def assign(cls, view, linter_name=None, reset=False):
         vid = view.id()
         persist.views[vid] = view
         syntax = util.get_syntax(view)
+        persist.debug("detected syntax: {}".format(syntax))
 
-        if not syntax:
-            cls.remove(vid)
+        if not syntax:  # seems like a very rare, edge case
```
kaste: Is it possible that a view has no syntax?

braver: I don't think so. `get_syntax` is rigged to return an empty string in some cases, but I don't think Sublime Text has views without a syntax. Perhaps you can use the syntax map to map to `''` as a syntax to make this happen? Seems like a weird edge case anyway. Perhaps it's a kind of defensiveness that made sense during refactoring this, but no longer.

braver:
I sometimes get missing icons in the gutter. It does have a region I can hover over to get the popup, but no icon (line 599):

screen shot 2018-02-08 at 11 20 24

Hm, this happens on master as well. 

What I'd really like to see is the styles to update immediately, but apparently that doesn't work on master anymore either. 

Enabling and disabling works well on this branch. 

So, without digging any further, this PR looks good for me but we may have some (maybe related, maybe not) minor regressions.

PR #969, opened 2018-02-09
Clear state before parsing the settings

### LINKED ISSUE #967: apply new linter-specific styles without ST restart
This is a recent regression where creating a linter specific style doesn't apply that style until ST has been restarted.

kaste:
Fixes #967


braver:
LGTM ⛵️ it

PR #971, opened 2018-02-09
Draw the protected regions first so that ST prioritizes the actual re…

### LINKED ISSUE #968: missing gutter icons
Sometimes not all icons are rendered in the gutter, apparently primarily during editing. Discovered while testing #966, note the absent icon on line #599:

kaste:
…gions

Fixes #968


braver:
Does seem to help, I don't see the problem anymore. I do see what looks like conflicts with gitgutter's popup, but that's not introduced in this PR. 

🛥 it

PR #973, opened 2018-02-09
Change 'python' setting semantics

### LINKED ISSUE #945: How do I get flake8 to work with different versions of Python?
I've tried setting `@python` to "/usr/bin/python" and "2.7", but it just uses `/usr/local/bin/flake8`, which is linked with Python 3, instead.

Is there a way to tell SublimeLinter to run `/path/to/python -m flake8` instead of looking for a `flake8` command?

kaste:
'python' points to a python binary and generally executes 'python -m script_name'.
If you set 'python' to a version string, it will search PATH for such a
binary.  See discussion in #945 

Fixes #945

(Old behavior: python points to a python environment and then executes 'python/bin/script_name'.)

### DIFF
--- a/lint/base_linter/python_linter.py
+++ b/lint/base_linter/python_linter.py
@@ -24,13 +24,12 @@ class PythonLinter(linter.Linter):
 
     - Support for a "python" setting.
 
-      python can be set to a version, in which case we try to find
-      a python on your system matching that version.
+      python should point to a python binary on your system. Alternatively
+      it can be set to a version, in which case we try to find a python
+      binary on your system matching that version (using PATH).
 
-      python can alternatively set to a base path of a python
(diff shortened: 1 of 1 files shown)


braver:
This is very 👍 IMO. I don't have the setup to really test this, or the inclination to dig into it, but it does LGTM just from reading it.

FichteFoll:
Not sure if it matters here, but on windows we could first try the py launcher with the version as an argument, i.e. `py -3.6 --version` and if that works, use that instead of trying to resolve stuff manually. It's also venv aware, but only if the environment vars have been set.

FichteFoll:
No, this can obviously only be used if it's available, but can be prioritized over the manual lookup.

PR #974, opened 2018-02-10
Iterate on controller/queue/backend integration

### LINKED ISSUE #758: Last hit_time is irrelevant to invalidate/abort a lint

### LINKED ISSUE #838: Trigger lints based on `buffer_id` not `view_id`
Otherwise, we get multiple, concurrent lints per buffer if the user has multiple views into the same file. 

E.g. changing a setting, triggers a fresh lint per view

kaste:
- Factor proper debounce function which can be reused
- Exchange 'hit_time' with 'change_count' - #758 
- Exchange 'view_id' with 'buffer_id' (not completely!) - #838 

Fixes #838 
Fixes #758

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -20,7 +20,7 @@ task_count = count(start=1)
 counter_lock = threading.Lock()
 
 
-def lint_view(view, hit_time, next):
+def lint_view(view, view_has_changed, next):
     """
     Lint the given view.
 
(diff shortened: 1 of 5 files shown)


braver:
Excellent, much better. LGTM 💯

PR #975, opened 2018-02-10
Print user messages in `context_sensitive_executable_path`

### LINKED ISSUE #837: `context_sensitive_executable_path` should be responsible to print a useful error
Bc it can provide a detailed reason. Currently we print in `build_cmd` which does not have a good enough context to be helpful here.

kaste:
Fixes #837


braver:
LGTM ⛵️

PR #976, opened 2018-02-10
Use ${file_name}, ${tmpfilename}, and ${args} as special cmd identifiers

### LINKED ISSUE #810: Replace `@` and `*` in cmd with self-explanatory templates
Via https://sublimelinter.readthedocs.io/en/latest/linter_attributes.html#cmd,
`@` and `*` stand for the view's filename and for the computed args respectively.

It would be better if those used some more descriptive names like `{file_path}` and `{args}` (Python format syntax) or `$file_path` and `$args` (ST template format) respectively. Neither `{}` nor `$` should occur in command parameters regularly.

For backwards compat, the old placeholders could be treated as aliases.

kaste:
Fixes #810

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -717,26 +717,22 @@ class Linter(metaclass=LinterMeta):
 
     def insert_args(self, cmd):
         """Insert user arguments into cmd and return the result."""
-        args = self.build_args(self.get_view_settings())
-        cmd = list(cmd)
+        settings = self.get_view_settings()
+        args = self.build_args(settings)
 
-        if '*' in cmd:
+        if '${args}' in cmd:
+            i = cmd.index('${args}')
+            cmd[i:i + 1] = args
+        elif '*' in cmd:  # legacy SL3 crypto-identifier
             i = cmd.index('*')
-
-            if args:
-                cmd[i:i + 1] = args
-            else:
-                cmd.pop(i)
+            cmd[i:i + 1] = args
         else:
             cmd += args
 
         return cmd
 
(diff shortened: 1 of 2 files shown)


braver:
I think there are some comments in linter.py that need updating and perhaps we should put something in the docs about this too? Functionally etc this is 👍 👍

FichteFoll:
Note that this will only replace exact occurances of e.g. `${file_name}` in an individual arg, and not `$file_name`. For replacing the current cryptic symbols, this works, but I think we can do even better here.

I thought about using `sublime.expand_variables` as we do for the settings, so linters could also access e.g. the file's folder on the command line without custom python code. `args` could be replaced with a placeholder, such as some special unicode character or a control character that you wouldn't see in command lines, which would then be replaced with multiple arguments once the command line has been split. But that's just an idea since idk whether any linter currently has a use for this.

kaste:
This currently out of scope and was not the intention of the original issue. Note that this `expand_variables` thing would require a refactoring, and we're speaking of 10 or more commits here. 

We first need a use case for this.

kaste:
Hm, I forgot to update the comments.

kaste:
Actually, I don't find the comments.

braver:
@kaste at line 364 of linter.py?

Currently we don't have docs for plugin authors that cover these features. Now linter_attributes.rst still speaks of `'@'` and of using `'-'` for `tempfile_suffix`. Can you update those docs?

kaste:
line 364 in linter.py is unrelated

PR #977, opened 2018-02-10
Remove config_file setting

### LINKED ISSUE #891: Remove `config_file` setting
We have special handling for *config files*. Like `config_file = ('--config', '.jshintrc')`. These are set on `cls` so basically configurable for plugin authors. 

Now, the 90% case is probably these days that linters find their configuration automatically, esp. if you start with a good working_dir, which we probably do, right now. 

If that's not the case, a *user* setting is probably way better and more useful. WDYT?

(Note: The last remaining lrucached function in util.py would go away with that. 😄)

kaste:
Fixes #891

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -330,23 +330,6 @@ class Linter(metaclass=LinterMeta):
     # this attribute to the other stream.
     error_stream = util.STREAM_BOTH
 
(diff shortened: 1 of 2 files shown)


braver:
LGTM 👍

PR #980, opened 2018-02-10
Simple logging using the standard library

### LINKED ISSUE #930: Improve logging

kaste:
Fixes #930

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -4,6 +4,7 @@ from functools import lru_cache
 from numbers import Number
 import threading
 
+import logging
 import os
 import re
 import shlex
@@ -11,6 +12,10 @@ import sublime
 
 from . import persist, util
 from .const import WARNING, ERROR
+
+
+logger = logging.getLogger(__name__)
+
 
 ARG_RE = re.compile(r'(?P<prefix>@|--?)?(?P<name>[@\w][\w\-]*)(?:(?P<joiner>[=:])(?:(?P<sep>.)(?P<multiple>\+)?)?)?')
 NEAR_RE_TEMPLATE = r'(?<!"){}({}){}(?!")'
@@ -125,8 +130,8 @@ class LinterMeta(type):
                 if isinstance(syntax, str) and syntax[0] == '^':
                     setattr(cls, 'syntax', re.compile(syntax))
             except re.error as err:
-                util.printf(
-                    'ERROR: {} disabled, error compiling syntax: {}'
+                logger.error(
+                    '{} disabled, error compiling syntax: {}'
                     .format(name.lower(), str(err))
                 )
                 setattr(cls, 'disabled', True)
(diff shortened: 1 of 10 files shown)


braver:
It looks good basically, but it seems "info" level logs don't get printed when I set debug to true. For the end user I expect there to be just an on-and-off switch and "info" has important information right now.

kaste:
? This is what I get with `debug: true`

image

braver:
Specifically I was missing the version info that’s printed on start up.

kaste:
Good catch. This is random behavior actually. 

I thought naming the file `_logging.py` would ensure that its `plugin_loaded()` callback would be invoked before any other thing. But that's not the case, loading order here is random.

kaste:
I'm not so much into log message styling. So this LGTM.

braver:
👍

braver:
#909 can be closed then I think?

kaste:
Current status: 12 issues and 0 PRs.

kaste:
image

braver:
🎉 Let's freeze it, check some docs and then release it. I'm away-from-laptop from Thursday till the end of next week, so let's pin release day on Feb 24?

braver:
Just in time too, I just tagged the penultimate pre-release @ `4.0.0-rc.9.9.8`. After `4.0.0-rc.9.9.8` the numbering becomes super ridiculous.

lint/linter.py
```diff
@@ -921,9 +925,19 @@ def lint(self, code, hit_time, settings):
         - If the view has been modified since the original hit_time, stop.
         - Parse the linter output with the regex.
         """
+        # We 'name' our threads, for logging purposes.
+        threading.current_thread().name = task_name
```
FichteFoll: I believe this should be happening in `execute_lint_task` instead. `Linter.lint` does not need to know or care about the backend being used and how tasks are scheduled.

lint/log_handler.py
```diff
@@ -0,0 +1,114 @@
...
+        formatter = TaskNumberFormatter(
+            fmt="SublimeLinter: {LEVELNAME}{message}",
+            style='{')
+        handler = logging.StreamHandler()
+        handler.setFormatter(formatter)
+    else:
+        if level is True:
+            level = DEBUG_TRUE_LEVEL
+        else:
+            level = logging.getLevelName(level.upper())
```
FichteFoll: Shouldn't this be `getattr(logging, level.upper())`? `logging.getLevelName` is the reverse operation.

lint/linter.py
kaste: Yeah, that would be better

lint/log_handler.py
kaste: Hm, I have it from here https://github.com/python/cpython/blob/504f19145ca5738162d6a720fa45b364ac8c0384/Lib/logging/__init__.py#L120-L141

FichteFoll: Yeah, but you already have a string in `level` and want the numeric level for `handler.setLevel`.

kaste: You did not read the code, did you

FichteFoll: No, I actually did not. I only read the docstring and it looks like this behavior is undocumented (or rather a backwards compatibility thing that has been removed for 3.4) which is why this behavior confused me. And admittedly, for a function named `getLevelName` to return an integer with the level (and not its name) *is* pretty confusing.

Related upstream issue: https://bugs.python.org/issue22386

PR #983, opened 2018-02-12
Extract tooltip view

kaste:
Tooltip view is the last one. Now, `sublime_linter.py` is startup plus backend controller.


braver:
👍🏻

PR #984, opened 2018-02-12
5 safe(?) simplifications in sublime_linter.py

kaste:


braver:
I was kinda expecting `view_has_file_only_linter` to do something useful, apparently not. All of this works just fine on my end. ⛵️ it.

PR #992, opened 2018-02-14
Support wsl env

kaste:
Fixes https://github.com/SublimeLinter/SublimeLinter-eslint/issues/179

Note that this is on top of #989 bc it was easier that way. Needs cooperation from OP bc I cannot test that.

@CrshOverride

### DIFF
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -116,19 +116,23 @@ class NodeLinter(linter.Linter):
         return self.find_ancestor_cmd_path(cmd, cwd)
 
     def find_ancestor_cmd_path(self, cmd, cwd):
-        """Recursively check for command binary in ancestors' node_modules/.bin directories."""
-        node_modules_bin = os.path.normpath(os.path.join(cwd, 'node_modules/.bin/'))
+        """Recursively check for command binary in node_modules/.bin directories."""
+        node_modules_bin = os.path.normpath(
+            os.path.join(cwd, 'node_modules/.bin/'))
+        path_to_cmd = os.path.join(node_modules_bin, cmd)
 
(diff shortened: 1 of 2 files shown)


braver:
Ok, it didn't break SL for me on a Mac and I'm not going to test using WSL because I don't have a (virtual) machine ready to do so. So, LGTM ⛵️

kaste:
I doubt that a simple python `popen` spawns into the wsl shell. I don't know if you can make the linux subshell the default on windows.

braver:
I probably have to set up a win10 machine next week anyway, so I can check then. As for default shells, I bet there is a registry thing for it....

kaste:
Honestly, @CrshOverride could just test this. 🙄

kaste:
Okay, this alone does not work. We need to return `['wsl', 'path/to/eslint']`. (Maybe?)

caylorme:
Please see my comment on the issue for ESLint plugin: https://github.com/SublimeLinter/SublimeLinter-eslint/issues/179#issuecomment-378829613

I believe I could assist with implementing a proper fix to have this work under the scenario where you have a Windows instance of Sublime Text and node/eslint installed under WSL.

alexkavon:
If you need to convert a path. WSL has a wslpath tool:

['wsl', 'wslpath', '$filepath']

This will convert a windows path like 'C:/Users/myuser/code' to '/mnt/c/Users/myuser/code'

kaste:
Ha, cool, but I think we need a python version here. A command can have several paths in it and calling wsl for each of it maybe a bit too much overhead.

Let's say we don't discover wsl automatically, we could as a first step patch finalize_cmd in linter.py to convert all paths in cmd iff the first item on cmd equals wsl.

PRs? Anybody wants to try this?

PR #993, opened 2018-02-14
Refactor `Linter.assign` and remove cruft

kaste:
Basically simplify `assign`and `register_linter`

**Important**: **Removes** `initialize` and `reinitialize`, which nobody implemented (if I checked correctly). (We used it for RubyLinter.)

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -168,10 +168,11 @@ class LinterMeta(type):
     def register_linter(cls, name):
         """Add a linter class to our mapping of class names <-> linter classes."""
         name = name.lower()
+        reloading = name in persist.linter_classes
         persist.linter_classes[name] = cls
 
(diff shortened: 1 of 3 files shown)


kaste:
Should I continue to work against master?

braver:
Unless you create a release (i.e. tag it) no changes get pushed to anyone, so working on master is safe in that sense. Working on master makes it easier to track what's going on for others, so that's easiest. I can always make a diff later if I want to review anything, or just scroll through the commit messages.

It seems there are a bunch of people already trying the beta, so also feel free to make releases if you think that makes sense. There's a stupid issue in package control about the pre-release semver numbering, so the next release would be `4.0.0-rc.9.9.9.0`.

kaste:
Seems like package control annoys me today.

PR #995, opened 2018-02-14
add post-install/upgrade messaging

### LINKED ISSUE #991: Updating SublimeLinter should just work, no?
It seems that every (silent) update breaks SublimeLinter and requires a restart. I guess you, @FichteFoll , should know everything about this. 😄 I know nothing about this and am a bit disappointed that PackageControl can't just do it. What to do?

- Toast the user after every update, that a restart is required
- Fix the underlying issue

braver:
Fixes #991 by using the message panel to tell users to restart after installation and upgrades.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -44,6 +44,12 @@ def plugin_loaded():
     log_handler.install()
     backup_old_settings()
 
+    from package_control import events
+    if events.install('SublimeLinter'):
+        util.message('SublimeLinter has been installed. Please restart Sublime Text.')
+    elif events.post_upgrade('SublimeLinter'):
+        util.message('SublimeLinter has been upgraded. Please restart Sublime Text.')
+
     persist.settings.load()
     logger.info("debug mode: on")
     logger.info("version: " + util.get_sl_version())


kaste:
Where the hell did you get this API?

kaste:
Obviously, nobody can test this, 🤣😵.

So, it's in plugin loaded hook. What else need I know?

braver:
The API is here: https://packagecontrol.io/docs/events

I basically just slightly tweaked the example code on that page, so it "should work", but yeah, kinda hard to test.

sublime_linter.py
```diff
@@ -44,6 +44,12 @@ def plugin_loaded():
     log_handler.install()
     backup_old_settings()
 
+    from package_control import events
+    if events.install('SublimeLinter'):
+        util.message('SublimeLinter has been installed. Please restart Sublime Text.')
```
FichteFoll: Why would you need to restart when SL is installed? I mean, unless you already had other linter plugins installed that depend on it.

braver: I guess only if you already had some linter plugins installed and then (re-) installed SL.

PR #999, opened 2018-02-16
Update docs

kaste:
Esp. change what we *want* the users to use, not what we actually support. E.g. `cmd` should be a tuple, not a list or string. E.g. `syntax` should be a tuple, not a list, not a regex pattern.

Do not document methods we don't want the users to override. There should be no need to override `tmpfile`, `communicate`, and `run`.


braver:
🎉

FichteFoll:
`run` is interesting for linters that do not run an external linter, i.e. annotations and sublimesyntax.

I didn't check, but is the current behavior of the `python` setting explained?

kaste:
I would prefer for the future, people would just use `lint`. So they don't have to somehow construct fake output which gets then parsed via a regex. Using `lint` they just had to return `[error]`. This is a bit undecided still. For example, one could say `lint` is too complicated to use atm.

FichteFoll:
Fair enough. The sublimesyntax linter already receives string output from the internal linter, so I'll continue to use that until it "stops working", but I can see that overriding `lint` is a better idea. Now if only "error" objects (created by `process_match` were namedtuples and not some dict ;)

braver:
Excellent, 🚢 it.

Issue #941, opened 2018-02-03
De-highlight (invalidate) error as soon as your editing on it


kaste:
Take a normal spell checker. It squiggles unknown or wrong words. But as soon as you *edit* (**not** *move* the cursor into) such a word, the squiggles disappear (and then maybe reappear of course, if you're still incorrect).

braver:
I would totally love this 👍🏻👍🏻

FichteFoll:
Just to be clear, this is only about the word that was changed and not all markers, right?

braver:
That’s my reading of it too, just the region(s) at the cursor(s) that is(are) being changed.

PR #1002, opened 2018-02-17
Switch to fallback if region contains *some* WS

kaste:


braver:
I don’t think we should do this. In general we don’t highlight the full erroneous region. For one because not all linters report like that, but also because

- it doesn’t usually convey additional useful information 
- the UX is horrible for the underlined styles

The highlighting only hints at the problematic area, the message will tell you more about what’s wrong. If there is anything at all that we can highlight with underlines, we don’t need to highlight anything else with outlines.

braver:
Uhm, let’s not close until I gave it a whirl though

kaste:
I don't think you understand that PR. This was actually a no-brainer I would have merged today.

braver:
Oh I do understand. But you’re moving in the wrong direction. We should not highlight regions ever. Perhaps #941 will convince me otherwise about the UX part of it. However, I also think it’s an incorrect interpretation of the output of a linter.

Your example should be solved by highlighting only the first {.

kaste:
The main feature of SublimeLinter is to highlight *regions*. We usually highlight words, although the linters usually only tell us about the beginning of an error. 

If you have an unused import in flake8, we highlight that name, although flake8 only tells us the line (not even a column). (This is what the 'near' property is all about, and our main fn `reposition_match`. (I just extracted that fn, before it was in 'highlights.py'.))

Besides that, you rarely ever see WS in regions bc we prevent that. We match using our 'WORD_RE' at word boundaries.

braver:
If there is a solution for #941 I’m open to completely revising my opinion on this. But I really want to see a solution for the UX problem.

braver:
Hmmmm, this also means all multi-line regions would switch to outlines anyway, and they merge. 🤔 that also solves most of the UX issues probably. That may be kinda unexpected behavior, but it is *good*.

Ok, a bit hard to imagine how all this works out, but I think I’m convinced. It may actually solve a bunch of reservations I have about highlighting more than a single word. Good. 

Ok, let’s roll with this. Sorry, but I had to come all the way from seriously opposed to the concept to seeing the light. I think I see it now. I’ll get out of your hair. Please carry on 👍🏻

FichteFoll:
I haven't tested this, nor do I knowingly use a linter that makes use of this (besides rustc I guess?), but I generally support this idea. :+1:

PR #1004, opened 2018-02-19
De-emphasize highlights while editing

### LINKED ISSUE #941: De-highlight (invalidate) error as soon as your editing on it

kaste:
- Implement IdleViewController
- De-highlight some regions (none, warnings, ws_errors) while editing, these regions re-appear 'on idle', e.g. after x seconds, or after save
- De-highlight the word(s) under the cursor while editing, these region re-appear as soon as we get updates from the backend

Fixes #941

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -28,7 +28,7 @@
     // Essentially picks a predicate fn from `highlights.py` to hide some
     // lint errors while editing (busy). These errors appear after
     // `time_to_idle` seconds or immediately on save.
-    "highlights.demote_while_editing": "ws_regions",
+    "highlights.demote_while_editing": "none",
 
     // Lint Modes determine when the linter is run
     // background: asynchronously on every change
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -21,6 +21,15 @@
     // to an icon in the theme (they usually include "warning" and "error")
     "gutter_theme": "none",
 
+    // EXPERIMENTAL: See below `demote_while_editing`.
+    "highlights.time_to_idle": 5,
+
+    // EXPERIMENTAL: One of 'none', 'some_ws', 'warnings'
+    // Essentially picks a predicate fn from `highlights.py` to hide some
+    // lint errors while editing (busy). These errors appear after
+    // `time_to_idle` seconds or immediately on save.
+    "highlights.demote_while_editing": "none",
+
     // Lint Modes determine when the linter is run
     // background: asynchronously on every change
     // load_save: when a file is opened and every time it's saved
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -24,11 +24,11 @@
     // EXPERIMENTAL: See below `demote_while_editing`.
     "highlights.time_to_idle": 5,
 
-    // EXPERIMENTAL: One of 'none', 'some_ws', 'warnings'
+    // EXPERIMENTAL: One of 'none', 'ws_regions', 'warnings'
     // Essentially picks a predicate fn from `highlights.py` to hide some
     // lint errors while editing (busy). These errors appear after
     // `time_to_idle` seconds or immediately on save.
-    "highlights.demote_while_editing": "none",
+    "highlights.demote_while_editing": "ws_regions",
 
     // Lint Modes determine when the linter is run
     // background: asynchronously on every change
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -31,6 +31,8 @@ UNDERLINE_STYLES = (
 
 SOME_WS = re.compile('\s')
 FALLBACK_MARK_STYLE = 'outline'
+
+WS_REGIONS = re.compile('(^\s+$|\n)')
 DEMOTE_WHILE_BUSY_MARKER = '%DWB%'
 
 highlight_store = style_stores.HighlightStyleStore()
@@ -107,7 +109,7 @@ def get_demote_predicate():
     if setting == 'none':
         return demote_nothing
 
-    if setting == 'some_ws':
+    if setting == 'ws_regions':
         return demote_ws_regions
 
     if setting == 'warnings':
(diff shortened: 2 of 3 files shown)


kaste:
0o9wwvmtmw

braver:
🌶

kaste:
Notable: If you set the TIME_TO_IDLE to a large number, the e.g. ws regions effectively appear on save only.

braver:
The only "complaint" I have here is that all range highlights disappear, not just the one you're currently editing.

de-highlight

kaste:
That's the 'de-highlight on busy' feature, not the 'de-highlight word under cursor' (LOL).

'de-highlight on busy' is general pluggable. (Although I did not (yet) expose any settings.) What you're seeing here is 'de-highlight all regions that contain ws' (These are the same that apply to our fallback method for underlines). I would like to see another predicate-fn, that selects all regions only containing whitespace *or* span multiple lines. (Other than that, we have the 'none' predicate (turn off the feature) and select warnings predicate.

braver:
I find the "blinking" highlights outside what I'm currently focussing on (ie. editing) very distracting actually. What I'd like is to have all types of highlights, also without whitespace, to disappear when busy. In general, and here too, I expect that if we get it right there doesn't have to be a setting because it'll just feel natural.

kaste:
C'mon @braver, in the code it says

```
TIME_TO_IDLE = 2  # [seconds]  2 is good for testing, but noisy later on
```

For this PR, you really have to read and edit the code and experiment a bit.

And your suggestion is way too opinionated, it is meant to be pluggable at least during beta as long as its in master. Bc you really have to code with such UI features to get a feeling. E.g. if you set TIME_TO_IDLE to 300 or even more (that's how long it often takes for me to get the next edit) you basically don't see warnings or ws errors until you hit save, then you should see them immediately. What you suggest is probably a predicate fn  

```
def demote_everything(*args, **kwargs):
    return True
``` 

I would hate it (maybe, this is still an UI experiment) bc I want to see the typos in fn or var names immediately. 

It is important to get this into master somehow, so I/anyone can actually use it while doing real coding. (For me: I have to switch SL branches too often, so I did not had this feature reliable for long enough time.)

braver:
> And your suggestion is way too opinionated

Hey, I'm a designer, having strong opinions (weakly held) is what I do 😉 

> For this PR, you really have to read and edit the code and experiment a bit.

We should iterate and discuss, which is what we're doing right now. 

I guess I *should* dig into the code and tweak it a bit, but for me that always takes a lot of extra effort. I'm really not even sure I *can* change it to my liking, I'm just not much of a programmer. I can try, but at least I'll just give you the feedback now.

> I would hate it (maybe, this is still an UI experiment) bc I want to see the typos in fn or var names immediately.

Good point, totally agree. 

I'm fine with merging this quickly and tweak it later except for this point:

De-highlighting anything that's not under my cursor attracts my attention and achieves the opposite of the goal of this PR. This isn't about the idle timeout, that's fine, it's about the immediate de-highlighting of regions not under the cursor. Our brain is super wired to notice changes like that.

kaste:
Actually, master is really irritating here. Bc it is *way* faster than SL3 I see these syntax errors and ws errors popping up everywhere while editing this exact line. Don't forget that if you set up IDLE_TIME these errors don't appear while editing in the first place, so they don't disappear as well.

braver:
> Don't forget that if you set up IDLE_TIME these errors don't appear while editing in the first place, so they don't disappear as well.

Oh that's true for new errors. But what about errors that are already there and that I don't want to bother with right now? In my [gif](https://github.com/SublimeLinter/SublimeLinter/pull/1004#issuecomment-367991240) eslint complains about the quotes on line 114, but I don't care. Now that I'm editing the if block at 110, and perhaps I'm slow and need to look up stuff, and the error at line 114 blinks out every time I start typing again. There is no reason for that error to disappear, it doesn't make sense to me that it does, and it attracts my attention.

braver:
You are right though, master is a bit annoying right now. Guess I never noticed because the last time I really used ST I didn't have the outline fallback and I didn't notice the whitespace errors popping up all the time. So, I guess I understand where you're coming from with this one, and we need to figure out a good middle ground. 

So, I agree with you now. Let's merge this and then think about it some more.

kaste:
Okay, added settings, so this is pluggable by the end-users. Marked as an experiment, we should add predicate functions as we wish. I did not add the 'True' predicate which just demotes everything yet.

SublimeLinter.sublime-settings
```diff
@@ -21,6 +21,15 @@
...
     "gutter_theme": "none",
 
+    // EXPERIMENTAL: See below `demote_while_editing`.
+    "highlights.time_to_idle": 5,
+
+    // EXPERIMENTAL: One of 'none', 'ws_regions', 'warnings'
+    // Essentially picks a predicate fn from `highlights.py` to hide some
+    // lint errors while editing (busy). These errors appear after
+    // `time_to_idle` seconds or immediately on save.
+    "highlights.demote_while_editing": "ws_regions",
```
braver: I'd say `hide_while_editing`, because that's easier to understand (the word "hide" is also used 2 lines earlier, so it would also be more consistent).

That `ws_regions` means "regions containing whitespace isn't super clear. 

I'll leave it to you if you want to bikeshed about this right now. It's ok with me if you want to merge this as is, because it's good stuff and it's clearly marked as experimental so people can expect things to change here.

kaste: I already made these changes, but I revert and take 'demote' now. Currently, I actually hide these regions. But, we could also (if I were not that dumb with these coloring/scope stuff) de-emphasize them using a soft gray.

kaste:
Ah, fuck it. This was actually a relatively hard problem. The first commit here is a tricky one.

SublimeLinter.sublime-settings
braver: Oh that’s also not a bad idea, using the `’comment’` scope for color when busy

PR #1006, opened 2018-02-20
Enable hot reloading SublimeLinter and its linter plugins

kaste:
~~TODO: Old log handler does not get removed during hot reload, so we end up with multiple log messages.~~

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -20,6 +20,11 @@
         }
     },
     {
+        "caption": "SublimeLinter: Reload SublimeLinter and its Plugins",
+        "command": "sublime_linter_reload"
+
+    },
+    {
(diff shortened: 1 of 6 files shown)


braver:
Woooo 🔥

Default.sublime-commands
```diff
@@ -19,6 +19,11 @@
             "wrap": true
         }
     },
+    {
+        "caption": "SublimeLinter: Reload SublimeLinter and its linters",
```
braver:
```
"caption": "SublimeLinter: Reload SublimeLinter and Plugins",
```
❓

braver:
Ok, that's a bit hard to test for real, but the parts of it seem to work. Ship it :shipit:

PR #994, opened 2018-02-14
User defined syntaxes

kaste:
🌶 This allows to extend the syntax a linter plugin might support. Since we all know, eslint can basically lint everything. A user can e.g. define
 ```
eslint: {syntaxes: ['php', 'coffeescript', 'python', 'rust', 'java']}
```

in the global or project settings.

**Notable**: **Removed** `can_lint_syntax` API which was again unused by the community.

PR #1007, opened 2018-02-20
Choose linters based on selectors

kaste:
~~On top of #994 (bc of the refactoring)~~

~~TODO: Does *not* check on each hit, only when the view's syntax changes.~~

~~TODO: Refactor, `view_syntax` is probably unused. Remove user-defined-syntaxes feature (#994).~~ 

Otherwise this is how I understand #691 

Tested with the following ESLint

image

If you take your unpatched eslint, you need to specify a user setting `selectors` which will then take precedence over everything the plugin author defined.

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -119,17 +119,14 @@ def translate_lineno_and_column(errors, offset):
 def get_lint_regions(linters, view):
     syntax = util.get_syntax(view)
     for (linter, settings) in linters:
-        selectors = settings.get('selectors')
-        if selectors:
-            joined_selectors = '|'.join(selectors)
-
-            # Inspecting the first char mimics 'file type' matching
-            if view.score_selector(0, joined_selectors):
+        selector = settings.get('selector')
+        if selector:
+            # Inspecting just the first char is faster
+            if view.score_selector(0, selector):
                 yield linter, settings, [sublime.Region(0, view.size())]
             else:
                 yield linter, settings, [
-                    region
-                    for region in view.find_by_selector(joined_selectors)
+                    region for region in view.find_by_selector(selector)
                 ]
 
             continue
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -119,6 +119,16 @@ def translate_lineno_and_column(errors, offset):
 def get_lint_regions(linters, view):
     syntax = util.get_syntax(view)
     for (linter, settings) in linters:
+        selectors = settings.get('selectors', False)
+        if selectors:
+            for selector in selectors:
+                regions = view.find_by_selector(selector)
+                if regions:
+                    yield linter, settings, [region for region in regions]
+                    return
+            return
+
+        # legacy
         if (
             syntax not in linter.selectors and
             WILDCARD_SYNTAX not in linter.selectors
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -156,10 +156,10 @@ def get_selectors(linter, wanted_syntax):
 
 def get_linters(view):
     filename = view.file_name()
-    vid = view.id()
+    bid = view.buffer_id()
 
     enabled, disabled = [], []
-    for linter in persist.view_linters.get(vid, []):
+    for linter in persist.view_linters.get(bid, []):
         # First check to see if the linter can run in the current lint mode.
         if linter.tempfile_suffix == '-' and view.is_dirty():
             disabled.append(linter)
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -8,7 +8,7 @@ import logging
 import os
 import threading
 
-from . import persist, util
+from . import util
 
 
 logger = logging.getLogger(__name__)
@@ -20,7 +20,7 @@ task_count = count(start=1)
 counter_lock = threading.Lock()
 
 
-def lint_view(view, view_has_changed, next):
+def lint_view(linters, view, view_has_changed, next):
     """
     Lint the given view.
 
@@ -39,14 +39,14 @@ def lint_view(view, view_has_changed, next):
     aggregated, and for each selector, if it occurs in sections,
     the corresponding section is linted as embedded code.
     """
-    linters, disabled_linters = get_linters(view)
+    enabled_linters, disabled_linters = filter_linters(linters, view)
 
     # The contract here is that we MUST fire 'updates' for every linter, so
     # that the views (status bar etc) actually update.
     for linter in disabled_linters:
         next(linter, [])
 
-    lint_tasks = get_lint_tasks(linters, view, view_has_changed)
+    lint_tasks = get_lint_tasks(enabled_linters, view, view_has_changed)
 
     run_concurrently(
         partial(run_tasks, tasks, next=partial(next, linter))
@@ -154,12 +154,11 @@ def get_selectors(linter, wanted_syntax):
             pass
 
 
-def get_linters(view):
+def filter_linters(linters, view):
     filename = view.file_name()
-    bid = view.buffer_id()
 
     enabled, disabled = [], []
-    for linter in persist.view_linters.get(bid, []):
+    for linter in linters:
         # First check to see if the linter can run in the current lint mode.
         if linter.tempfile_suffix == '-' and view.is_dirty():
             disabled.append(linter)
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -119,19 +119,19 @@ def translate_lineno_and_column(errors, offset):
 def get_lint_regions(linters, view):
     syntax = util.get_syntax(view)
     for (linter, settings) in linters:
-        selectors = settings.get('selectors', False)
+        selectors = settings.get('selectors')
         if selectors:
-            # Inspecting the first char mimics 'file type' matching
-            if any(view.score_selector(0, selector) for selector in selectors):
-                yield linter, settings, [sublime.Region(0, view.size())]
-                continue
+            joined_selectors = '|'.join(selectors)
 
-            # Now, search for embedded syntaxes
-            for selector in selectors:
-                regions = view.find_by_selector(selector)
-                if regions:
-                    yield linter, settings, [region for region in regions]
-                    break
+            # Inspecting the first char mimics 'file type' matching
+            if view.score_selector(0, joined_selectors):
+                yield linter, settings, [sublime.Region(0, view.size())]
+            else:
+                yield linter, settings, [
+                    region
+                    for region in view.find_by_selector(joined_selectors)
+                ]
+
             continue
 
         # Fallback using deprecated `cls.syntax` and `cls.selectors`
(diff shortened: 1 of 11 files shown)


kaste:
Okay, we could then put `cmd` into `defaults` as well. 

image

lint/linter.py
```diff
@@ -164,8 +164,13 @@ def __init__(cls, name, bases, attrs):
         if 'defaults' in attrs and attrs['defaults']:
             cls.map_args(attrs['defaults'])
 
-        if 'syntax' in attrs:
-            cls.register_linter(name)
+        if not cls.syntax and not cls.defaults['selectors']:
+            logger.error(
+                "{} disabled, either 'syntax' or 'selectors' must be specified"
```
FichteFoll: What about when a linter wants to lint all views (annotations)? I find the `*` special syntax to be an anti-pattern that we've been actively removing in the past and would suggest having neither a syntax nor selectors specification causes the linter to be active for all views.

lint/linter.py
```diff
@@ -1003,90 +1021,81 @@ def strip_quotes(text):
...
+
+        return syntax in syntaxes
+
     @classmethod
     @lru_cache(maxsize=None)
-    def can_lint(cls, syntax):
+    def can_lint(cls, _syntax=None):  # `syntax` stays here for compatibility
         """
-        Determine if a linter class can lint the given syntax.
+        Determine *eager* if the linter plugin can be used.
```
FichteFoll: eagerly

lint/base_linter/python_linter.py
```diff
@@ -39,23 +39,9 @@ class PythonLinter(linter.Linter):
...
-                can = True
-            elif isinstance(cls.syntax, str):
-                can = syntax == cls.syntax
-            else:
-                can = cls.syntax.match(syntax) is not None
-
-        return can
+    def can_lint(cls):
+        """Assume the linter can lint."""
+        return True
```
FichteFoll: I know this is technically not a part of this PR, but why override `can_lint` at all?

lint/backend.py
```diff
@@ -119,6 +119,16 @@ def translate_lineno_and_column(errors, offset):
...
     syntax = util.get_syntax(view)
     for (linter, settings) in linters:
+        selectors = settings.get('selectors', False)
+        if selectors:
+            for selector in selectors:
+                regions = view.find_by_selector(selector)
+                if regions:
+                    yield linter, settings, [region for region in regions]
+                    return
+            return
```
FichteFoll: Pretty sure these two `return`s should be `break`s.

FichteFoll:
Did you notice any slow down for hits? If it's noticeable, there are a couple places we can optimize.

1. As mentioned in https://github.com/SublimeLinter/SublimeLinter/issues/691#issuecomment-366935716, check the scope of the position `0` first and pick the entire view's region on success.
2. A list of multiple selectors can always be combined to a single selector with `|` (or `,`, same operation). That way we only call `find_selector` once per linter and let ST optimize this internally. I don't think we're doing anything with individual selectors, are we?

lint/backend.py
kaste: I should try gotos, haha. 👍

lint/base_linter/python_linter.py
kaste: I don't want the base classes can_lint. 🤔

FichteFoll: Ah right, the specialized linters don't rely on the global exeutable and instead have this context sensitive executable path thing.

lint/linter.py
braver: Can’t annotations specify the text and source scopes and thereby cover everything?

kaste: Yeah, * is not needed anymore bc it meant selectors based anyway. annotations automatically gets enabled when there are comments in the buffer, or something like that.

lint/backend.py
kaste: Second one is probably a continue

lint/linter.py
kaste: This check is heavily js influenced, and will just throw when selectors is not defined. Haha.

FichteFoll: Okay, different example: SublimeLinter-contrib-sublimesyntax. I need to check the view's file name (or its first line) to determine whether it is a syntax test file.

kaste: These are yaml files.?!

cls.selectors is optional

FichteFoll: Syntax test files use whatever syntax they test for and could be any.

I'll wrap this up in a separate issue if I don't forget about it. This was just something I noticed while browsing the diff.

FichteFoll:
Just a short question: As far as I understand, currently `view.find_selector` is called twice, once for determining whether a linter should be "active" and once to find the regions to lint. Couldn't we refactor this somehow to only have to call it once? 

As in, if there are no regions to lint, the linter isn't active (obviously), and to determine whether a linter *would* be active, we need to apply the selector on every hit anyway.
This doesn't apply to syntax-based selection of linters, but it would once the adaption code in #691 is applied. Could probably done in a separate PR though.

Another thing: Can we rename `selectors` to just `selector` and make it a single string? I don't see the need to have multiple strings when they can just combined tbh.

kaste:
Haha, you may noticed I already pushed 15 or more refactoring commits here. (The feature implementation is about 20 LOC.) Now, the big step here is, that we have to lift the functions in the multi-threaded backend. T.i. `can_lint` and of course instantiation of the linter (anyone implemented `__init__` here?) itself. 

The biggest performance gain would be, if we could call said function at least debounced, so they actually don't run `on_modified`, but at most every x seconds. (T.i. in `sublime_linter.py` move `check_linters_for_view` from `hit` to `lint`, 10 lines of code between them. (After the refactoring.)) But debounced is in a thread. And `check_linters_for_view` is not nearly thread-aware.

For the strings <-> array thing, I mean I just learned that like 6 hours ago. Yesterday I used `selectors` clearly as an array, today I just use a joined string. Please, take a look at the user requirements I posted in #691.

FichteFoll:
`can_lint` is run on every hit? That doesn't sound good tbh. Every lint should certainly suffice.

I believe the PR sufficiently adds the requested feature. Any optimizations and refactoring should be added eventually, but not necessarily in this PR. I mentioned I had a bit more time this week, but it turned out that there is apparently so much to do I am not able to make some time for SL stuff besides the usual comments and shallow reviews.

kaste:
Okay TODO then is: 

- `selectors` -> `selector` (type: `list -> str`)
- `score_selector` -> `match_selector` (I just saw this, already returns bool and does not have to compute a score.)
- Put `check_linters_for_view` after debounce and protect it with a Lock. (Actually, since we're *after* debounce, which should guarantee that this fn only runs every x seconds, races are already rare.)

✋ @FichteFoll

FichteFoll:
LGTM

kaste:
FWIW: in sublime_linter.hit

```
    import timeit
    res = timeit.timeit(lambda: check_linters_for_view(view), number=100)
    print('check_linters_for_view', res, view.buffer_id())

    res = timeit.timeit(lambda: view.match_selector(0, 'source.python'), number=100)
    print('match_selector', res, view.buffer_id())
	 
	...
```

Testing simple selector 'source.python'

```
small file:
check_linters_for_view 0.10854489439481085 100
match_selector 0.005745120881044841 100
score_selector 0.005787886407915721 100
find_by_selector 0.019786753976767102 100

>1000 LOC file
check_linters_for_view 0.14967934407945904 104
match_selector 0.010366363715675675 104
score_selector 0.007436497469129222 104
find_by_selector 0.06240816103102986 104
```

sublime_linter.py
```diff
@@ -191,7 +194,7 @@ def lint(view, view_has_changed):
     bid = view.buffer_id()
 
     # We're already debounced, so races are actually unlikely
-    with guard_check_linters_for_view[bid]:
+    with lock:
```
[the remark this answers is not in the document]
kaste: Doubt that this will work.

lint/linter.py
```diff
@@ -164,8 +164,13 @@ def __init__(cls, name, bases, attrs):
         if 'defaults' in attrs and attrs['defaults']:
             cls.map_args(attrs['defaults'])
 
-        if 'syntax' in attrs:
-            cls.register_linter(name)
+        if not cls.syntax and not cls.defaults.get('selectors'):
```
kaste: Oh, I have to change this one

kaste:
@braver This is working for me. 

Now, this is completely backwards compatible, as-in opt-in. AFAIK none of the plugins have to change, the user can opt-in by setting 'selector' and then this will take precedence. Also, a plugin author can opt-in by setting 'selector' and it will just work. The plugin author can ship both `cmd.syntax` and `defaults: {'selector'...}` and it should work for SL3 and SL4. SL4 will pick up 'selector', SL3 will pick 'syntax'.

braver:
@kaste what does your eslint look like for this?  I've got

```py
npm_name = 'eslint'
cmd = ('eslint', '--format', 'json', '--stdin', '--stdin-filename', '@')

crash_regex = re.compile(
        r'^(.*?)\r?\n\w*(Oops! Something went wrong!)',
        re.DOTALL
)
line_col_base = (1, 1)
defaults = {
        'selectors': ['source.js']
}
```

But it complains that `eslint disabled, either 'syntax' or 'selector' must be specified`.

Also, is `cmd` now configurable via defaults?

kaste:
Haha, you didn't read the error message. 'selectors' -> 'selector', and instead of a list it takes a string.

image

kaste:
'cmd' is not in defaults (yet?). Nobody said anything.

braver:
> Haha, you didn't read the error message. 'selectors' -> 'selector', and instead of a list it takes a string.

Oh, jeez, good morning. ☕️ 

> 'cmd' is not in defaults (yet?). Nobody said anything.

Ah, my fault for again not reading everything with both sides of my brain switched on. I think it's a good idea to make `cmd` configurable like that. I don't see any downsides?

braver:
Oh wow, this is seriously neat 🔥

braver:
Ok, let's ⛵️ it and then update the docs and the plugin template.

serhiipalash:
OMG, you are the goods of SublimeText and all other text editors! 👍 
Thank you very much for this awesome enhancement!

Recently we switched to TypeScript, and TSlint is still miles behind ESlint. There are much less rules and there is no error type "warning", all rules are "errors" and highlighted with red color, in TSlint.

I did this
screen shot 2018-03-17 at 1 31 03 am
and now I can lint my files with ESlint, TSlint and even with TypeScript compiler.

screen shot 2018-03-17 at 1 58 45 am
screen shot 2018-03-17 at 1 59 11 am

braver:
🙇🏼‍♂️ Glad to be of service!

PR #1010, opened 2018-02-21
Implement position markers for the panel view

### LINKED ISSUE #806: Add a visual marker in the panel when we are between two errors
Maybe we could mark a region with underlines. Or do something with Phantoms

kaste:
Fixes #806


kaste:
**Note** Icons are copied over from GitGutter. @braver Please ensure we can use them.

braver:
Gitgutter is MIT, I don't see anything special about the icons and it's a bit hard to claim copyright on a handful of red pixels.

I'm getting an error:

```py
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 993, in run_
    return self.run(edit, **args)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/panel_view.py", line 109, in run
    selected_region = sel[0] if sel else None
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime.py", line 649, in __getitem__
    raise IndexError()
IndexError
```

Why did you change `line_padding_bottom`, `line_padding_top` and `margin`? It's super cramped now on my machine, whereas it looked fine before.

kaste:
Hm, unfortunately I get this

image

if I don't adjust the line-padding. 

I don't know where this is coming from. (At least, I don't set line-padding in my own user settings.) 

Do you see this as well? Do I copied the wrong icons? Are they too small?

kaste:
`panel.line_height()` is `20` on my machine, font-size: 13. If I reduce font-size to 10, line-height is 16, and the the gap disappears.

kaste:
Ah, the IndexError is fixed in #1011. I thought the error is unrelated to this PR.

kaste:
Rebased on master, so the IndexError should be fixed.

braver:
> Do you see this as well? Do I copied the wrong icons? Are they too small?

It depends on font and settings. For me the icons actually overlap a bit usually, with an 11px font. So, it's best not to mess with it, it will look slightly different for everyone anyway.

Perhaps it's better with just one icon?

screen shot 2018-02-23 at 21 31 26

Also, as you can see in this screenshot the positioning in the panel is completely broken and useless on my machine (the cursor is at 123, not between 338 and 348). On master as well, so not a regression here... apparently I haven't used that feature for a few weeks and this clearly visualises the problem.

kaste:
It seems that I read the line_height() (20 on my machine) and then I substract 2 on top and bottom to get to 16. So we could apply a computed padding here.

kaste:
Yeah, the position is usually okay, but it is buggy.

braver:
> Yeah, the position is usually okay, but it is buggy.

Do you feel like this is something we can fix, or is it just a black magic box someone else wrote?

> So we could apply a computed padding here

Hmmm, the padding is a readability thing, people should be in control of that for their system and font settings.

kaste:
The bug is just a bug. Had no time to fix it.

The padding patch was actually okay on my system bc it seems that it just adjusted the line-height which is just greater with my settings. On your system this must look over-dense bc you don't see that spacing in the first place.

kaste:
But, maybe you do have a better icon here? if we would have arr-down arr-up, it wouldn't be that important if they really touch each other. We just need two matching icons which look great with and without spacing.

braver:
> The bug is just a bug. Had no time to fix it.

👍 #1014

> On your system this must look over-dense bc you don't see that spacing in the first place.

It depends on what font you're using, fonts have built in spacing too. I guess that's a reason this isn't  the default icon in GitGutter, ST lacks features to make this really nice.

> But, maybe you do have a better icon here?

I think the space between the icons inevitable and it looks bad. I actually kinda like my red bar icon in the screenshot I shared?

kaste:
I like that too, send me that file. (But probably it doesn't look that good on my system bc it will not look centered between the two lines?)

kaste:
(If I look at that picture, this is really as dense as my system *with* the padding patch.)

braver:
The deleted_bottom_arrow icon:

deleted_bottom_arrow

I just lazily made the other one completely transparent.

kaste:
I prefer 

image

over 

image

bc in the latter I clearly see that it is not centered.

braver:
Yes, agreed. 🥇

kaste:
Oh dear, I comletely removed the 'next_result' feature here. I've readded it, now it always highlights the line in the gutter.

image

braver:
Unless I put margin at `0` I get a gap between the gutter and line highlights.

> now it always highlights the line in the gutter.

Crap. That looks lame. Perhaps we need a different approach, e.g. can we put an solid_underline decoration on the line-numbers of the line above the current selection?

braver:
Suggestion: 

```py

OUTPUT_PANEL_SETTINGS = {
    "auto_indent": False,
    "draw_indent_guides": False,
    "draw_white_space": "None",
    "gutter": False,
    "is_widget": True,
    "line_numbers": False,
    "match_brackets": False,
    "rulers": False,
    "scroll_past_end": False,
    "spell_check": False,
    "tab_size": 4,
    "translate_tabs_to_spaces": False,
    "word_wrap": False
}


def draw_position_marker(panel, line, error_under_cursor):
    if error_under_cursor:
        panel.erase_regions('SL.PanelMarker.bottom')
    else:
        bottom_region = sublime.Region(panel.text_point(line - 1, -1), panel.text_point(line - 1, -1) + 12)
        panel.add_regions(
            'SL.PanelMarker.bottom', [bottom_region], scope='source',
            flags=(sublime.DRAW_STIPPLED_UNDERLINE | sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))
```

kaste:
Superseded by #1024

PR #1011, opened 2018-02-21
Fix: Proper guard empty selection on panel updates

kaste:
Bug: The panel does not update e.g. when you quickly switch to a
different application while linter updates are yet coming in


braver:
I can't really reproduce what this guards against, but the change looks good. 🚤 it.

PR #1015, opened 2018-02-23
fix random behaviour in panel position sync

### LINKED ISSUE #1014: syncing cursor position to the panel is buggy
https://github.com/SublimeLinter/SublimeLinter/pull/1010#issuecomment-368131162

braver:
fixes #1014

`get_next_panel_line` tries to find the next nearest error that's visible in the panel, to scroll to that position. However, it breaks out as soon as it finds something, but doesn't sort the errors before doing so. So sometimes it may find the right one by accident but it's super likely to find the wrong one.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -316,12 +316,12 @@ def get_next_panel_line(line, errors):
     In that case return last error's panel line incremented by one, which will
     place panel selection in empty space between buffer sections.
     """
-    for error in errors:
+    for error in sort_errors(errors):
         if error["line"] > line:
             panel_line = error["panel_line"]
             break
-    else:
-        panel_line = errors[-1]["panel_line"] + 1
+        else:
+            panel_line = errors[-1]["panel_line"] + 1
 
     return panel_line, panel_line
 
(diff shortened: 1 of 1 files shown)


braver:
@kaste PTAL 👋

panel_view.py
```diff
@@ -309,19 +309,22 @@ def fill_panel(window, update=False, **panel_filter):
...
+    for error in sort_errors(errors):
+        if error["line"] == line and col < error["start"]:
             panel_line = error["panel_line"]
             break
-    else:
-        panel_line = errors[-1]["panel_line"] + 1
+        elif error["line"] > line:
+            panel_line = error["panel_line"]
+            break
+        else:
```
kaste: This else block belonged to the for loop. I don't understand why it shouldn't anymore.

Other than that, this should do the trick.

braver: Oh, you’re right, I messed that up. Should effectively have the result, but it does belong to the for loop.

braver:
@kaste Think I got it now.

kaste:
The code looks good. But I don't do computer on Sunday. If I should check this out first for functional test, you have to wait until tomorrow.

braver:
> But I don't do computer on Sunday

Sounds like a good policy 👍 . I'd feel better if you do a functional check and there's no hurry.

kaste:
👍

PR #1020, opened 2018-02-26
Log backend exceptions early

### LINKED ISSUE #955: Add debug context for exceptions in linter tasks
When an exception is encountered, it's sometimes not possible to tell from the traceback which linter caused the exception as a lot of the work is happening inside the main Linter class. A try-catch would be needed somewhere to catch a linter's exception and at least print the linter's name responsible for it.

Could use `logging.Logger.exception()` with #909.

Example of a confusing traceback: #937.

kaste:
Fixes #955

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -90,13 +90,18 @@ def get_lint_tasks(linters, view, view_has_changed):
 
 
 def execute_lint_task(linter, code, offset, view_has_changed, settings, task_name):
-    # We 'name' our threads, for logging purposes.
-    threading.current_thread().name = task_name
+    try:
+        # We 'name' our threads, for logging purposes.
+        threading.current_thread().name = task_name
 
-    errors = linter.lint(code, view_has_changed, settings) or []
-    translate_lineno_and_column(errors, offset)
+        errors = linter.lint(code, view_has_changed, settings) or []
+        translate_lineno_and_column(errors, offset)
 
-    return errors
+        return errors
+    except BaseException:
+        # Log while multi-threaded to get a nicer log message
+        logger.exception('Linter crashed.\n\n')
+        raise
 
 
 def translate_lineno_and_column(errors, offset):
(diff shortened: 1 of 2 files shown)


lint/backend.py
```diff
@@ -90,13 +90,18 @@ def get_lint_tasks(linters, view, view_has_changed):
...
-    translate_lineno_and_column(errors, offset)
+        errors = linter.lint(code, view_has_changed, settings) or []
+        translate_lineno_and_column(errors, offset)
 
-    return errors
+        return errors
+    except BaseException:
+        # Log while multi-threaded to get a nicer log message
+        logger.exception('Linter crashed.\n\n')
+        raise
```
FichteFoll: Could also `return None` here and have `await_futures` check for `None` results (or filter out `None` results wherever that is used). That will stop the exception from bubbling through the entire thread and into the future when it has, in fact, already been "handled".

`BaseException` isn't needed here as neither KeyboardInterrupt nor SystemExit will ever be thrown inside an ST plugin context, but w/e.

kaste: `w/e`?

Otherwise a bit too lazy here.

log_handler.py
```diff
@@ -134,6 +134,9 @@ def emit(self, record):
 
 class StatusBarHandler(logging.Handler):
     def emit(self, record):
+        if record.levelno != logging.WARNING:
```
braver: Shouldn’t you use `STATUS_BAR_LEVEL` here?

kaste: That would be clever indeed.

lint/backend.py
FichteFoll: w/e = whatever

log_handler.py
FichteFoll: Do we want to filter everything that's not the specified error or just `exception` output? That uses the ERROR level as well, so the only way to filter would be to check for a log record's exc_info attribute.

kaste: The intention here was to only show warnings in the status bar, for errors we already open the error panel.

PR #1024, opened 2018-02-27
Use underline regions as position markers for the panel view

### LINKED ISSUE #806: Add a visual marker in the panel when we are between two errors

kaste:
That's an alternative take for #1010 

image

Fixes #806

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -394,6 +394,8 @@ def update_panel_selection(active_view, current_pos, **kwargs):
         return
 
     if panel_lines[0] == panel_lines[1]:
+        draw_position_marker(panel, panel_lines[0], is_full_line)
+
         region = get_panel_region(panel_lines[0], panel, is_full_line)
     else:  # multiple panel lines
         is_full_line = True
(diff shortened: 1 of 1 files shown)


braver:
I have only aesthetic remarks here, because it works well and it's elegant.

- I would use a visually lighter style by using STIPPLED_UNDERLINE and no color (at least no signal color, e.g. red)
- `line_start + 4` doesn't reach to the line number on single digit lines, and `line_start + 3` isn't necessary because it will not draw on white space anyway. I think we should try to cover the entire line:col region as that will be the most visually consistent.

-----

Also, know what I just remembered? ST will draw a little bar at the start of the line for `sublime.Region(line_start, line_start)`. I think that's actually what we want here.

kaste:
Okay, `line_start` is the winner here. (STIPPLED_UNDERLINE doesn't work for me anymore, bc my eyes are too old.)

kaste:
Just tell me a good scope, so that users can override this. (As I understand this, we usually take 'foo.bar.sublime_linter'.)

kaste:
I have actually a dimmed red for `markup.deleted`.

braver:
Yeah, the red is fine here. Just use `"region.redish markup.deleted.sublime_linter markup.error.sublime_linter"`, just like we do for errors. Both markup.deleted and markup.error so more color schemes will actually have a color for this.

braver:
Pretty sweet! LGTM 🛥 it!

PR #1029, opened 2018-02-27
fix #877 use find_resources to load icons

### LINKED ISSUE #877: [beta] How to use gutter themes in 4.0rc 9.9.2
How to use gutter themes in beta, @braver?

With these settings:

```json
// SublimeLinter Settings - User
{
  "lint_mode": "load_save",
  "gutter_theme": "Default",
  "styles": [
      {
          "mark_style": "solid_underline",
          "priority": 1,
          "scope": "region.yellowish markup.changed.sublime_linter markup.warning.sublime_linter",
          "icon": "cog",
          "types": [
              "warning"
          ]
      },
      {
          "mark_style": "solid_underline",
          "priority": 1,
          "scope": "region.redish markup.deleted.sublime_linter markup.error.sublime_linter",
          "icon": "cog",
          "types": [
              "error"
          ]
      }
  ]
}
```
I have this error:

screen shot 2018-01-23 at 19 47 33

In fact, I want to use the `Hands` theme, how can I do this with beta?

braver:
We reconstruct paths to the icons, which doesn't work if SL is installed somewhere else (which happens when installed in a packaged form via package control). ST has the find_resource api to find where your files are without you having to know about the internals of where stuff is installed (e.g. the "Packages" folder, or the "Installed Packages" one, etc). 

This strips out loading up the GUTTER_ICONS with all paths to all icons, instead just using the icon and theme name to, just-in-time, get the path to an icon via find_resources.

I don't really know how to test this without releasing and then installing via package control.

Also, this stops it from throwing errors if the theme-icon combination doesn't result in a resolvable path to an icon. It will now silently refrain from showing icons until you fix the configuration.


kaste:
That's not how I understand the Sublime API. Loosing the cache but calling `find_resource` is probably also not what we want. After all, the icon path we can pass to `add_regions` must be something like e.g. `Packages\SublimeLinter\gutter-themes\Default\error.png`.

So to get an icon path, I think the function is given `(theme, icon)`

- Given a theme via settings, expand it 'Packages\SublimeLinter\gutter-themes\\{theme}' if it is not a third party theme (maybe if it does not contain an `os.path.sep`).
- Add '.png' to icon name if not there
- Further expand `{rel_theme_path_from_step_1}\{icon_name_from_step_2}`

lint/style.py
```diff
@@ -46,16 +43,6 @@ def load_gutter_icons():
         colorize = True
 
     new_gutter_dict["colorize"] = colorize
-    dir_path, _ = os.path.split(theme_file)
-    pck_path = sublime.packages_path().split("/Packages")[0]
```
kaste: To make the bug clearer. This is broken, the original author meant probably `os.path.dirname`.

lint/style.py
```diff
@@ -46,16 +43,6 @@ def load_gutter_icons():
...
-    abs_dir = os.path.join(pck_path, dir_path)
-
-    png_files = glob(os.path.join(abs_dir, "*.png"))
```
kaste: Obviously, we can't `glob` through a zip file.

braver: Indeed, we can't pre-emptively loop over the file system to find the available icons for a given theme.

braver:
I was operating under the assumption that we shouldn't construct "Packages/SublimeLinter" paths at all, but that's probably incorrect. 

So, we can't loop over the file system to cache paths to icons, but we can fill the cache based on user settings. I'll try again.

kaste:
We probably don't need the old cache, if we construct the icon path just via string manipulation. We just need to store the `theme_path`, if we want to allow third party themes.

Otherwise, it's basically just `'Packages\SublimeLinter\gutter-themes\{theme}\{icon}.png'.format(...)` as far as I understand Sublime here. For something like that, we could probably use a `lru_cache`.

braver:
Let's not cache if we're sure we need to. Even with the solution using find_resources a lot I think it was super fast. Usually API features like that are super optimised. 

My intention right now is to keep the design mostly as is, but call `load_gutter_icons` every time the settings are changed and then construct paths to the selected icons. I.e. the minimum to fix the bug. You want to do it differently?

lint/style.py
```diff
@@ -87,16 +73,14 @@ def wrapper(*args):
...
                 if res in ("circle", "dot", "bookmark", "none"):  # Sublime Text has some default icons
                     return res
-                elif res != os.path.basename(res):
-                    return res
                 else:
-                    icon_path = GUTTER_ICONS["icons"].get(res)
-                    if icon_path:
-                        return icon_path
-                return GUTTER_ICONS["icons"][error_type]
+                    theme = persist.settings.get('gutter_theme')
```
kaste: I thought (in general keeping this PR) but after here you just 

```
	return 'Packages\SublimeLinter\gutter-themes\{}\{}.png'.format(theme, res)
```

lint/style.py
```diff
@@ -87,16 +73,14 @@ def wrapper(*args):
             if key != "icon":
                 return res
             else:
-                # returning paths
                 if res in ("circle", "dot", "bookmark", "none"):  # Sublime Text has some default icons
                     return res
-                elif res != os.path.basename(res):
-                    return res
```
kaste: You *could* leave this branch. I never used theming, but this one means users can set a full path to a 'png' and we will just use that?

lint/style.py
braver: Oh, that's also good. If the specified icon doesn't exist it will fail like this:

screen shot 2018-02-28 at 14 23 16

```
Unable to decode Packages/SublimeLinter/gutter-themes/Hands/pointer-outline.png
```

Which actually is a really nice way to fail IMO.

braver: 👍

braver:
@kaste Like this then?

lint/style.py
kaste: I actually used this for testing. Bc I never used themes before, I thought I had to do it like that. 🍭 🤣

kaste:
👍 That's at least how I understand the Sublime API. Should be really simple.

braver:
Ok, let's try this then. @kaste wanna merge?

kaste:
Hm, doesn't work for the Blueberry and Knob theme. Bc of the nested directory structure. Also you're not much into refactoring, are you?

`GUTTER_ICONS` should become `CURRENT_THEME = {path: '', colorize: true|false}`, unless you can come up with a directory/structure for the themes so we don't have to compute a theme_base_path. AFAIU you set `Blueberry/Blueberry - round.gutter-theme` but the icons are at `Blueberry/round`.

braver:
Oh good catch. I’d rather standardize what we expect wrt the file structure in a gutter theme.

I’ll look into refactoring this some more.

Anything you’d like the get done before release? We could publish 4.0 on Friday...

kaste:
ATM this one and maybe #1004 bc ui experiments without feedback...

braver:
Oh yes, that one too. We can merge that in its current state IMO

braver:
Ok, so themes now all have the same layout: a theme file next to some icons. This way we don't have to be smart about building paths and just use the `'Packages/SublimeLinter/gutter-themes/{}/{}.png'` template.

This time I did do some refactoring, because GUTTER_ICONS is no more. All that's left to do at start and on settings changes, is to check if the icons should be colorized. 

@kaste PTAL

lint/style.py
```diff
@@ -38,27 +30,9 @@ def load_gutter_icons():
...
-        png_file = os.path.basename(png)
-        name, ext = os.path.splitext(png_file)
-
-        new_gutter_dict["icons"][name] = os.path.join(dir_path, png_file)
-
-    global GUTTER_ICONS
-    GUTTER_ICONS = new_gutter_dict
+        if opts:
+            global COLORIZE
+            COLORIZE = opts.get("colorize", True)
```
kaste: If you change from a 'colorize = False' Theme to `'none'`, it will stay ... 'False' 😄

braver: Oh duh, trying to be too efficient here

braver:
@kaste I do hope I got it now...

PR #1037, opened 2018-03-01
Introduce 'demote_scope' for the highlights

kaste:


braver:
I'm generally 👍 🎉 🕺 on this. LGTM, 🚤  it. We can discuss tweaks to this in #1038.

braver:
@kaste I want to send you your share of the donations that we received these past few months. Where can I reach you personally, or can you send me your PayPal address?

PR #1042, opened 2018-03-02
Begin building a changelog

kaste:
@braver Are you generally okay with a CHANGELOG? Usually or I would say, we call it x.x.next and write stuff in there. Before we release we can rename the file. Or we just point to x.x.next in messages.json.


braver:
Yes, I definitely want a changelog. sl3 to sl4 was to big to make sense and I didn't want to have to bother with it for the beta, but now we should start one. 👍

PR #1055, opened 2018-03-02
always lint files once when they're opened

### LINKED ISSUE #1054: [Version 4] with the default lint mode, "background", the linter is not run when a file is opened
Using ST 3 dev build 3157, SublimeLinter 4.0.1, SublimeLinter-eslint 4.0.0, ESlint 4.18.1 in win10x64

I quote from the SublimeLinter.sublime-settings:

```javascript
    // Lint Modes determine when the linter is run
    // background: asynchronously on every change
    // load_save: when a file is opened and every time it's saved
    // manual: only when calling the Lint This View command
    "lint_mode": "background",
```

i.e. i. SL4 with the default lint mode, "background", whenever you open a file, the linter is not run. It's only run if you modify the file, i.e. type or delete something on it.
But, if you don't want or just forget to modify it, then it's not linted, giving you the wrong impression that the file is fine.

_For reference, in SL3, with the "background" lint mode, the file was linted on load._

What I'd prefer is having a 4th mode, something like" load_and_background",
i.e. lint the file on load, and then lint it again asynchronously on every change.

<br/>

So, could you please consider adding such a lint mode? That's my suggestion.

braver:
fix #1054 when you open a file it isn't linted until you start editing. I think we can just lint right then, regardless of lint_mode.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -124,6 +124,10 @@ class BackendController(sublime_plugin.EventListener):
         if has_syntax_changed(view):
             hit(view)
 
+    def on_load_async(self, view):
+        if util.is_lintable(view):
+            hit(view)
+
     def on_post_save_async(self, view):
         if persist.settings.get('lint_mode') == 'manual':
             return


braver:
@kaste what do you think about this? I think linting a file right away is actually pretty important.

kaste:
That's just a bug, and I'll fix it not on the release day bc it's just too distracting right now. 

I knew the bug before, it came to me while walking with my kids, I think. I think I have a cache in on_activated but the cache returns False on first call, but it must return True or vice versa.

braver:
> it came to me while walking with my kids, I think

🤣 

I'll wait for your fix then.

PR #1056, opened 2018-03-02
Must assume 'syntax-changed' for new views

### LINKED ISSUE #1054: [Version 4] with the default lint mode, "background", the linter is not run when a file is opened

kaste:
Fixes #1054

Replaces #1055

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -162,7 +162,7 @@ def has_syntax_changed(view):
     try:
         old_value = buffer_syntaxes[bid]
     except KeyError:
-        return False
+        return True
     else:
         return old_value != current_syntax
     finally:


kaste:
Edited on GitHub so others must test. That is just the bug I saw while playing/walking with my kids. I think this is causing #1054.

braver:
Looks good 👍

PR #1079, opened 2018-03-04
Fix broken code blocks on linter settings page.

alimony:
These code blocks are not formatted correctly because of a missing space in the RST markup.


braver:
Thanks!

PR #1081, opened 2018-03-04
adds settings for statusbar counters/messages

### LINKED ISSUE #1090: Remove element from status bar
I could be wrong but I think sublimelinter 4 has added this section to the status bar
screen shot 2018-03-05 at 11 30 40
Is there a setting I can use to delete or modify this section? 

I've tried navigating to here: http://www.sublimelinter.com/en/latest/mark_styles.html#status-bar-info via the documentation but there's a 404 there

braver:
From a forum post: https://forum.sublimetext.com/t/sublimelinter-4-beta/34022/30

This adds options to toggle the counters and/or the messages in the status bar. 

Fixes #1090

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -88,11 +88,12 @@
     // Highlight problems in the minimap.
     "show_marks_in_minimap": true,
 
-    // Show the number of errors and warnings in the current view
+    // Show the number of errors and warnings in the current view.
     "statusbar.counters": true,
 
-    // Show the number of errors and warnings in the current view
-    "statusbar.counters.template": "W: {} E: {}",
+    // How to display the counters.
+    // {warning} and {error} will be replaced with the actual numbers.
+    "statusbar.counters.template": "W: {warning} E: {error}",
 
     // Show the messages for problems at your cursor position
     "statusbar.messages": true,
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -88,6 +88,12 @@
     // Highlight problems in the minimap.
     "show_marks_in_minimap": true,
 
+    // Show the number of errors and warnings in the current view
+    "statusbar.counters": true,
+
+    // Show the messages for problems at your cursor position
+    "statusbar.messages": true,
+
     // Global styles for all linters.
     // - mark_style options:
     //   "fill", "outline",
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -91,12 +91,12 @@
     // Display counters in the status bar.
     // {warning} and {error} will be replaced with the actual numbers.
     // Set to "" to display nothing
-    "statusbar.counters": "W: {warning} E: {error}",
+    "statusbar.counters_template": "W: {warning} E: {error}",
 
     // Show the messages for problems at your cursor position
     // {message} will be replaced by the actual messages.
     // Set to "" to display nothing
-    "statusbar.messages": "{messages}",
+    "statusbar.messages_template": "{messages}",
 
     // Global styles for all linters.
     // - mark_style options:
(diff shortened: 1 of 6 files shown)


SublimeLinter.sublime-settings
```diff
@@ -88,6 +88,12 @@
     // Highlight problems in the minimap.
     "show_marks_in_minimap": true,
 
+    // Show the number of errors and warnings in the current view
+    "statusbar.counters": true,
+
+    // Show the messages for problems at your cursor position
+    "statusbar.messages": true,
```
FichteFoll: Shouldn't these rather be more explicit and say "show_counters" and "show_messages"?

status_bar_view.py
```diff
@@ -70,15 +70,24 @@ def on_selection_modified_async(self, view):
 
 
 def draw(active_view, we_count, current_pos, errors_per_line, **kwargs):
-    if not we_count:
+    counters = persist.settings.get('statusbar.counters')
+    messages = persist.settings.get('statusbar.messages')
+    msgs = messages_under_cursor(errors_per_line, current_pos)
```
FichteFoll: These two variable names (messages, msgs) are confusing. If "show_" is added to the settings, it should be for the variables as well.

SublimeLinter.sublime-settings
braver: I get this point for the variables, but showing stuff is all the statusbar does. Also, the comment is there.

kaste:
If we introduce options here, I want a template string for the counter. We had one PR here whose purpose was to to do some unicode stuff here we couldn't add as a default bc it would break depending on the font. So this should become a (name bikeshedding allowed) `counter: 'W: {warnings} E: {errors}'.

Then we should use different keys for 'messages' and 'count'. 'STATUS_COUNTER_KEY', 'STATUS_MSG_KEY', this simplifies the logic.

braver:
Ok sounds good. I'll see if I can get something going in that direction.

braver:
@kaste PTAL. I considered splitting the draw function, but didn't feel like it had big benefits.

kaste:
We don't need boolean switches here, I think it is okay to take the empty string here. If it works formatting the empty string should return the empty string, so this makes the implementation also trivial.

I thought it is better to do the messages also with a template so people can choose a prefix they like. And it's trivial to implement.

kaste:
Ah, and the constants should be removed from const. You did that before but we obviously lost this commit.

braver:
Okido!

braver:
@kaste ok, looks like that works well.

messages/4.0.next.txt
```diff
@@ -0,0 +1,7 @@
+SublimeLinter 4.0.next
+======================
+
+Introduces settings to customize the information in the status bar
+
+- "statusbar.counters" toggles the W:/E: counters in the status bar
+- "statusbar.messages" toggles the messages for the problem at your cursor
```
FichteFoll: This needs to be updated.

SublimeLinter.sublime-settings
```diff
@@ -88,6 +88,16 @@
     // Highlight problems in the minimap.
     "show_marks_in_minimap": true,
 
+    // Display counters in the status bar.
+    // {warning} and {error} will be replaced with the actual numbers.
+    // Set to "" to display nothing
+    "statusbar.counters": "W: {warning} E: {error}",
```
FichteFoll: I prefer `counter_template`. At least something that contains "template".
Same for the messages setting.

I mean, you already use variables named like this, so why not expose this more descriptive name to the user?

braver: I opted for brevity, but you're right that's better.

kaste:
Good enough

braver:
Such enthusiasm 😉

kaste:
LOL

PR #1109, opened 2018-03-06
Change "overwrite" to "override".

jtheletter:
"Overwrite" indicates replacement of data. 
"Override" indicates taking of precedence.


braver:
Makes sense. Thanks!

PR #1111, opened 2018-03-06
Reword English grammar.

jtheletter:


braver:
Goood catch!

FichteFoll:
That setting isn't used anywhere.

braver:
Dang, that's an even better catch!

FichteFoll:
(I found out yesterday while investigating the macOS path issues.)

PR #1129, opened 2018-03-06
Enforce 'type' for line and col

kaste:
Issue https://github.com/SublimeLinter/SublimeLinter-javac/issues/16


braver:
Ok, looks good 👍

PR #1130, opened 2018-03-06
Fix: Check for None as long as `0` is among the expected values

### LINKED ISSUE #1115: Linter does not show phantom if the cursor is not on the current line
back3

1. Debug https://gist.github.com/evandrocoan/263843916891ce188dde4c22883c34c4

kaste:
Fixes #1115

### DIFF
--- a/tooltips_view.py
+++ b/tooltips_view.py
@@ -58,10 +58,10 @@ def open_tooltip(active_view, point=None, line_report=False):
     if active_view.is_popup_visible():
         return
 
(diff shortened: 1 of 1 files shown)


kaste:
You see I just need this errors/wanrings in my face

braver:
Lol, yeah we need that panel popup thing back!

This one works for me, looks like a good solution. Ship it!

kaste:
I have my own pop up currently, fixed the issue but forgot to save. 🤣

PR #1134, opened 2018-03-06
Correct verb tense ("is show" -> "is shown").

jtheletter:


braver:
Awesome. Are you an editor by trade perhaps? 😁

PR #1136, opened 2018-03-06
Fix non-colorized icon colors for light color schemes

braver:
Ok, the problem is this:

When a gutter theme sets "colorize" to false, we set the `" "` scope on the region. ST *always* colorizes the icons, so all this does is make it colorize with the default foreground color. In dark color schemes that color is probably white, which effectively nullifies the colorization. In light color schemes however it's probably black which really f****'s up the icons. You don't want the default foreground color, you want the background color here.

There is no scope we can use that says "use the background color".

bf17720 introduces a scope nobody implemented in their color scheme. This scope can be used to patch a color scheme by adding this:

```xml
<dict>
	<key>name</key>
	<string>SublimeLinter4</string>
	<key>scope</key>
	<string>sublime_linter_non_colored_scope</string>
	<key>settings</key>
	<dict>
		<key>foreground</key>
		<string>#FFFFFF</string>
	</dict>
</dict>
```

This isn't ideal, but at least its possible. The scope `" "` you can't do anything with.

I don't know of other solutions that don't include generating color schemes and we're definitely not going there.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -152,5 +152,11 @@
         "php": "html",
         "python django": "python",
         "pythonimproved": "python"
-    }
+    },
+
+    // Tell us what scope is a white color in your color scheme.
+    // This is used to render pre-colored icons (e.g. from the Hands gutter theme)
+    // in their original color.
+    // For most dark color schemes, " " should be correct.
+    "white_scope": " "
 }
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -154,9 +154,11 @@
         "pythonimproved": "python"
     },
 
-    // Tell us what scope is a white color in your color scheme.
+    // Tells us what scope is a white color in your color scheme.
     // This is used to render pre-colored icons (e.g. from the Hands gutter theme)
     // in their original color.
-    // For most dark color schemes, " " should be correct.
-    "white_scope": " "
+    // For most dark color schemes, you probably don't need to change this.
+    // You can also put the "SublimeLinterWhiteScope" scope in your color scheme
+    // and set its foreground to white (#FFFFFF).
+    "white_scope": "SublimeLinterWhiteScope"
 }
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -158,7 +158,7 @@
     // This is used to render pre-colored icons (e.g. from the Hands gutter theme)
     // in their original color.
     // For most dark color schemes, you probably don't need to change this.
-    // You can also put the "SublimeLinterWhiteScope" scope in your color scheme
+    // You can also put the "sublime_linter.white" scope in your color scheme
     // and set its foreground to white (#FFFFFF).
-    "white_scope": "SublimeLinterWhiteScope"
+    "white_scope": "sublime_linter.white"
 }
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -358,7 +358,7 @@ def get_icon_scope(icon, error):
     if style_stores.COLORIZE:
         return get_scope(**error)
     else:
-        return " "  # set scope to non-existent one
+        return persist.settings.get('white_scope')
 
 
 def draw(view, linter_name, highlight_regions, gutter_regions,
--- a/resources/settings-schema.json
+++ b/resources/settings-schema.json
@@ -139,7 +139,9 @@
                 "additionalProperties":false
             }
         },
-        "white_scope": "string"
+        "white_scope": {
+            "type": "string"
+        }
     },
     "additionalProperties": false
 }
--- a/resources/settings-schema.json
+++ b/resources/settings-schema.json
@@ -138,7 +138,8 @@
                 ],
                 "additionalProperties":false
             }
-        }
+        },
+        "white_scope": "string"
     },
     "additionalProperties": false
 }


markusdd:
That looks like a hookup we can work with for now. Would need to try.
Probably wee need to talk to Sublime Devs directly. I do not know why you would put a shade over a simple png icon. it's a picture...just display it already...

braver:
> Probably wee need to talk to Sublime Devs directly.

I'll see if I can get more info in the Discord chat.

braver:
Ok, we talked about it in the chat. It's a non-fixable problem from our perspective, at least without modifying color schemes (which we're obviously not going to do). There is no way to reliably get a white color from all color schemes. There is no use in introducing a scope here because no-one is going to set that in their color scheme. What we can do is make the scope we set configurable so users can extend their color scheme by hand and tell us what scope they used.

braver:
@kaste the short version:

If people use the Hands, Danish Royalty, etc gutter theme with a bright color scheme that doesn't work. 

- We assumed the scope `" "` returns a white color, this is not correct for bright color schemes
- We cannot know or find out which scope is white and we shouldn't assume
- So the proposed solution is to ask the user what scope is white

This will probably result in questions like "what is a scope" and "how do I find out what is white", but this can be documented.

FichteFoll:
Does anyone actually *want* to use a non-colorized icon? All of the default gutter themes have a red error and a yellow warning icon. If those were simply monochromized (with shades in tact for perspective and stuff), they would just behave like the default icons, assuming that a color is defined for the scope scope used (`region.redish markup.deleted markup.error` et al).

braver:
I could set "SublimeLinterWhiteScope" as the default value for this setting so that approach would also work. But then you don't have to modify the color scheme if it already has a scope that's white, you can change the setting.

braver:
> If those were simply monochromized (with shades in tact for perspective and stuff)

That doesn't work because they have white parts (e.g. the hands) in addition to the coloured parts (the sleeve), and there is no way to keep the white parts white.

FichteFoll:
True, white would need to be converted to black since black isn't tinted and that probably wouldn't look too nice for some icons.

kaste:
I have made my own theme, and either I had to look up a value in my theme file and copy that to the SL settings, or I just copy the value from the readme to my theme file. 

For downloaded themes, I would think they can/will just add this in the long or shorter run. But I can avoid the settings lookup here.

braver:
@kaste You can now put `sublime_linter.white` in your color scheme and that'll work. Best of both worlds.

Note to self: squash this.

markusdd:
> Ok, we talked about it in the chat. It's a non-fixable problem from our perspective, at least without modifying color schemes (which we're obviously not going to do). 

So what's the actual way forward then? The configurable scope is a workaround, but not an actual fix. Did they comment on how they want to address this in the future? It cannot be a long term solution if you want to switch the theme that you have to patch it manually all the time. I know that this is not in your hands as plugin developers, but I think we'll need to push hard that they figure this out cleanly by providing an API which allows to display an icon independent of the color scheme settings.

And even this sublime_linter.white scope is not enough, because even if you use colorable icons you still  depend on the fact that your theme has a yellow color and red color available to use for warnings and errors. If this is not the case (like Eiffel, it has red but no yellow) you're stuck again.

So actually we need sublime_linter.white sublime_linter.warning and sublime_linter.error as scopes which can be put into a theme file so that you can set the proper colors there without overriding some code context you don't actually want to color yellow or red.

braver:
Well there’s two aspects to this.

One: our gutter themes are the only thing in the entire ST universe that puts icons in the gutter and doesn’t want to color them. And even those themes aren’t used by a lot of people. So getting the ST theme to prioritize this isn’t going to happen.

Two: we cannot rely on any scopes being available in color scheme, but there is a minimal set any reasonable color scheme should provide. Other packages also need colors for concepts like “removed” or “error”, so there is traction to get color schemes to have those. Additionally, future versions of ST will give use reliable scopes for colors like red and yellow, so that problem will disappear. The set of colors may not include white, that’s not final yet, we’ll see.

The API gives us lemons, we make lemonade. Honestly, there are far more important improvements to be made to the API and they are doing a good job of that.

Also honestly, if I knew about this problem a week ago, we wouldn’t even have had gutter themes in SL4. Our Default theme has has more options than you can swing a stick at and should be enough for anyone. Think of the pre-colored themes as a legacy fan-service thing.

markusdd:
> Additionally, future versions of ST will give use reliable scopes for colors like red and yellow, so that problem will disappear. The set of colors may not include white, that’s not final yet, we’ll see.

But if these scopes will exist 'automatically' they would solve the problem I think, at least for the colorized icons things will become much more convenient. It should not be a huge challenge for them to give us this white scope as well for the png icons.

As you have seen from the other thread we had here also other packages have challenges with this strange tinting behavior. And I think SublimeLinter should've some weight in the discussion, it's a very popular plugin...rightfully so.

But anyway, thanks a lot for looking so far into this, very much appreciated!

And if we need some 'standard' extra scopes in themes that should be easy to do, I'm happy to patch the legacy ones if needed, they are still very popular amongst the community and backporting stuff like this is not much work.

braver:
> And if we need some 'standard' extra scopes in themes that should be easy to do, I'm happy to patch the legacy ones if needed, they are still very popular amongst the community and backporting stuff like this is not much work.

Well that wouldn't be a very bad idea. Most of the color schemes are on GitHub somewhere, the old default ones are too. So some pull requests to add `sublime_linter.white` and `markup.error|warning` isn't a bad idea. It'd take some work and perhaps only worth it if people complain. ST itself is making moves to improve the situation here, so perhaps a white scope is coming too. They don't publish timelines though, so it's kinda hard to get a feel for when such things might land in stable.

kaste:
If you feel the need for a setting then be it so. But, we can't introduce options and settings every day. I don't see it here bc it is like a one time init/setup and then I can modify my theme to get the same. For me, it's more like I setup my theme/colors to work with SublimeLinter (which I've done before, and for other plugins as well) and not I setup SublimeLinter to work with my theme. Either the theme provides the key/scope I need or I add it, I don't think the theme's readme will suggest "hey, use this key and put it in SL's settings bc then they could have named it anyway.

The PRO side of a setting is, this is something people actually read, so they see the name and comment and can self-repair if they know SublimeText. This is something I would really like about it.

braver:
Totally feel te same. I think the setting has benefits, but I don't feel good about any of this. I'm just going to drop a fixed scope into it (`region.whitish` in hopes of future ST support for it) and call it a day.

markusdd:
I get all the concerns, and I totally understand why you changed the way SublimeLinter works, because the theme-patching and overriding was not pretty. (So I agree with all the design decisions you made)

Just to give you my perpective: I was one of the people spreading the use of SublimeLinter and Sublime Text as a whole amongst my colleagues to increase efficiency and code quality and to finally get rid of the ancient editors like Vim,Emacs or even worse gedit, kwrite etc. . Especially the latter do not provide any advanced editing features and it was a pain to watch some peoples inefficient editing style...

Now with the last few updates everything stopped working and you can imagine that most found it not very funny in a productive environment where they do not want to deal with their tool setup but just do their normal work.

I will try to rectify most of this with our internal Sublime Package where I provide patched themes. Ultimately things are likely to get better with newer ST releases when we have the new default scopes,
it was just a pain to find out what is causing the issue.

braver:
@markusdd I would have hoped to catch people like you with the pre-releases to help you prepare for the update. The most impactful changes have been out there for over a month. So keep an eye out for that. I hope the handling of this issue shows a willingness on our end to figure out something that works.

alkuzad:
For others who will struggle with this and will find this thread:

```xml
<dict>
    <key>name</key>
    <string>White for ST3 for non-colorized scheme</string>
    <key>scope</key>
    <string>region.whitish</string>
    <key>settings</key>
    <dict>
        <key>foreground</key>
        <string>#FFFFFF</string>
    </dict>
</dict>
```

PR #1162, opened 2018-03-07
Do not reload but ask for a restart.

kaste:
This would be like an A/B test in the blind. We actually don't have enough information, we never see full logs for the crashes that go away with a restart. 

The command is still very useful for development bc you (definitely I) can tweak something, add a print statement, reload and done.

Actually, we get scary reports here, and the people want something we could only achieve if we send tracebacks and metrics to something like Sentry. If we had that I could just fix the stuff. 😞

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -49,11 +49,11 @@ def plugin_loaded():
     try:
         from package_control import events
         if events.install('SublimeLinter'):
-            reloader.reload_linter_plugins()
-            return
+            util.message(
+                'SublimeLinter has been installed. Please restart Sublime Text.')
         elif events.post_upgrade('SublimeLinter'):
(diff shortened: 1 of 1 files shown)


braver:
Ok, yeah I agree with this approach. It pretty much impossible to recreate what happens during an upgrade and to figure out all the things that can go wrong. Just asking for a restart is probably safer. And easier than trying  to deal with all the weird reports weird getting,

braver:
Shall we merge this and #1136 tomorrow morning and push out a release?

kaste:
I don't have a feeling about the priority of #1136. But the other two PR I made today were actually crashes. And I'll merge this one.

braver:
It has no priority but it’s also super minor and I want to be done with it.

PR #1169, opened 2018-03-08
View cleaning

### LINKED ISSUE #1148: remove our views on package unload
When you disable SublimeLinter, the highlights and the statusbar, etc. remain. We should clean up right before the package unloads.

braver:
Fix #1148 by cleaning up stuff on unload. panel_view already did, and I decided tooltips_view didn't need to (you can't really do anything with ST without closing the tooltips anyway).

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -62,6 +62,9 @@ def plugin_loaded():
 
 def plugin_unloaded():
     events.off(on_lint_result)
+    for window in sublime.windows():
+        for view in window.views():
+            undraw(view)
 
 
 @events.on(events.LINT_RESULT)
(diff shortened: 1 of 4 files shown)


highlight_view.py
```diff
@@ -361,6 +364,14 @@ def get_icon_scope(icon, error):
         return "region.whitish"  # hopefully a white color
 
 
+def undraw(view):
+    current_region_keys = get_regions_keys(view)
+    current_linter_keys = {key for key in current_region_keys
+                           if key.startswith('SL.')}
+    for key in current_linter_keys:
+        view.erase_regions(key)
+
```
kaste: All regions start with 'SL.' so you don't need to filter here. So, basically it is ok to do `for key in current_region_keys`. 

But afterwards you must call `remember_region_keys(view, set())` to clean the store!

braver: Excellent. Thanks!

PR #1172, opened 2018-03-08
show_panel_on_save if errors in view or window + make it lazy

### LINKED ISSUE #1084: Show the output panel on-save if there are errors.
Since the upgrade to 4, there isn't a tooltip any more that screams in my face that my code contains errors when I save a file that contains errors. This used to be the most useful feature to me in Sublimelinter. Is there any way to get it back? Maybe I've just not set the right settings?

### LINKED ISSUE #1163: Do not create and update the panel in the background
- It's sane, c'mon haha what were we smoking? 😒 
- Most of the time probably the panel-syntax would be unused 🤡 
- If the user re-opens it via ctrl-k a, it may gain enough focus, so that `next_result` works again. 🤘

braver:
Fixes #1084 by introducing a setting to show the output panel if there are problems:

"show_panel_on_save"

- Set to "window" to check if the window has problems.
- Set to "view" to only check the current file/buffer.
- Set to "never" to disable this feature. This is the default.

Additional changes to fix #1163 

- [x] panel is not created until the user does so manually
- [x] the panel and its state aren't updated if the panel doesn't exist
- [ ] make next_result work as soon as you open the panel
- [ ] clean up the panel if it's no longer in use

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -72,6 +72,27 @@ class UpdateState(sublime_plugin.EventListener):
         if window:
             sublime.set_timeout_async(lambda: fill_panel(window, update=True))
 
+    def on_post_save_async(self, view):
+        if persist.settings.get('show_panel_on_save') == 'never':
+            return
+
+        window = view.window()
+        panel = ensure_panel(window)
+        # If we're here and the user actually closed the window in the meantime,
+        # we cannot create a panel anymore, and just pass.
+        if not panel:
+            return
+
+        errors_by_bid = get_window_errors(window, persist.errors)
+
+        if persist.settings.get('show_panel_on_save') == 'window' and errors_by_bid:
+            window.run_command('show_panel', {'panel': 'output.' + PANEL_NAME})
+        else:
+            for bid in errors_by_bid:
+                if bid is view.buffer_id():
+                    window.run_command('show_panel', {'panel': 'output.' + PANEL_NAME})
+                    return
+
 
 class SublimeLinterPanelToggleCommand(sublime_plugin.WindowCommand):
     def run(self, force_show=False, **kwargs):
@@ -208,7 +229,7 @@ def filter_and_sort(buf_errors, panel_filter):
     return sort_errors(buf_errors)
 
 
-def get_window_errors(window, all_errors, panel_filter):
+def get_window_errors(window, all_errors, panel_filter=None):
     bid_error_pairs = (
         (bid, all_errors[bid]) for bid in buffer_ids_per_window(window)
     )
--- a/panel_view.py
+++ b/panel_view.py
@@ -109,12 +109,12 @@ class SublimeLinterPanelToggleCommand(sublime_plugin.WindowCommand):
         is_active_panel = (active_panel == "output." + PANEL_NAME)
 
         if is_active_panel and not force_show:
-            self.toggle_panel(PANEL_NAME, show=False)
+            self.toggle_panel(show=False)
         else:
             fill_panel(self.window, **kwargs)
-            self.toggle_panel(PANEL_NAME)
+            self.toggle_panel()
 
-    def toggle_panel(self, name, show=True):
+    def toggle_panel(self, show=True):
         """
         Change visibility of panel with given name.
 
@@ -126,7 +126,7 @@ class SublimeLinterPanelToggleCommand(sublime_plugin.WindowCommand):
         else:
             cmd = "hide_panel"
 
-        self.window.run_command(cmd, {"panel": "output." + name or ""})
+        self.window.run_command(cmd, {"panel": "output." + PANEL_NAME})
 
 
 class SublimeLinterUpdatePanelCommand(sublime_plugin.TextCommand):
--- a/panel_view.py
+++ b/panel_view.py
@@ -313,42 +313,6 @@ def fill_panel(window):
         update_panel_selection(**State)
 
 
-# logic for updating panel selection
-
-def get_next_panel_line(line, col, errors):
-    """Get panel line for next error in buffer.
-
-    If not found this means the current line is past last error's buffer line.
-    In that case return last error's panel line incremented by one, which will
-    place panel selection in empty space between buffer sections.
-    """
-    for error in sort_errors(errors):
-        if error["line"] == line and error["start"] > col:
-            panel_line = error["panel_line"]
-            break
-        elif error["line"] > line:
-            panel_line = error["panel_line"]
-            break
-    else:
-        panel_line = errors[-1]["panel_line"] + 1
-
-    return panel_line, panel_line
-
-
-def update_selection(panel, region=None):
-    selection = panel.sel()
-    selection.clear()
-    if region is not None:
-        selection.add(region)
-
-
-def get_panel_region(row, panel, is_full_line=False):
-    region = sublime.Region(panel.text_point(row, -1))
-    if is_full_line:
-        region = panel.line(region)
-    return region
-
-
 def update_panel_selection(active_view, current_pos, **kwargs):
     """Alter panel selection according to errors belonging to current position.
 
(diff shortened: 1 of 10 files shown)


derekjones:
Thanks for working on this so quickly! Works great for me, better than the SL3 way, to boot.

derekjones:
Clarification: with: `"show_panel_on_save": "view"` the linter is still linting and showing errors from all open files in my project. Shouldn't that only be happening with `window`?

braver:
No, it doesn’t filter the panel. It could perhaps, I’d have to look at that, maybe that makes sense. The panel always shows stuff for the entire window, so it might not be super clear that in this case it hides those other problems. In any case, it should be centered on your current file now.

panel_view.py
```diff
@@ -72,6 +72,27 @@ def on_pre_close(self, view):
...
+            return
+
+        errors_by_bid = get_window_errors(window, persist.errors)
+
+        if persist.settings.get('show_panel_on_save') == 'window' and errors_by_bid:
+            window.run_command('show_panel', {'panel': 'output.' + PANEL_NAME})
+        else:
+            for bid in errors_by_bid:
+                if bid is view.buffer_id():
+                    window.run_command('show_panel', {'panel': 'output.' + PANEL_NAME})
```
kaste: Now, iirc for `errors_by_bid` is a dict so you can basically test `view.buffer_id() in errors_by_dict`.

But note that as soon as we have a lazy panel_view it is not enough to just show it. So firing our own command `sublime_linter_panel_toggle` with `force_show=True` is probably the right thing to do.

kaste:
Aside: Reviewing this revealed that `panel_filter` is actually not implemented. I did not implement nor feature requested it. Maybe it was only implemented bc "yes, we can!". Apparently, nobody uses this feature.

panel_view.py
braver: Calling our own command is not a bad idea anyway, so let's do that.

braver:
The idea behind the panel filtering was that you could create a panel that would tell you only flake8 errors, or only annotations, or only D100 errors, or to remove the warnings from the panel. The filtering is done by passing args when calling via keybinding, but users could add their own variations to the command palette.

I didn't test it for a while (I don't need or want the feature personally), but @pykong added that in a very early phase of SL4. I notice now that it is only respected when the panel is first created, not when it is updated later on. 

Indeed apparantly nobody uses the feature and I have no problems removing it now. We can always bring it back later if we really want to. It has some quirks (e.g. it will show the file name even if after filtering it has no errors and the messaging isn't super clear either), so it wasn't fully fine tuned anyway.

braver:
I'll take some extra time to work on the panel and merge the other PR (#1173) into this to prevent conflicts.

braver:
> For #1163 I wanted:
> 
> Since next_result is not 100% reliable, make it reliable at least for when the panel is open. (So it doesn't feel buggy)
> Avoid background work. Esp. on_selection_modified should return early
> Also: on_pre_close calls fill_panel (Q: Can it be that the panel gets destroyed on uninstall, and we get a pre_close event from somewhere (we currently don't check anything there) which recreates the panel immediately? So, it does not find the syntax?)

derekjones:
You probably are already aware, but just in case not, when you double click a panel error now it opens an empty new file with the name of the file you are editing rather than jumping you to the error. This was working as of b852c7e

braver:
Nope, didn’t catch that yet, thanks a bunch for tagging along and reporting!

braver:
@derekjones I removed a but too much when taking out the remnants of the filtering feature. be60857 restores it.

braver:
Ha, so I learned something new. Apparently next_result continues working after the panel is destroyed, I didn't expect that. Now of course since we stop updating the panel when it's closed, at some point an error won't be in the same place it was last time you looked at the panel. But all of that is just excellent and to be expected IMO. So, that means we can safely destruct the panel when it closes. This should greatly reduce the chance of a panel being in use during an upgrade (i.e. when our panel syntax gets banged around by ST).

derekjones:
As of the time of this comment, if the panel is closed, it does not open the first time I hit save, I have to save twice for the panel to pop up. It is linting correctly on the first save and marking the error in the sidebar, it's just not popping the console open until a subsequent save.

braver:
Yeah, I'm doing some scrubbing to clean it up a bit, try get my head around it and to figure out a good trigger for removing the panel. So, sorry if I muck it up in some commits but the last one seems quite stable.

braver:
@kaste Thanks for your notes. One big question I have is as follows.

The original panel is designed around the idea that there is always a panel and that we can (and should) continuously update it. 

We can make it more lazy and that will prevent panels existing until someone needs it, and prevents background work.

Ideally we remove the panel after we're done with it. However, there is no event that says "the user is done with the panel". Of course the user can close the panel and that's explicit and clear, but usually you just hit ESC and that hides the panel instead of closing it. Also, you can switch to using a different panel and that also doesn't send us a useful event.

So, options:

1. Let the panel linger after it has been created, until the user removes it explicitly. We can optimise some of our updates around the panel's visibility so that we're still lazy and efficient. However, I don't we will be able to guarantee that next_result starts working as soon as you open the panel. Perhaps someone with more chops than me should have try at fixing that.
2. Pick some event to check the panel's visibility ("active") and remove the panel if it's no longer visible. I think it's sloppy to do this on some unrelated event. However it would usually mean the panel is created anew and that seems to help with it binding to the next_result.

I have tried to make 2 work but it doesn't sit well with me. I feel better about 1.

kaste:
I always thought that the panel_view is a custom view. 

So, I want it to behave exactly as we tell it. I expect it to only work when it is visible. 

Do you think the UX is sloppy or the implementation?

braver:
The implementation. I want an event like `panelview.on_close`, so I can destroy the panel in the callback, but I don't think that exists. All I can think of is destroying the panel `on_selection_modified` or `on_lint_result`.

panel_view.py
```diff
@@ -63,10 +63,10 @@ def on_selection_modified_async(self, view):
         if not panel_is_active(view.window()):
             return
 
-        active_view = State['active_view']
-        current_pos = get_current_pos(active_view)
+        current_pos = get_current_pos(view)
         if current_pos != State['current_pos']:
             State.update({
+                'active_view': view,
```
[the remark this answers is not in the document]
braver: > The view your getting here is always the primary view of a buffer.

Oh, so that's what that's for...

braver:
> I always thought that the panel_view is a custom view.
> So, I want it to behave exactly as we tell it. I expect it to only work when it is visible.

It is a view, but being an output panel view means something, and it's not completely like other views. We don't have full control, instead we configure the panel so that ST gives us a bunch of features for free. So, we don't tell it how to behave and we don't seem to have a full API to do so either. So, with other views we can perhaps get more control for more effort, but that would mean breaking out of the output panel pattern. And more effort.

Now, my understanding here is limited, but that's how I think it works.

So, the current state of this PR is as good as I'm going to get it. I added a feature (show panel on save), made it more lazy (don't update an invisible panel, don't make the panel unless someone wants it) and removed a broken an unused feature (filtering). Thinks that work as before that aren't ideal:

- `next_result` always works, unless you switch between panels or hide the panels (i.e. not explicitly destroy our output panel). This is the same for all other panels in Sublime Text (e.g. find results). This leads me to believe fixing this would require a change in ST core and it's not something we can (or should try to) do in this package. Someone else can have a go at it though, I might be wrong here. 

- Scrolling to the selected line in the panel doesn't when you open the panel. `panel.run_command("scroll_lines")` is called perhaps to try to remedy this, but I can't find the core bug referenced in the comments and I don't see any difference with that in there or not (so I removed it). This issue may be related: https://github.com/SublimeTextIssues/Core/issues/485. May also be a timing issue: if results roll in after you open the panel, it *does* scroll to your position. Master has the same problem by the way, except there the panel always exists so you don't notice.

So, @kaste I'd like to get this reviewed, corrected if needed, and shipped, so someone else can have a go at understanding and improving the things I don't know how to improve.

kaste:
I'll take it and push some stuff.

kaste:
- Redo selection bc it just wasn't reliable
- Apply focus fix
- Try loading syntax file before applying
- Ensure panel is always in panel list (ready to go for mousers)
- 'Fill' panel when it becomes visible (either per mouse or keyboard binding)

Known: It will not wait for the next lint result if you hit save in load/save mode

derekjones:
> Known: It will not wait for the next lint result if you hit save in load/save mode

Is this why I have to hit save twice to pop open the panel? Or is that a different issue?

braver:
@kaste this is really really 🌶 👍

kaste:
@derekjones If your linter only works on save then yes. But, this iteration on the panel took us 36 commits. So, the save mode must go to the next iteration.

@braver Yeah 🔥

kaste:
Now, let's see if we have a working master.

ingluisjimenez:
@braver @kaste thanks for working on this. Although I really liked the tooltips from v3, this is a big improvement from the previous implementation in v4.

kaste:
Thank you. 

These tooltips?

image

And these popped up automatically? (Asking bc I never had them. I had a pinned SL3 version bc SL3 didn't work with any kind of 'virtual' environment.)

braver:
Indeed. SL3 had popups that popped up automatically and followed your cursor around and made it impossible to use any other popups (e.g. auto-complete). They were very broken in several ways, which is why we now have popups you can open using a command (and thus via a keybinding if you like that).

ingluisjimenez:
@kaste Yes, they were like that, popping up automatically on save. I used to just hit Esc and the tooltip/popup would close automatically, and then I would just go and address the issue.

@braver can you please point me where I can read more about the new popups and how to automate it.

braver:
Oh, the on-save popup you mean. 

Check this: https://github.com/SublimeLinter/SublimeLinter/blob/master/SublimeLinter.sublime-settings#L94
(You can see the same file when you edit your settings in ST).

We also have a tooltip that kinda looks like kaste's screenshot. You can set a keybinding to open it (or just open it from the command palette):
https://github.com/SublimeLinter/SublimeLinter/blob/master/keymaps/Default%20(OSX).sublime-keymap#L9

ingluisjimenez:
Thanks for the tips @braver. Wonder if there would be a way to trigger the line report on save, same as `show_panel_on_save`?

braver:
No, don't think so. You'd have to be really lucky to be on the same line as an error too. The panel opens centered on your current line though, so that's almost as good (or better?).

ctf0:
@braver is it possible to configure the error type that shows up in the panel ?

atm both error & warnings shows, but mostly am only after the errors which i cant find any option to specify for the panel

braver:
No, don’t think we have that currently.

kaste:
@ctf0 https://packagecontrol.io/packages/SublimeLinter-addon-filter implements something similar, try it!

ctf0:
@kaste isnt it for filtering the errors result ? ex when having like 20 error reported & u look for something in specific or am i mistaken ?

still, what am looking for is more to hide the warning msgs from the panel output, so if we have warnings only then dont show the panel.

kaste:
Yes, it is for black and whitelisting reported errors. If you filter out warnings, only errors remain, and the panel will thus only popup (on save) if you have errors. But it is dynamic, and I wrote in the README over there, that you can define a keyboard shortcut for example to toggle between only errors/all errors etc.

So, that's why I think it can totally fulfill your needs. It's probably really powerful.

Issue #1122, opened 2018-03-06
Sometimes throws an error after uninstall about panel-syntax


Shelagh-Lewins:
I am trying to remove SublimeLinter 4.0.3 which installed itself this morning and see if I can figure out how to reinstall the old version 3 so I can do some work.

But when I selected Package Control > remove package > SublimeLint 4.0.3 I saw an error popup:

`Error loading syntax file "Packages/SublimeLinter/panel/panel.sublime-syntax": Unable to read Packages/SublimeLinter/panel/panel.sublime-syntax`

I also noticed that there was a second package SublimeLinter-eslint in Package Control, listed as v 4.0.1. Is there some conflict between the recent and latest versions?

FichteFoll:
@braver I think destroying the panel when plugins are unloaded might just be enough to get rid of that message.

braver:
Ok, that's worth a try.

kaste:
We do this https://github.com/SublimeLinter/SublimeLinter/pull/1025

FichteFoll:
So the problem would be that ST unloads the syntax definition before running plugins? That's quite unfortunate and should be reported upstream.

braver:
@FichteFoll didn't you already report something along those lines?

FichteFoll:
I reported something for Package Control, but that's only for mitigating the issue, and not in a clean way. Ideally, ST itself would just unload plugins first before it unloads other resource files that might have been in use before the plugins' unload handlers removed the references.

kaste:
We just put the '.sublime-syntax' in the root dir and 🎉 🎆 ?

FichteFoll:
You mean `Packages/SL_panel.sublime-syntax`?

Interesting idea, but I'd copy the file to `sublime.cache_path() / "SublimeLinter" / "panel.sublime-syntax"` instead.

Cooperdale:
As you can see in the above reference, I am having this error after upgrading. It all worked initially, then it started doing this and the plugin doesn't seem to work anymore.

kaste:
We still need full console logs when this happens!

Cooperdale:
Actually I'm getting different errors each time I start Sublime, now that particular error seems to be gone, but for instance I got this:
`SublimeLinter:` #1 csslint stylesheet.css ERROR:
 Linter crashed.`

kaste:
csslint is broken https://github.com/SublimeLinter/SublimeLinter/issues/1151

[...]

PR #1173, opened 2018-03-08
fix #1163 don't fill the panel if it doesn't exist

### LINKED ISSUE #1163: Do not create and update the panel in the background

braver:
This introduces a check if the panel exists in the current window. If it doesn't we don't `fill_panel()`, because that would create the panel. This way the user can create the panel and *then* we start updating it. 

This might help with #1122 or at least reduce occurrences of that. In any case it looks like a good idea to not spend effort updating the panel if the user doesn't want to look at it.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -44,8 +44,9 @@ def plugin_unloaded():
 @events.on(events.LINT_RESULT)
 def on_lint_result(buffer_id, **kwargs):
     for window in sublime.windows():
-        if buffer_id in buffer_ids_per_window(window):
-            fill_panel(window, update=True)
+        if window.find_output_panel(PANEL_NAME):
+            if buffer_id in buffer_ids_per_window(window):
+                fill_panel(window, update=True)
 
 
 class UpdateState(sublime_plugin.EventListener):
--- a/panel_view.py
+++ b/panel_view.py
@@ -44,9 +44,8 @@ def plugin_unloaded():
 @events.on(events.LINT_RESULT)
 def on_lint_result(buffer_id, **kwargs):
     for window in sublime.windows():
-        if window.find_output_panel(PANEL_NAME):
-            if buffer_id in buffer_ids_per_window(window):
-                fill_panel(window, update=True)
+        if window.find_output_panel(PANEL_NAME) and buffer_id in buffer_ids_per_window(window):
+            fill_panel(window, update=True)
 
 
 class UpdateState(sublime_plugin.EventListener):


panel_view.py
```diff
@@ -44,8 +44,9 @@ def plugin_unloaded():
 @events.on(events.LINT_RESULT)
 def on_lint_result(buffer_id, **kwargs):
     for window in sublime.windows():
-        if buffer_id in buffer_ids_per_window(window):
-            fill_panel(window, update=True)
+        if window.find_output_panel(PANEL_NAME):
+            if buffer_id in buffer_ids_per_window(window):
+                fill_panel(window, update=True)
```
kaste: Instead of the nested if you could just AND both conditions.

Also maybe add a comment `don't fill the panel if it doesn't exist`.

kaste:
Except the minor comment I made, I thought we would somehow receive an on close event for panels as soon as they hide. But that never happens in Sublime. 

So, I thought Sublime would kill the panel automatically (when unused) and we would recreate it very often. But that is not the case. I esp. hoped that we would create the panel on every ctrl-k a then, so that `next_result` would work again.

panel_view.py
braver: I think in one of my iterations it got a bit long so I split it. But that's better indeed.

braver:
We actually want to keep the panel after it's closed, because then next_result keeps on working even if the panel is invisible and that's super handy. What we could do perhaps is somehow check if the panel stops being the active_panel and  destroy it then. I'll look into that.

kaste:
(The problem I have is that `next_result` stops working, and then even ctrl-k a doesn't bring the keyboard binding back. Only mouse clicking.)

braver:
That’s odd. Perhaps a plug-in that’s interfering?

braver:
> (The problem I have is that next_result stops working, and then even ctrl-k a doesn't bring the keyboard binding back. Only mouse clicking.)

Ah I get what you're saying here now. That's correct, the panel doesn't steal focus when it opens so next_result doesn't start working on when it opens. We could try to do something here, but that's a different issue. 

Ok, active_panel doesn't mean "the panel next/prev_result is bound" but "the panel that is currently visible". You, if you open the panel at least once, we cannot clean it up. Or at least not any more than we're doing right now in `plugin_unloaded`. 

What this also does is remove our panel from the panel list until you create it. That's not a real problem, but a bit of a downside to this change. All in all I think it's good move and may alleviate some of #1122, but it won't really solve anything.

kaste:
Since, we cannot push single features to the users (bc updates are broken, we can only justify fixes and bigger improvements to actually ship) it is okay to iterate here, I think.

For #1163  I wanted:

- Since `next_result` is not 100% reliable, make it reliable at least for when the panel is open. (So it doesn't feel buggy)
- Avoid background work. Esp. `on_selection_modified` should return early 

Also: on_pre_close calls `fill_panel` (Q: Can it be that the panel gets destroyed on uninstall, and we get a pre_close event from somewhere (we currently don't check anything there) which recreates the panel immediately? So, it does not find the syntax?)

braver:
> Avoid background work.

braver:
Yeah, sure. I can play with this some more.

braver:
This work continues in #1172

kaste:
> What this also does is remove our panel from the panel list until you create it. That's not a real problem, but a bit of a downside to this change. 

I don't know what this means. What panel list?

> active_panel doesn't mean "the panel next/prev_result is bound" but "the panel that is currently visible". 

Isn't that fine. Then we can destroy it. I thought we get predictable behavior then, as long as its visible `next_result` works, otherwise not.

braver:
The panel list is shown if you click the icon furthest left on the status bar.

By convention panels are still bound to next_result after you close that panel. The Find Results panel works the same way. So that’s also why we took that approach initially.

Now though, I agree and let’s destroy our panel if you close it. Easier to understand and get reliable interaction that way.

PR #1174, opened 2018-03-08
make the fallback configurable and optional

### LINKED ISSUE #1171: improve the UX of the outline fallback for underline styles
After upgrading to the latest version, the one where the [settings changed](https://github.com/SublimeLinter/SublimeLinter/blob/master/messages/4.0.0.txt), and updating my settings, some things are using the `outline` mark style even though I've overridden the defaults to `squiggly_underline`. 

This screenshot illustrates the issue:

* There is a red outline on line 1
* There is a purple outline on line 4
* I expect both of these scenarios to use squiggly underlines like they did prior to the upgrade

image

I'm not sure where the purple is coming from. I expect that to be yellow. That's less of an issue, but I'd also like to know where that color is coming from and how to change it.

It's quite possible I could be missing some new config option. Any guidance would be appreciated.

[edit: removed logs and settings as they're irrelevant to the discussion /@braver]

braver:
Fix #1171. I think the fallback is in general nice enough, but I'm not a fan. However, I don't want to force my opinion here. I'm pretty opinionated about a lot of the user-facing stuff we do, but here there is no better or worse, just that I like better. 
I think the outlines are well implemented and suppressing them while editing really works very well for me. They're well done and I can see why you'd want them, but I kinda want the option of not having them. With flake8 it's not so bad, but with eslint you sometimes get outlines over huge areas and it just makes everything harder to read. Another user had similar feedback. 

So, this PR probably doesn't cover what needs to be done to turn of the outlines, I think flake8 and will need to be adjusted to work without them (e.g. get an option)?

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -35,6 +35,11 @@
     // the regions. E.g. try 'comment' to get a soft gray.
     "highlights.demote_scope": "",
 
+    // Fall back to "fill" or "outline" when using *_underline mark styles.
+    // Sublime Text doesn't draw underlines on whitespace, so you might
+    // miss information otherwise. Set to "none" to disable the fallback.
+    "highlights.fallback": "outline",
+
     // Lint Modes determine when the linter is run
     // background: asynchronously on every change
     // load_save: when a file is opened and every time it's saved
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -30,7 +30,6 @@ UNDERLINE_STYLES = (
 )
 
 SOME_WS = re.compile('\s')
-FALLBACK_MARK_STYLE = 'outline'
 
 WS_REGIONS = re.compile('(^\s+$|\n)')
 DEMOTE_WHILE_BUSY_MARKER = '%DWB%'
@@ -310,10 +309,12 @@ def prepare_highlights_data(view, linter_name, errors, demote_predicate):
         selected_text = view.substr(region)
 
         demote_while_busy = demote_predicate(selected_text, **error)
+        fallback_style = persist.settings.get('highlights.fallback')
 
         # Work around Sublime bug, which cannot draw 'underlines' on spaces
         if mark_style in UNDERLINE_STYLES and SOME_WS.search(selected_text):
-            mark_style = FALLBACK_MARK_STYLE
+            if fallback_style != 'none':
+                mark_style = fallback_style
 
         flags = MARK_STYLES[mark_style]
         if not persist.settings.get('show_marks_in_minimap'):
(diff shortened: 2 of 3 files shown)


kaste:
I'm really 👎 here. 

The problem here is that without the fallback we would get VSCode like massive squiggles, underlines over huge regions. 

Consider what we had before. E.g. flake8. A clean, less noisy editor. After I implemented this, I thought immediately "WTF why does it show me all these nonsensical ws issues. Nobody cares about that." But the fact is, a) you could adjust the linter rules, this is what *you* say to all the people wanting the 'ignore_match' option back. b) We could as well provide the option `highlights.ignore_ws_errors`. (Bc that's what these people actually had before. We had **no** underlines/squiggles for these lint errors.)

Maybe the regex when we switch to fallback is wrong. We could instead of SOME_WS use WS_REGIONS (which is ws only or multiline). 

(I think most people have strict rules in linter config, often these rules are anal-masochistic and cannot be changed by a single person. These are basically for CI and for the managers. But people do not want all these highlighted regions. Also a problem in SL4: The backend is way faster and responsive, so with the default debounce of .3s. The UI updates with eslint .8s after the last edit on my machine. In SL3 it took 1.3s in the same scenario, but 1.3s is also more time to start typing again, and thus skip a re-render altogether.)

braver:
I was under the impression that previously in the case of a white space error, we would highlight the word next to the white space. If that is not the case, or not possible, then sure we need the fallback. I am definitely not saying people want to ignore the white space errors! Indeed if they want to they should configure that.

So as I understand it now, if we don’t want the fallback we would have to completely rethink highlighting and make a super complicated system to re-interpret what linters tell us to highlight. That’s not a path I’m willing to take. We’re also definitely doing a lot better than say VS Code. 

I should also clarify that I think our (well, yours really, credit where it’s due) implementation with the fallback is awesome, and my problem lies more with what eslint reports.

So perhaps I should go complain with eslint.

Ok. Closing this. We can point to this conversation if people want to hear our reasoning behind the feature.

kaste:
The contract here is that linters/plugins tells us where the error is/what to draw. And we ensure that it does look acceptable. 

- We think that multiline regions shouldn't be squiggled.
- We work-around Sublime bug 1: ws only regions are hidden for underline styles, but drawn for outline styles (implemented)
- We work around Sublime bug 2: empty regions at the beginning of a line are hidden for outline styles, but drawn with an underline for all other styles (still to fix, I thought I did that, but apparently not)

---

IIRC SL3 eslint didn't show missing semi and indentation errors using underline styles. SL3 flake8 didn't show most ws errors, some errors marked actually the next 'valid' token.

braver:
> The contract here is that linters/plugins tells us where the error is/what to draw. And we ensure that it does look acceptable.

Yes, I'm coming around to this view too. 👍

PR #1175, opened 2018-03-09
Fix typo: launch not lunch

ericwb:
I believe the troubleshooting docs meant to refer to launch and
not lunch. :)

Signed-off-by: Eric Brown <browne@vmware.com>


braver:
LOL, must’ve been hungry when I wrote that 😄

PR #1176, opened 2018-03-09
Add missing documentation on regex group 'code'

ericwb:
The linter attributes documentation had a couple issues. Firstly,
the 'line' and 'message' name and description appear as one line
in the documentation. This makes it confusing to know that there
are actually two separate capture groups.

Secondly, the capture group 'code' was undocumented. This is a
known group that plugins such as pylint use to show the error code
corresponding to the message.

Signed-off-by: Eric Brown <browne@vmware.com>


braver:
Thanks for reviewing the docs and all these pull requests, it really helps 💯

PR #1180, opened 2018-03-10
Expand info about overriding selector for given linter.

### LINKED ISSUE #1177: Question: how to activate explicitly multiple linters for given file types?
Let say I have ansible files, which are plain .yml files.
How can I add ansible-lint and yaml-lint to such files explicitly?

From what I see the only plugin that activates itself is 'annotations'...

nvtkaszpir:
Fixes #1177

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage.xml
 _build/
 .mypy_cache/
 .ropeproject/
+docs-out
--- a/docs/linter_settings.rst
+++ b/docs/linter_settings.rst
@@ -101,3 +101,36 @@ binary on your system matching that version (using PATH).
 
 It then executes ``python -m script_name``
 (where script_name is e.g. ``flake8``).
+
+selector
+--------
+
+This takes precedence over deprecated `syntax` property.
+This allows override when given linter is activated for specfic file types.
+It should be a string containing a list of comma separated selectors.
+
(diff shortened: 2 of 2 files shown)


kaste:
That's a super nice contribution. 

I don't know if you need PackageDev honestly, I thought this is ScopeHunter. Lol.

nvtkaszpir:
Hm actually now I'm not sure how I managed to get that menu option....
Updated PR.

kaste:
I remember braver said this is built-in. 😁

braver:
Show Scope Name is built in. You can’t make color shemes without it.

braver:
@nvtkaszpir I totally appreciate you putting effort into the docs here. I have some minor nitpicking here wrt the writing style (and I really think we should adhere to the one true brace style). Want me to provide feedback, or should I just merge and edit afterwards?

nvtkaszpir:
Whatever you prefer.
I added the braces to avoid issues with sphinx shouting errors, so it's merely issue with whe way we use sphinx directives.

If we do not want to have a syntax highlight there then maybe use `text` - but also loose formatting.
Otherwise I suggest sticking to proper syntax with braces. We assume project users are advanced enough to edit json files ;-)

braver:
Ok, cool. I need to write some docs myself later, so I’ll edit then. Thanks!

PR #1183, opened 2018-03-11
Add sphinx-build linkcheck to travis, speed up builds

nvtkaszpir:
I'm a bit mental about docs recently...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,5 @@ install:
 
 script:
   - flake8
-  - sphinx-build docs docs-out
+  - sphinx-build -j auto docs docs-out
+  - sphinx-build -j auto -b linkcheck docs docs-out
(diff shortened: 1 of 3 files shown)


braver:
I usually read the travis.yml for how-to (ie. what to install and what commands to run). Perhaps add some comments there about what happens (e.g. the html you can read locally is in the docs-out folder) and how to review your edits that way?

nvtkaszpir:
As @braver said, read .travis.yml.
Actually  I was thinking about simple script (or makefile) with different targets, which could eventually replace travis `script:` lines and allow run multiple tests concurrently. But this requires further discussion.

For now I can add a new document to docs about how to develop the plugin using pyenv+virtualenv. You want it in this PR or a new one?

kaste:
Nah, that's okay. I thought they ship an editor to use locally. (Auto-build, etc.)

nvtkaszpir:
for auto builds you can try using https://pypi.python.org/pypi/sphinx-autobuild

kaste:
Okay, thanks.

PR #1184, opened 2018-03-11
Changed default theme from "none" to "Default"

pykong:
**Because:**
- does not interfere with using inbuilt gutter icons: `"circle"`, `"dot"` and `"bookmark"`
- lowers barrier for using new icons


braver:
Oh, I think I didn't fully realise you could still use the built-in icons. 👍

PR #1185, opened 2018-03-11
Added gutter theme pattern

pykong:
Additional prevention for shooting yourself in the foot.
**Untested.** (Does space need to be escaped?)
Gutter theme author would need to register with us first when adding a new theme.

### DIFF
--- a/resources/settings-schema.json
+++ b/resources/settings-schema.json
@@ -9,7 +9,8 @@
             "type":"number"
         },
         "gutter_theme":{
-            "type":"string"
+            "type":"string",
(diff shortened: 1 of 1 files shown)


braver:
I think people can copy paste well enough to work this out themselves. And we do kinda want people to put full paths in here for their custom icons, right? I would definitely don't want to introduce any burden on us about new themes (I'd much rather have there be no themes at all here, just our Default).

pykong:
Now we have discussed it, our stance on the issue is clear.

PR #1190, opened 2018-03-12
Two fixes for NodeLinter

### LINKED ISSUE #1118: NVM: cannot locate 'eslint' / 'node' etc
After recent update, Sublime always popup the tips:

`SublimeLinter: WARNING: eslint cannot locate 'eslint'`.

In detail:

I haven't install global eslint, because I have to extend the eslint rules in config file in projects, If using the global eslint, the `xxx-eslint-config` package is not used.

When the project is configed properly, and with local eslint installed, everything works very well.

But if I open a file not in project (in a single window), there always a popup tips in the sublime console.

In previous versions (not sure which version), no such problem.

### LINKED ISSUE #1188: NodeLinter `disable_if_not_dependency` doesn't work as intended

kaste:
Fixes #1118 
Fixes #1188

### DIFF
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -53,11 +53,9 @@ class NodeLinter(linter.Linter):
             local_cmd = self.find_local_cmd_path(cmd[0])
             if local_cmd:
                 return True, local_cmd
-            elif self.get_view_settings().get('disable_if_not_dependency', False):
-                util.printf(
-                    "Disabled {}. Did you `npm install {}`?."
-                    .format(self.name, cmd[0]))
-                return True, None
+
+        if self.get_view_settings().get('disable_if_not_dependency', False):
+            return True, None
 
         global_cmd = util.which(cmd[0])
         if global_cmd:
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -3,11 +3,15 @@
 import codecs
 import json
 import hashlib
+import logging
 import os
 
 import sublime
 
 from .. import linter, util
+
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


kaste:
Better name for `disable_if_not_dependency`?

braver:
This disables the linter if it’s not listed as a dependency in the package json, right? It’s not a terrible name. 🤷🏼‍♂️

kaste:
Yeah, but we don't do that. We don't parse `package.json`. We search for `package.json` and take its folder as the project folder so we can peek at `node_modules/bin/linter_name`.

kaste:
(`context_sensitive_executable_path` is also correct but why didn't they choose something like `which_executable`)

braver:
Ah, that's a bit misleading then. `disable_if_not_installed`? `requires_installation`?

braver:
I don't really care about the setting name, LGTM for me otherwise.

Grantlyk:
Hey guys, when will this PR be released?

braver:
@Grantlyk We're having some issues with the upgrade process itself, so we probably want to merge #1172 too to (hopefully) help in that regard.

broonage:
Any update? My linter is all broken (been a couple of weeks since SublimeLinter was updated), just updated to 4.2.0, but no sign of this fix. :)

braver:
The contents of this PR have been released. If you're having problems please open a new issue with some more info (debug logs) so we can figure out what's wrong with your setup.

broonage:
Ok, sorry I didn't realise as I still get the same message and assumed it was not pushed. I will do, thanks.

broonage:
I got it working, I had not realised the user settings had changed. thanks

braver:
🎉

PR #1191, opened 2018-03-12
Better excludes

### LINKED ISSUE #1158: Better `excludes` patterns
Today I wanted to do something like this. 

```
		"excludes": [
             "*/node_modules/*",
             "!${folder}*",
			 "<untitled>"
         ]
```

This is not supported by our poor 😥 `fnmatch` from python std library but would be trivial to add.

We could also support a special `'<untitled>'`.

The above should then be read as, do only lint files from my current project, exclude new/untitled buffers and files from 'node_modules'.

kaste:
Fixes #1158 

This is somewhat naive, but maybe good enough (KISS, YAGNI you name it)

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -181,12 +181,16 @@ def filter_linters(linters, view):
             matched = False
 
             for pattern in excludes:
-                if fnmatch(filename, pattern):
+                if pattern.startswith('!'):
+                    matched = not fnmatch(filename, pattern[1:])
+                else:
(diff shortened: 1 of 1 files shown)


kaste:
Works for something like 

image

Or the example in the original issue.

braver:
Fancy! 🎉

braver:
Works! 🛥 it!

FichteFoll:
Docs?

braver:
I updated the docs

Issue #1149, opened 2018-03-07
Reduce noise/Allow quiet working


kaste:
User (short) stories (from various comments):

> I want to disable everything to write pseudo-code or something quick.

> I open a random file on my disk, or from the internet, which obviously doesn't match my lint style if any at all.

> I just want to *read* code.

Proposed *solutions* are:

- Enable/disable linting (per shortcut) of a view
- Enable/disable SublimeLinter (per shortcut)
- Quickly toggle highlighted regions (with *instant* redraws!).

Related #1101, #98

braver:
Switching off the views seems like a nice and easy way to go about it? Perhaps you don’t care that the linter runs, but you just don’t want to know about it right now.

kaste:
This will most likely be a moderated thread. Please add user stories that vastly differ from what is already in the first description.

braver:
Ok, I can't think of any. At least none that cannot be dealt with using project configuration for instance. I do have opinions on the proposed solutions in relation to the user stories. Want them now or later?

kaste:
I hope we get some more feedback. I only do bugfixes anyway as long as we have real crashes.

I'm especially unsure about: toggle for one view, the current window, or globally.

FichteFoll:
>I'm especially unsure about: toggle for one view, the current window, or globally.

Imo all of these are useful, though personally I found disabling it for the current view to be the most frequently needed version.

braver:
> I'm especially unsure about: toggle for one view, the current window, or globally.

For user story 2, per view wouldn't be enough but per window would. I think per window would be achieved by looping over all views in the window so you kinda get that for free. So, I think per view makes the most sense unless I think of another user story.

FichteFoll:
When implementing per-view settings, you need to acces view settings anyway, so you might as well use the same setting name in project settings and get both functionalities at the same time. The disable state would be saved across sessions for free as well.

I just thought of a really neat UX using Input Handlers, but unfortunately those aren't available in the stable build iirc.

Mattiaslndstrm:
Two thoughts:

Mattiaslndstrm:
Two thoughts: 

For current view would be most useful for me. It would only be confusing if I am to switch file and find that it's not linting. It might also be confusing to remember if I have disabled linting on a file, so maybe keeping one error telling the user that linting is disabled would be a good idea?

[...]

PR #1192, opened 2018-03-12
Fast toggle highlighted regions

kaste:
Related #1149 

This implements fast toggling of the highlighted regions of the current view. 

Since the linter stays active, we have *instant* visual feedback.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -20,6 +20,11 @@
         }
     },
     {
+        "caption": "SublimeLinter: Toggle highlighted regions",
+        "command": "sublime_linter_toggle_highlights"
+
+    },
+    {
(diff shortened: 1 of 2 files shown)


braver:
I really like this approach. Because the linter keeps running you can always configure it to show the panel on save so you don't miss anything. And it just works really well.

I *would* like this to completely silence the editing area of the view (I think it's good that the status bar still reports that there are errors you're ignoring), so the gutter marks should go too.

kaste:
I thought about the gutter mark but this would be the next iteration (Actually bc of this 'protected_regions' thing which doesn't even work that good, I don't know if this would take an hour to implement or if it just has to sit two weeks and then you have the right (simple) idea.)

So, I think this is a good iteration and would merge.

braver:
Ok, sure, let's merge then iterate further 👍

PR #1194, opened 2018-03-12
Executable for the win

### LINKED ISSUE #1086: Make the `executable` setting available for all linters
Currently, the `executable` setting is 'on' for NodeLinter, PythonLinter et.al. But not for base linters bc all 'base' linters first have to pass `Linter.can_lint`. 

After the refactorings here `can_lint` has become easy. We inline the logic of `can_lint` into `Linter.build_cmd`. And remove all call sites. Note that `can_lint` only does a `cls.which` call, and we already have a `self.which` in `build_cmd`, so we basically only have to get the ifs right.

### LINKED ISSUE #1087: Remove: Intent to remove 'versioning' feature
The version check feature is mostly unused and confuses people. Note that all sub linters like NodeLinter, PythonLinter et.al. do not implement it. (But of course, plugin authors still provide the versioning attributes without knowing that they don't do anything.)

Currently 'versioning' was a simple 'on-off' check. This is not very useful.

kaste:
Fixes #1087 
Fixes #1086 

This is basically how I want this.

### DIFF
--- a/lint/base_linter/composer_linter.py
+++ b/lint/base_linter/composer_linter.py
@@ -173,8 +173,3 @@ class ComposerLinter(linter.Linter):
         """Calculate the hash of the manifest file."""
         f = codecs.open(self.manifest_path, 'r', 'utf-8')
         return hashlib.sha1(f.read().encode('utf-8')).hexdigest()
-
-    @classmethod
-    def can_lint(cls):
-        """Assume the linter can lint."""
-        return True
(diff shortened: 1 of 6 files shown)


kaste:
image

braver:
🌶 🌶

braver:
Ok, looks very good, let's 🚤 . Any real down sides to this?

kaste:
I'm 100% that this is the right change. But of course we can break people here.

braver:
Well, only one way to find out.

kaste:
So, we're heading v4.1 now?

braver:
Looks like it. I'll update the messages etc.

braver:
Ok, it's 🚀

kaste:
🌟 🌛 

Now we get all these 💥 issues again 😒

PR #1213, opened 2018-03-14
fix #1137 add code, type and linter to template

### LINKED ISSUE #1137: Display rule name(s) in status bar
the previous version used to display the name of the rule(s) in the status bar. I would like for this behavior to return (or to be configurable), because I find it a bit cumbersome having to open the `show all errors` panel first (or even worse: having to pick up the mouse and hover over the gutter ...).

screen shot 2018-03-06 at 21 27 40

braver:
In stead of just "messages", adds the other info available in `errors_per_line` as optional parts in the template. E.g.:

```
"statusbar.messages_template": "{message} -- {type} << {linter} ^^ {code}"
```

Oh yeah, this breaks for people who customized the current template because it’s no longer messages plural. Can’t imagine a lot of people did anything else with it than setting it to “” though.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -99,8 +99,11 @@
 
     // Show the messages for problems at your cursor position
     // {message} will be replaced by the actual messages.
+    // {linter} will be replaced by the linter reporting the error.
+    // {type} will be replaced by either warning or error.
+    // {code} will be replaced by the error code.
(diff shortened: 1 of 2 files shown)


kaste:
Didn't even know anymore that was a requested feature. 

👍

braver:
Yeah, I was looking something fun and simple yesterday.

PR #1221, opened 2018-03-15
loadsave/manual: show panel on_lint_end if errors

### LINKED ISSUE #1200: Implement 'there are errors' popup for 'save' only linters
- The user has `lint_mode` set to `load/save`.
- A linter opts-in to run only after save.

In both cases we cannot evaluate if we have errors `on_post_save` but have to wait for the next result of the linter. `LINT_END(bid)` is probably the correct event here.

braver:
This might take some more commits, but attempts to fix #1200.

- [ ] show the panel for load_save and manual modes (ie. on `events.LINT_END`)
- [ ] show the panel if results come in after `on_post_save_async` for background linting
- [ ] show the panel if the panel if there is a lint-only linter active

My first attempt addresses the first issue here by showing the panel if results come in and `lint_mode` is not "background". It should however not show `on_load` I think.

The second issue seems really tricky to me. If you're editing immediately after saving, how do we know if those errors are related to saving? When do we unsubscribe from LINT_END? Asynchronicity breaks my brain, so perhaps someone else should try to tackle that, or perhaps point me in the right direction.

-----

Edit: ok, that's as far as I get right now.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -39,6 +39,7 @@ def plugin_loaded():
 
 def plugin_unloaded():
     events.off(on_lint_result)
+    events.off(on_finished_linting)
 
     for window in sublime.windows():
         window.destroy_output_panel(PANEL_NAME)
@@ -49,6 +50,19 @@ def on_lint_result(buffer_id, **kwargs):
     for window in sublime.windows():
         if panel_is_active(window) and buffer_id in buffer_ids_per_window(window):
             fill_panel(window)
+
+
+@events.on(events.LINT_END)
+def on_finished_linting(buffer_id, **kwargs):
+    show_panel_on_save = persist.settings.get('show_panel_on_save')
+    lint_mode = persist.settings.get('lint_mode')
+
+    if show_panel_on_save == 'never' or lint_mode == 'background':
+        return
+
+    for window in sublime.windows():
+        if buffer_id in buffer_ids_per_window(window) and not panel_is_active(window):
+            show_panel_if_errors(window, buffer_id)
 
 
 class UpdateState(sublime_plugin.EventListener):
@@ -94,12 +108,7 @@ class UpdateState(sublime_plugin.EventListener):
         if not window or show_panel_on_save == 'never' or panel_is_active(window):
             return
 
-        errors_by_bid = get_window_errors(window, persist.errors)
-
-        if show_panel_on_save == 'window' and errors_by_bid:
-            window.run_command("show_panel", {"panel": OUTPUT_PANEL})
-        elif view.buffer_id() in errors_by_bid:
-            window.run_command("show_panel", {"panel": OUTPUT_PANEL})
+        show_panel_if_errors(window, view.buffer_id())
 
     def on_post_window_command(self, window, command_name, args):
         if command_name != 'show_panel':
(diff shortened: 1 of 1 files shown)


kaste:
PTAL I squashed it. So if I messed it up, I can restore the old branch from my local repo. (I can also try to rewrite the branch bc you're not even an author anymore. Boy, I messed it up with git this time.)

braver:
Oh, I have my local copy too, so no worries.

braver:
Ok, this looks pretty good to me. Some thoughts though, not necessarily related to this iteration but I think something we skipped over the last time (e.g. case no. 2 in my OP here). 

I think there is still a small chance that lint results come in, in background mode, after `on_post_save_async` runs. Right now we only wait for lint_end when `buffers_effective_lint_mode_is_not_background`. 

Try this:

- set `"delay"` to 5
- create the first lint error in a file (e.g. too many blank lines)
- immediately hit save, or at least within 5 secs

So, if you beat out the delay + however long the linter takes, the panel will not show. 

If we can't know if lint_start was from before the save, we can't reasonably wait for the next lint_end, so I *think* we might not be able to do a lot better. Or perhaps we should always wait for the delay period.

kaste:
Grrr. I push some more. Also we popup in manual mode, no?

braver:
> Also we popup in manual mode, no?

Yeah, for both `load_save` and `manual` we don't have to look at the `on_post_save_async` event. We can just show the panel if stuff comes in on lint_end.

kaste:
Ah, you think we should open the panel in manual mode.

braver:
Yeah, that's nice feedback right? It's called "on save", but in manual mode it's nice to get the panel popup if there's a result. I think.

kaste:
> Yeah, for both load_save and manual we don't have to look at the on_post_save_async event. We can just show the panel if stuff comes in on lint_end.

I liked your implementation a lot and it is easier, only if you add the dirty check via `view.change_count()` it becomes relatively hard. The dirty check comes almost for free/no effort in my variant. Also LINT_END events are fired way more often then `save` events, so I hoped it would be more efficient.

braver:
Ah yes. The parts of it are doable, but the total orchestration of it is 💥

kaste:
PTAL again, easy implementation

panel_view.py
```diff
@@ -92,12 +107,15 @@ def on_post_save_async(self, view):
...
-        if show_panel_on_save == 'window' and errors_by_bid:
-            window.run_command("show_panel", {"panel": OUTPUT_PANEL})
-        elif view.buffer_id() in errors_by_bid:
-            window.run_command("show_panel", {"panel": OUTPUT_PANEL})
+        if buffers_effective_lint_mode_is_background(bid):
+            show_panel_if_errors(window, bid)
+
+    def on_modified_async(self, view):
+        bid = view.buffer_id()
+        State['awaited_views'].discard(bid)
```
braver: Perhaps we need to do the same thing `on_close(view)`?

braver:
👍 I think the approach here is really solid. I'll see if I can find any holes in it, but looks really good right now.

panel_view.py
kaste: Yes

kaste: Also awaited_view -> just_saved_buffers. The important part here is to have buffer in the name bc that's what's usually confusing.

panel_view.py
```diff
@@ -132,15 +132,15 @@ def on_post_window_command(self, window, command_name, args):
...
 
     def on_pre_close(self, view):
         bid = view.buffer_id()
-        State['awaited_views'].discard(bid)
+        State['just_saved_buffers'].discard(bid)
 
     def on_modified_async(self, view):
         bid = view.buffer_id()
-        State['awaited_views'].discard(bid)
+        State['just_saved_buffers'].discard(bid)
```
braver: So, this means, if I save, then edit, before the panel would be shown, the panel is canceled. At least, that's the behaviour I see now. Not saying that is wrong, but is that also the intention here?

kaste: Yes, that's exactly the intention here.

braver: Excellent, makes sense to me. 👍🏻

braver:
LGTM I guess, can’t think of a new way to break this so we might as well let more people play with it.

PR #1227, opened 2018-03-17
fix #1222 On save only mode

### LINKED ISSUE #1222: lint_mode "save only" missing
Not sure why "save only" lint_mode is missing in the new version. Please bring it back.

braver:
Back by popular demand: "save" `lint_mode`. Kinda like manual because it only runs on demand, kinda like "load_save", but without the "load" part. It's a cheap feature, so why not.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -39,6 +39,7 @@
     // background: asynchronously on every change
     // load_save: when a file is opened and every time it's saved
     // manual: only when calling the Lint This View command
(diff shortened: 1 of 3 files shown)


kaste:
Okay that's an easy iteration. We should open a new issue for https://github.com/SublimeLinter/SublimeLinter/issues/1222#comment-373849671

braver:
Oh, I didn’t consider that one at all. Sorry. Let’s cover that separately.

braver:
Ok, that comment is now separated at #1230

PR #1231, opened 2018-03-18
Feature - docs: sort linter settings

nvtkaszpir:
Move sections of linter_settings.rst in such way, so that they are in alphabetical order.
Section move per commit for easier readability.


braver:
Thanks!

braver:
Oh, dang, now it conflicts.

Just a single commit is good enough by the way, for docs only the end result counts IMO and that’s what I review.

nvtkaszpir:
Fixed merge conflict.

PR #1232, opened 2018-03-18
Feature - docs: expand troubleshooting page and add installation page

### LINKED ISSUE #1091: Installation docs are missing
Hello, I can't find any docs for the installation. Is this because of the upgrade to version 4?

http://www.sublimelinter.com/en/latest/installation.html

There are also no docs in the repository here for the installation.

nvtkaszpir:
Should theoretically reduce number of issues reported on GitHub ;)
Fixes #1091.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -7,3 +7,5 @@ First install PackageControl and then see its `usage <https://packagecontrol.io/
 which explains how to install plugins to SublimeText.
 
 Notice, that it may be required to restart SublimeText after plugin installation.
+
+Make sure to read `SublimeLinter messages <https://github.com/SublimeLinter/SublimeLinter/tree/master/messages>`_.
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,0 +1,9 @@
+Installation
+==================
+
+SublimeLinter and corresponding linter plugins should be installed using `PackageControl <https://packagecontrol.io/installation>`_.
+
+First install PackageControl and then see its `usage <https://packagecontrol.io/docs/usage>`_,
+which explains how to install plugins to SublimeText.
+
+Notice, that it may be required to restart SublimeText after plugin installation.
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -23,15 +23,37 @@ Here are the most common reasons why a linter does not work:
   to a known syntax (e.g. ``"html"``). The detected syntax is printed to the
   console in debug mode.
   Also note that plugins should move to using the selector setting
-  instead of the old syntaxes attribute. You can use the "selector" linter
+  instead of the old syntaxes attribute. You can use the :ref:`selector` linter
   setting right now instead of the "syntax_map".
 
-- The linter binary is not installed.
+- The linter binary or it dependencies are not installed.
   Be sure to install the linter as documented in the linter plugin’s README.
 
 - The linter binary is installed,
   but its path is not available to SublimeLinter.
   Follow the steps in :ref:`debugging-path-problems` below.
+
+Less common reasons:
+
+- Extra SublimeLinter plugin is not compatible with SublimeLinter 4.x.
+  Such plugin should be updated according to the new API. First, search
+  PackageControl for updates, or check out corresponding repos of the plugins.
+  You may need to create issue to request update of the plugin.
+  Example: SublimeLinter-eslint.
+  You may need to read `how to upgrade <https://github.com/SublimeLinter/SublimeLinter/blob/master/messages/4.0.0.txt#L80>`_ secion.
+
+- Only gutter theme icons are not working.
+  This usually happens when you upgrade. If you want to use another gutter
+  theme, you have to update your styles to use the correct icons
+  (``warning`` and ``error`` usually).
+
+- Other plugins are blocking SublimeText features, such as popups or gutter
+  icons. Usually you can see in the console that SublimeLinter works and it
+  should show some output, but some events are blocked. This usually happens
+  due to other plugins - you can temporarily disable selected plugins and
+  restart SublimeText until you find the offending plugin.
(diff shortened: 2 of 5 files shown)


braver:
Well, at least it gives us a page to point to ;)

docs/troubleshooting.rst
```diff
@@ -23,16 +23,38 @@ Here are the most common reasons why a linter does not work:
...
 
-- The linter binary is not installed.
+- The linter binary or it dependencies are not installed.
   Be sure to install the linter as documented in the linter plugin’s README.
 
 - The linter binary is installed,
   but its path is not available to SublimeLinter.
   Follow the steps in :ref:`debugging-path-problems` below.
 
+Less common reasons:
```
braver: I wouldn't want to add a section like this. For the really uncommon reasons, either a user knows how to debug, or he asks us questions. I'd rather resolve them in the specific issue and perhaps search through the issues when something familiar comes up, than try to maintain a document about it. 

E.g. All plugins that aren't compatible already have issues open for this. Some simply aren't maintained.

If there are conflicts like with gutter icons or popups, either a user already knows basic troubleshooting tactics and figures this out without the docs, or he's simply not going to troubleshoot (or do it wrong).

nvtkaszpir: OK, I get the idea.
I suggest replacing it with just ' Anything else should be reported as GitHub issue.'?

nvtkaszpir:
One more thing, maybe it is worth to update https://github.com/SublimeLinter/SublimeLinter/blob/master/messages/install.txt with the url at the end? (SublimeLinter instead of SublimeLinter3)

docs/troubleshooting.rst
braver: Yeah, that’s a good idea.

braver:
Good catch, that should be without the 3

docs/troubleshooting.rst
```diff
@@ -23,16 +23,19 @@ Here are the most common reasons why a linter does not work:
   to a known syntax (e.g. ``"html"``). The detected syntax is printed to the
   console in debug mode.
   Also note that plugins should move to using the selector setting
-  instead of the old syntaxes attribute. You can use the "selector" linter
+  instead of the old syntaxes attribute. You can use the :ref:`selector` linter
   setting right now instead of the "syntax_map".
 
-- The linter binary is not installed.
+- The linter binary or it dependencies are not installed.
```
FichteFoll: Its

docs/troubleshooting.rst
```diff
@@ -23,16 +23,19 @@ Here are the most common reasons why a linter does not work:
...
   Be sure to install the linter as documented in the linter plugin’s README.
 
 - The linter binary is installed,
   but its path is not available to SublimeLinter.
   Follow the steps in :ref:`debugging-path-problems` below.
 
+Anythng else should be searched across GitHub issues and if not found then
+reported.
```
FichteFoll: Maybe also add a sentence about closed issues? I think many users only consider open ones.

docs/troubleshooting.rst
```diff
@@ -23,16 +23,19 @@ Here are the most common reasons why a linter does not work:
...
 
-- The linter binary is not installed.
+- The linter binary or it dependencies are not installed.
   Be sure to install the linter as documented in the linter plugin’s README.
 
 - The linter binary is installed,
   but its path is not available to SublimeLinter.
   Follow the steps in :ref:`debugging-path-problems` below.
 
+Anythng else should be searched across GitHub issues and if not found then
```
FichteFoll: Anything

docs/installation.rst
```diff
@@ -0,0 +1,11 @@
+Installation
+==================
+
+SublimeLinter and corresponding linter plugins should be installed using `PackageControl <https://packagecontrol.io/installation>`_.
+
+First install PackageControl and then see its `usage <https://packagecontrol.io/docs/usage>`_,
+which explains how to install plugins to SublimeText.
```
FichteFoll: The correct term is "packages" here.

docs/installation.rst
```diff
@@ -0,0 +1,11 @@
...
+
+Notice, that it may be required to restart SublimeText after plugin installation.
```
FichteFoll: The comma is wrong. Or well, I certainly would not write it.

docs/installation.rst
```diff
@@ -0,0 +1,11 @@
...
+
+Make sure to read `SublimeLinter messages <https://github.com/SublimeLinter/SublimeLinter/tree/master/messages>`_.
```
FichteFoll: This reference seems odd. Messages are shown on updates, so users should rather read the messages that apply to them in their editor when they open.

FichteFoll:
Aw, I was too slow.

braver:
I’m just glad to have it written and merged. I usually re-read and correct docs afterwards, just before releases. I don’t like doing the language corrections and typos via code review, instead I edit afterwards.

docs/troubleshooting.rst
nvtkaszpir: fixed

docs/troubleshooting.rst
nvtkaszpir: Good idea.

docs/troubleshooting.rst
nvtkaszpir: fixed.

docs/installation.rst
nvtkaszpir: fixed

docs/installation.rst
nvtkaszpir: fixed.

docs/installation.rst
nvtkaszpir: Yes, but looking at the amount of issues on GitHub it seems people are not reading those.
Also AFAIR it shows only last change, not sure if it shows all messages between current and freshly installed versions....

nvtkaszpir:
Fixes will be in new PR.

PR #1234, opened 2018-03-18
Variable debounce time and upper limit to reduce concurrency

### LINKED ISSUE #1201: Reduce concurrency of the backend
Okay

- Variable, automatic debounce time is generally 👍 . It is trivial to implement, without any downside. If you use fast and then slow linters, or just small files and then huge ones, the delay will adjust automatically. The only 'thing' is, we need to magic-guess an upper bound, currently at 2s, otherwise it feels buggy. (2s might be too much even bc this is a debounce time not a delay. 2s means you have a lot of time to restart the debouncer. delay actually means `idle_time`.) #1234 ✔️ 

- Upper limit. I don't like that one. Sure, normally you have it. But with a low upper limit, you basically get a long queue length. On my computer short spikes (excesses actually) were always okay (it surely does not freeze the computer). With an upper limit the actually lint times (t.i. wait time in the queue plus runtime of the linter) can get above 10s which was really not the better overall experience. #1234 ✔️ 

- Kill old processes. This *really* brings down CPU. 20%-50%. But this is a risky change without user feedback. #1250

kaste:
The debouncer is actually the central station to reduce concurrency and to not overload
the CPU. If the debounce time ("delay") and process time do not "match" we grow unbounded and the queue length 💥 

So, we measure the actual runtime of each lint and adjust the debounce delay. 

Otherwise, we just have a upper limit of n tasks to run in parallel

Fixes #1201

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -3,7 +3,7 @@
     "debug": false,
 
     // When in the "background" lint mode, this value determines
-    // a delay before a request is send to the linter
+    // the minimum delay before a request is send to the linter
     "delay": 0.3,
 
     // Available gutter themes:
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -4,7 +4,7 @@
 
     // When in the "background" lint mode, this value determines
     // the minimum delay before a request is send to the linter
-    "delay": 0.3,
+    "delay": 0.1,
 
     // Available gutter themes:
     // Blueberry Cross
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -5,6 +5,8 @@ from itertools import chain, count
 from functools import partial
 import logging
 import os
+import multiprocessing
+import time
 import threading
 
 from . import util, linter as linter_module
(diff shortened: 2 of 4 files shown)


braver:
Have any tips for how to test his behavior?

kaste:
I'm not really happy with this one. What is an acceptable upper bound here for the debouncer before it becomes ridiculous? 

What is actually a slow linter? Any numbers?

braver:
If those are our questions, is this even a viable approach? If you have a slow linter, or a slow machine, or you care about making fewer cpu cycles, or whatever, shouldn’t you just use something other than “background” mode? Or adjust the delay yourself? I mean, does SL need to try to be smart here, or do we already have enough settings and do people need to adjust their setup if it doesn’t work for them? If a linter isn’t really really fast (i.e. it returns results about as fast as the delay), you generally don’t want to run it on-the-fly, so perhaps we should focus on #244?

kaste:
From @kiddick https://github.com/SublimeLinter/SublimeLinter/pull/1250#issuecomment-374921722

> So, I've installed version from https://github.com/kaste/SublimeLinter3/tree/concurrency

Need a little bit testing, for the first sight count of `pylint` processes run simultaneously have reduced.

Huh with my `--init-hook` with appended paths for import resulution `pylint` is kind of greedy.
```
SublimeLinter: sublime_linter.py:279: Linting buffer 101 took 18.05s
INFO:SublimeLinter.sublime_linter:Linting buffer 101 took 18.05s
```

kiddick:
I should say it is usable for now. Going to test `kill-old-popens` branch.

kaste:
Cool, thanks

kaste:
Okay, I would like to merge this.

braver:
@kaste the others are merged, so you can fix the conflicts and then merge this one.

Issue #1217, opened 2018-03-15
Log deprecation warnings


kaste:
Log deprecation warnings for the following:

- Declared `syntax`, `selectors`
- Declared `version_args`, `version_re`, `version_requirement`
- Declared `inline_settings`, `inline_overrides`, `comment_re`, `shebang_match`
- ~~Declared `executable`~~
- Overriden `build_cmd`
- Overriden `insert_args`
- ~~Accessed `executable_path`~~ ✔️ #1241 
- Implemented `can_lint`
- Implemented `can_lint_syntax`
- Uses `@python` in cmd

braver:
With a checklist like this, I could also go through our plugins and update them, and open issues in the 3rd party ones. I don’t really expect the maintainers or users to read depreciation warnings and to undertake action. Usually maintainers are missing and I’m pessimistic about users fixing these things (perhaps I should be more optimistic here).

Issue #1229, opened 2018-03-18
TypeError: 'NoneType' object is not iterable for mdl linter


fmasuhr:
I am getting a similar bug as in https://github.com/SublimeLinter/SublimeLinter/issues/1207 which should be solved by version SL 4.1.1. But I am still getting it with the new SL version and the [mdl](https://github.com/roadhump/SublimeLinter-contrib-mdl) linter:

```
SublimeLinter: SublimeLinter.sublime-package:71: Logging installed; log level INFO
SublimeLinter: SublimeLinter.sublime-package:62: debug mode: on
SublimeLinter: SublimeLinter.sublime-package:63: version: 4.1.1
SublimeLinter: SublimeLinter.sublime-package:292: detected syntax: markdown
SublimeLinter: #1 SublimeLinter.sublime-package:747: 'mdl' is linting 'README.md'
SublimeLinter: #1 __init__.py:1273: ERROR: Linter crashed.

Traceback (most recent call last):
  File "lint.backend in /Users/fmasuhr/Library/Application Support/Sublime Text 3/Installed Packages/SublimeLinter.sublime-package", line 97, in execute_lint_task
  File "lint.linter in /Users/fmasuhr/Library/Application Support/Sublime Text 3/Installed Packages/SublimeLinter.sublime-package", line 759, in lint
  File "lint.linter in /Users/fmasuhr/Library/Application Support/Sublime Text 3/Installed Packages/SublimeLinter.sublime-package", line 469, in get_cmd
TypeError: 'NoneType' object is not iterable
```

As the mdl linter is still using the deprecated `executable_path` it is using the not iterable `NoneType` here. It is already failing some lines before the introduced fix in PR #1209.

kaste:
Hm, that linter is unmaintained. (@braver opened an issue there a few months ago without a response.) So, the fix is easy. But if I open a PR, it doesn't help if nobody merges.

braver:
Hmm, @fmasuhr please give the maintainers a notice and if they don’t respond we’ll take over the plugin.

braver:
Opened an issue wrt transfer https://github.com/roadhump/SublimeLinter-contrib-mdl/issues/3

fmasuhr:
@braver Thank you for taking care of that! I hope you can take over the plugin

kaste:
Fix is in #1241

fmasuhr:
@kaste Thanks a lot for the fast implementation here

[...]

PR #1236, opened 2018-03-18
Iterate on logging

### LINKED ISSUE #1204: Show the error panel once per same error per session/window/view
To reduce annoyance, otherwise every time we do a stupid error, we render SL useless bc the same error would pop up after every keystroke.

kaste:
- Show the same error once per window per session
- Beautify error message
- Inject current view and use its window for logging
- Early install std handler to allow logging at import/startup time

Fixes #1204 
Prepare #1217


braver:
I’ll try to test this and #1234 tomorrow morning and if see nothing weird I’ll merge and release. Sound like a plan to you?

kaste:
I want a fix for #1229

braver:
That’s in a plugin though, right?

PR #1242, opened 2018-03-19
Fixup log handler init. Prevent double logging.

kaste:


braver:
I don’t understand enough of this to say anything useful here.

PR #1247, opened 2018-03-19
fix #1125 Friendly configuration for icons

### LINKED ISSUE #1125: Icons in custom gutter theme
I've created custom gutter themes for https://www.monokai.pro and they worked well for SL3.

In SL4, it seems custom error / warning icons are gone, while it does pick up the right `colorize` value from the `gutter-theme` file. Looking through your code, I think this is the part where custom icons are resolved:

```python
def get_icon(f):
    def wrapper(*args):
        res = f(*args)
        key = args[1]

        if not res:
            logger.error("Styles are invalid. Please check your settings and restart Sublime Text.")
            return

        if key != "icon":
            return res
        else:
            if res in ("circle", "dot", "bookmark", "none"):  # Sublime Text has some default icons
                return res
            elif res != os.path.basename(res):
                return res
            else:
                theme = persist.settings.get('gutter_theme')
                return 'Packages/SublimeLinter/gutter-themes/{}/{}.png'.format(theme, res)

    return wrapper
```

I think this part `'Packages/SublimeLinter/gutter-themes/{}/{}.png'.format(theme, res)` shouldn't be hardcoded, but the base folder should be the same folder as where the `gutter-theme` file is found (if defined via the `gutter_theme` setting).

braver:
- make it less easy to make little mistakes
- make third party gutter themes easier to use

Fixes #1125

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -122,8 +122,9 @@
     // - codes is an array which can contain error codes provided by a linter.
     // - icon can be:
     //   one of the default icons: "circle", "dot" or "bookmark"
-    //   or a path to an icon file like (e.g. "Packages/icons/warning.png")
     //   or "none" to remove the icon
+    //   or a path to an icon file like
+    //      "Packages/SublimeLinter/gutter-themes/Blueberry Cross/error.png"
     //   or one provided by a gutter theme (e.g. "warning" or "error")
     //      in theme Default: "warning", "error", "cog", "x",
     //      and "diamond", "heart", "pointer", "square", "star", "triangle" all
--- a/lint/style.py
+++ b/lint/style.py
@@ -65,6 +65,9 @@ class HighlightStyleStore:
                     return res
                 elif res != os.path.basename(res):
                     return res
+                elif persist.settings.get('gutter_theme').endswith('.gutter-theme'):
+                    theme_path = os.path.dirname(persist.settings.get('gutter_theme'))
+                    return '{}/{}.png'.format(theme_path, res)
                 else:
                     theme = persist.settings.get('gutter_theme')
                     return 'Packages/SublimeLinter/gutter-themes/{}/{}.png'.format(theme, res)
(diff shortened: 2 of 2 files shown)


kaste:
Requesting @Monokai for review bc the python code itself is simple but I don't have any themes to do a functional test.

Monokai:
@kaste Already commented here: https://github.com/SublimeLinter/SublimeLinter/issues/1125#issuecomment-374529061

kaste:
Okay, thanks. Explicit is the base implementation, anything more convenient (automatic) must go on top of that. I like the failing icon we have right now.

So, what the code does.

User sets `theme: a/b/c.gutter-theme`, icon to 'warning' => resolved icon 'a/b/warning.png'
User sets `theme: a/nothere/c.gutter-theme` icon to 'warning' => resolved icon 'FAIL'
User sets `theme: a/b/c.grutterteme`, resolved icon 'FAIL'

So, a more convenient experience can be e.g. `theme: c` -> searches for `c.gutter-theme` or FAIL

braver:
> So, a more convenient experience can be e.g. theme: c -> searches for c.gutter-theme

We provide that experience for our built-in themes. You set it to `'Danish Royalty'` and we understand that to mean `'Packages/SublimeLinter/gutter-themes/Danish Royalty/Danish Royalty.gutter-theme'`. 

We could use `find_resources` instead (and that's basically what @Monokai proposes too) and let Sublime Text figure out where `Danish Royalty.gutter-theme` hangs out. However, that means that if `<Some Package Earlier In The Alphabet Than SublimeLinter>` has a `Danish Royalty.gutter-theme` it will try to use that and maybe fail because who knows what that is. Now, this is perhaps unlikely, but we get enough people asking us to debug their computer for them and debugging style things is about the least exciting thing I can think of. 

We could also first try to see if `'Danish Royalty'` is a theme we know, and if not then look for it using `find_resources`. However, as I noted in my reaction in #1125, how do you know to put `'Danish Royalty'` in that setting in the first place? You probably copied that from somewhere. Why do we have to go looking for it if you (or the provider of said gutter theme) knows exactly where it is?

kaste:
I just don't see that we need more than this. 🚢

braver:
> I just don't see that we need more than this.

Same. 🤘

PR #1249, opened 2018-03-20
Implement kill_switch for a better upgrade experience

### LINKED ISSUE #1239: Implement kill-switch to trigger before/after an update
Upgrading is still terrible. Bc SublimeLinter basically runs in an undefined state, everything throws, and our message 'Please restart' actually never displays bc it gets overwritten by all the other errors.

Before the update

- empty queue
- mark all views inconsistent (to abort all running tasks)
- empty linter_classes
- ~~update errors similar to what we do in `get_linters_for_view`~~

~~Refactor: move `view_linters` from persist to sublime_linter~~

Generally, the app state in `persist` stays as is during upgrade, also the old code is still running. (Esp. all threads see the old code.)

If we do not touch/clean `persist.errors` and `persist.view_linters` on `unload`, we still have a chance of a slow automatic cleanup after `load`. If `view_linters` survives an upgrade we can remove all highlighted regions and the gutter marks later. (Why? I'm unsure if it is advisable to do lot of work in `unload`. Some cleanup code could be async/threaded.)

kaste:
Fixes #1239

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -48,16 +48,16 @@ def plugin_loaded():
 
     try:
         from package_control import events
-        if events.install('SublimeLinter'):
+        if events.install('SublimeLinter') or events.post_upgrade('SublimeLinter'):
             util.show_message(
-                'SublimeLinter has been installed. Please restart Sublime Text.')
-        elif events.post_upgrade('SublimeLinter'):
-            util.show_message(
-                'SublimeLinter has been upgraded. Please restart Sublime Text.')
+                'SublimeLinter has been installed or upgraded. '
+                'Please restart Sublime Text.')
+            return
     except ImportError:
         pass
 
     persist.api_ready = True
+    persist.kill_switch = False
     persist.settings.load()
     logger.info("debug mode: on")
     logger.info("version: " + util.get_sl_version())
@@ -71,6 +71,16 @@ def plugin_loaded():
 
 
 def plugin_unloaded():
+    try:
+        from package_control import events
+
+        if events.pre_upgrade('SublimeLinter') or events.remove('SublimeLinter'):
+            logger.info("Enable kill_switch.")
+            persist.kill_switch = True
+            persist.linter_classes.clear()
+    except ImportError:
+        pass
+
     queue.unload()
     persist.settings.unobserve()
 
@@ -234,22 +244,26 @@ def lint(view, view_has_changed, lock):
     This method is called asynchronously by queue.Daemon when a lint
     request is pulled off the queue.
     """
+    # We call `get_linters_for_view` first and unconditionally for its
+    # side-effect. Yeah, it's a *getter* LOL.
+    with lock:  # We're already debounced, so races are actually unlikely.
+        linters = get_linters_for_view(view)
+
+    if not linters:
+        return
+
+    # Very, very unlikely that `view_has_changed` is already True at this
+    # point, but it also implements the kill_switch, so we ask here
     if view_has_changed():  # abort early
         return
 
     bid = view.buffer_id()
+    events.broadcast(events.LINT_START, {'buffer_id': bid})
 
-    # We're already debounced, so races are actually unlikely
-    with lock:
-        linters = get_linters_for_view(view)
+    next = partial(update_buffer_errors, bid, view_has_changed)
+    backend.lint_view(linters, view, view_has_changed, next)
 
-    if linters:
-        events.broadcast(events.LINT_START, {'buffer_id': bid})
-
-        next = partial(update_buffer_errors, bid, view_has_changed)
-        backend.lint_view(linters, view, view_has_changed, next)
-
-        events.broadcast(events.LINT_END, {'buffer_id': bid})
+    events.broadcast(events.LINT_END, {'buffer_id': bid})
 
 
 def update_buffer_errors(bid, view_has_changed, linter, errors):
@@ -304,6 +318,11 @@ def make_view_has_changed_fn(view):
     initial_change_count = view.change_count()
 
     def view_has_changed():
+        if persist.kill_switch:
+            logger.warning(
+                'Aborting lint. SublimeLinter needs a restart of Sublime.')
+            return True
+
         changed = view.change_count() != initial_change_count
         if changed:
             persist.debug(
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -49,6 +49,11 @@ def plugin_loaded():
     try:
         from package_control import events
         if events.install('SublimeLinter') or events.post_upgrade('SublimeLinter'):
+            # In case the user has an old version installed without the below
+            # `unload`, we 'unload' here.
+            persist.kill_switch = True
+            persist.linter_classes.clear()
+
             util.show_message(
                 'SublimeLinter has been installed or upgraded. '
                 'Please restart Sublime Text.')
--- a/lint/persist.py
+++ b/lint/persist.py
@@ -7,6 +7,7 @@ from .settings import Settings
 
 
 api_ready = False
+kill_switch = True
 
 settings = Settings()
 


kaste:
Generally, the app state in `persist` stays as is during upgrade, also the old code is still running. (Esp. *all running threads* see the old code.)

If we do not touch/clean `persist.errors` and `persist.view_linters` on `unload`, we still have a chance of a slow automatic cleanup after `load`. If `view_linters` survives an upgrade we can remove all highlighted regions and the gutter marks later. (Why? I'm unsure if it is advisable to do lot of work in `unload`. Some cleanup code could be async/threaded.)

`kill_switch` ensures all running threads/popen calls abort, and no new ones are generated. Cleaning `linter_classes` ensures that SublimeLinter core forgets immediately all registered linters. T.i. we cannot assign linters to views anymore. In turn: `view_linters` gets emptied, which in turn empties `errors`, which in turn removes all highlights etc.

kaste:
I would like this to be in the next version! Bc it acts on `unload` the user must first install this `unload` handler here, and then still wait for the next update to take effect.

sublime_linter.py
```diff
@@ -304,6 +318,11 @@ def make_view_has_changed_fn(view):
     initial_change_count = view.change_count()
 
     def view_has_changed():
+        if persist.kill_switch:
```
kaste: Maybe `persist.get('kill_switch', True)` bc otherwise we get an AttributeError here on the next upgrade?

sublime_linter.py
```diff
@@ -48,16 +48,16 @@ def plugin_loaded():
...
-        if events.install('SublimeLinter'):
+        if events.install('SublimeLinter') or events.post_upgrade('SublimeLinter'):
             util.show_message(
-                'SublimeLinter has been installed. Please restart Sublime Text.')
-        elif events.post_upgrade('SublimeLinter'):
-            util.show_message(
-                'SublimeLinter has been upgraded. Please restart Sublime Text.')
+                'SublimeLinter has been installed or upgraded. '
+                'Please restart Sublime Text.')
+            return
```
kaste: Or add `persist.kill_switch = True`?

braver:
I dunno, my brain is fried. If we depend on the previous version leaving something behind for the next version after an upgrade, we definitely need to be defensive about trying to get it.

braver:
Idea, not sure if other that what you already have here, but to put it out there.

Before upgrade, and/or before unload, we hit a switch so no new linter processes will run, we let already running processes run their course but we shut down all views etc. The only way to get it running again is to restart ST.

Again brain fried. Sorry. I want a fix too, I’m just completely no help at all here.

kaste:
What do you actually think I'm doing here. 🤹🤧

braver:
Something along those lines probably?

Want me to test or verify anything?

kaste:
You could ask your buddies on discord how to actually test an update from PackageControl.

Pushed a commit: Now we do on load what we already do on unload. This is for users that have an old `unload` handler installed. T.i. all users atm LOL (Between unload and load we're still in a somewhat undefined state for this first update.)

Okay, what I try to do here. Enable kill_switch. The kill_switch marks all views inconsistent all the time. That means that running threads abort, and new ones do not even start. Additionally, I unregister all linters by clearing linter_classes. That means all linters get unassigned from the buffers which in turn means the errors and with it all highlighted regions disappear. 

After that, you should have a clean SublimeText. Note that all our views are still running. But they act mostly on `persist.errors` which should be empty.

braver:
Ok, that sounds good. Simpler and more robust that trying to reload everything. I'll see if I can learn anything on discord.

braver:
Well, tonight "my buddies on discord" amounts to @FichteFoll. Not saying that's bad, I love @FichteFoll! However:

screen shot 2018-03-20 at 21 58 45


... sounds like a painful no-fun process. I'm not up for that at the mo. I'd much rather just merge this and break a few people's computers tomorrow. Or not if this is a good fix. Honestly, this isn't my best week, sorry.

kaste:
I would need to git push fake versions to a private channel? 🤐

Where's wbond?

Aside: the reloader is actually good, it just reliable crashes if threads are still running. So I first try to bring the old version down with this PR, and then we'll see if we can automatically reload and reactivate.

kaste:
I probably just merge tomorrow morning.

FichteFoll:
No, the process is to pack a `sublime-package` archive and construct the repository.json file manually. Like https://packagecontrol.io/prerelease/packages.json but locally on your disk.

kaste:
So basically `git archive --format=zip -o somewhere/SublimeLinter.sublime-package HEAD` and then bump the version in a somewhere/packages.json. The `url` in that packages.json points to a `file:://`. The start Sublime, PC will update. Then checkout a different branch. git archive, bump version, restart Sublime. Is that correct? Where do I register a local packages.json?

FichteFoll:
You register it in PC's `repositories` setting.

braver:
@kaste I set this up to simulate an upgrade from 4.1.1 to master, and got this after triggering the upgrade:

```
reloading plugin SublimeLinter.__init__
zipimport.ZipImportError: bad local file header in /Users/koenlageveen/Library/Application Support/Sublime Text 3/Installed Packages/SublimeLinter.sublime-package
reloading plugin SublimeLinter.busy_indicator_view
reloading plugin SublimeLinter.goto_commands
reloading plugin SublimeLinter.highlight_view
reloading plugin SublimeLinter.log_handler
reloading plugin SublimeLinter.message_view
reloading plugin SublimeLinter.panel_view
reloading plugin SublimeLinter.status_bar_view
reloading plugin SublimeLinter.sublime_linter
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 157, in reload_plugin
    m.plugin_loaded()
  File "sublime_linter in /Users/koenlageveen/Library/Application Support/Sublime Text 3/Installed Packages/SublimeLinter.sublime-package", line 57, in plugin_loaded
AttributeError: 'module' object has no attribute 'show_message'
reloading settings Packages/SublimeLinter/SublimeLinter.sublime-settings
```

It doesn't seem to reload `util`, so `show_message` doesn't work. Perhaps because the earlier error? I have no clue what I'm looking at here though. It *did* seem to effectively kill everything else, so that part works.

kaste:
A lot of people reported these ZipErrors and I have no clue. Really something you have to ask your buddies bc we don't do anything here. Our archives are broken?

The other error is a bummer. Sublime does not reload any module on upgrade regardless of the error above. And since I renamed message -> show_message this breaks. 

Should we guard it just for this update?

braver:
Yeah, the message is kinda important, perhaps we should even just implement one specifically for this right in sublime_linter.py so we don’t need an import to show it. 

The other issue I’ll ask around for.

kaste:
But actually, it doesn't throw in your face and no panel loading syntax error. That's really an achievement. 💪

braver:
It is! 👍

braver:
Other occurrences of the zip error:

- https://github.com/wbond/package_control/issues/947
- https://github.com/jdc0589/JsFormat/issues/170#issuecomment-213031454
- https://forum.sublimetext.com/t/zipimport-zipimporterror-bad-local-file-header-when-upgrading-plug-in-which-has-pyc-submodules/31629

From googling around a bit: https://trac.edgewall.org/ticket/7014

I'm asking questions on discord.

kaste:
Interesting links, but we don't do anything with sys.path afaik. The reloader would do this, but it is not running right now.

braver:
Well, the error does seem to originate from `reload_plugin `?

kaste:
Yes, but we don't call reload_plugin. That's the Sublime reloader.

braver:
cccdb4fd-7c82-4fe7-8cac-55a832a44d41

kaste:
I don't think that's actually a bad zip file here.

PR #1250, opened 2018-03-20
[WIP] Kill old lint processes before starting more

kaste:
**DO NOT MERGE** Just a proof of concept implementation!

Instead of just spawning lint processes (popen) as requested, `kill` outdated procs.

Note: Windows throws `BrokenPipeError` if we kill a proc which we catch and eat. If x systems throw something different, we will need to adjust the `except` clause in util.

Note: Only implemented for STDIN linters.

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -192,7 +192,9 @@ def combine_output(out, sep=''):
     return ANSI_COLOR_RE.sub('', output)
 
 
-def communicate(cmd, code=None, output_stream=STREAM_STDOUT, env=None, cwd=None):
+def communicate(
+        cmd, code=None, output_stream=STREAM_STDOUT,
+        env=None, cwd=None, bid=None):
     """
     Return the result of sending code via stdin to an executable.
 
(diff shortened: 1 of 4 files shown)


kiddick:
So, I've installed version from https://github.com/kaste/SublimeLinter3/tree/concurrency

Need a little bit testing, for the first sight count of `pylint` processes run simultaneously have reduced.

kaste:
@kiddick Moved your comment to #1234 For your issue (really slow linter) this PR (branch `kill-old-popens`) would be more interesting.

lint/util.py
```diff
@@ -216,14 +218,27 @@ def communicate(cmd, code=None, output_stream=STREAM_STDOUT, env=None, cwd=None)
     else:
         stdout = stderr = None
 
-    out = popen(cmd, stdout=stdout, stderr=stderr,
-                output_stream=output_stream, env=env, cwd=cwd)
+    proc = popen(cmd, stdout=stdout, stderr=stderr,
+                 output_stream=output_stream, env=env, cwd=cwd)
+
+    if proc is not None:
+        from . import persist
```
nvtkaszpir: I got weird feeling it should be on top of the file...

kaste: I just wrote this down very quickly, and surprisingly it worked on Windows, so I thought I push it to get some feedback.

kaste:
With this PR you basically (most of the time) have one process per view max bc we kill all old processes before launching new ones. It is unlikely that you edit different views faster than one lint process actually takes.

We could even kill processes before the debouncer bc as soon as we 'hit' the debouncer we already know that the current running process will be aborted and its result will be thrown away.

nvtkaszpir:
Right now I can have around 3 linters to single file for ansible:
- yamllint
- ansible-lint (slow)
- ansible-review

nvtkaszpir:
Looks like I'm getting SublimeText freezes with that change, where plugin_host is spawned mutiple times and eating cpu cores, but one is always in around 140% (I've got 4 cores, so its 1 and a half core in that case).
Ubuntu 16.04

kaste:
Did you see this after a restart of Sublime?

nvtkaszpir:
Yes.

kaste:
Oh. I see I used kill() here. You could try the friendlier terminate().

lint/util.py
FichteFoll: It doesn't matter where imports are located in most cases.
The convention is to put them at the top so they can be used throughout the entire module, but sometimes late imports might be necessary to prevent circular imports.

I didn't check whether that is the case here.

nvtkaszpir:
I've been running with that fixed PR for about 5 days now. So far so good, no freezes, everything works.
I'll probably be able to test it more thoroughly in upcoming 3 days.

kaste:
Okay, the deployment strategy could be to ship this behind a feature flag (`backend.kill_old_processes`) set to false (opt-in). Later we flip that to true (opt-out). And if everything works out nicely, we could then remove that flag. 

I could work on this one after #1284 bc it has a good refactoring.

PR #1254, opened 2018-03-20
Add a tooltip-on-cusor-move feature

bhuga:
Adds a feature to display tooltips based on cusor position. The preference is disabled by default.

I wanted this for me, but I thought you might want this in the main project as well. I can rework it if there's something you'd like to see before landing it. Or if you decide it's not something you want, no hard feelings.

linter video

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -83,6 +83,9 @@
     // Show a report for problems on a region by hovering over it.
     "show_hover_region_report": true,
 
+    // Show report for all problems on the cursor's line
+    "show_cursor_line_report": false,
+
     // Highlight problems in the minimap.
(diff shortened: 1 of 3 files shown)


FichteFoll:
I'm pretty sure this is not coming back. The functionality is now provided by the panel (<kbd>SublimeLinter: Show All Errors</kbd>) in a much less intrusive way.

braver:
Indeed, not gonna merge this. You can open the line report with a simple key binding. If you want to look at your errors all the time, just open the panel (it syncs with your cursor position).

bhuga:
👌

kaste:
Fwiw I have something similar though as a plugin not as a patch. Still not in core. 🕺

PR #1260, opened 2018-03-22
put killswitch and restart message in sublime_linter

braver:
@kaste from #1249

What seems to happen during an upgrade and then reload is that new parts of SL are talking to old parts of SL. We can try to keep things consistent between versions, but even simple renames can break this. We also cannot be sure that users aren't skipping versions, so we cannot deal with this effectively. 

I think we need to make sure that what happens in`events.post_upgrade` doesn't rely on anything outside `sublime_linter.py`, as much as possible. 

- This PR adds a `restart_message()`, because panels are really simple anyway so we don't lose much here by not reusing the other panel.
- Additionally, `kill_switch` isn't used by other parts of SL, so if we don't put it in persist that will perhaps save us some headaches next time we want to refactor this.

### DIFF
--- a/lint/persist.py
+++ b/lint/persist.py
@@ -7,7 +7,6 @@ from .settings import Settings
 
 
 api_ready = False
-kill_switch = True
 
 settings = Settings()
 
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -18,6 +18,7 @@ from .lint import reloader
 
 
 logger = logging.getLogger(__name__)
+kill_switch = False
 
 
 def backup_old_settings():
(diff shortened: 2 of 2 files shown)


kaste:
I don't think putting kill_switch in sublime_linter does even work. Did you test this change?

braver:
No, I didn't, just a suggestion. Do you think this is a good idea at all? Perhaps we should just make sure that the message shows.

kaste:
The kill_switch PR was too hard to make it work. It is basically really too hard to reason about something like this multiple times a week. So, since it is working, we need to be conservative and just fix the one bug with `util.show_message`. I hope #1261 works. 🤞

PR #1261, opened 2018-03-22
Fire delayed event to show 'restart message'

kaste:
Without the delay, even if the command was already registered and
getting executed, the panel just didn't show up.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -54,9 +54,7 @@ def plugin_loaded():
             persist.kill_switch = True
             persist.linter_classes.clear()
 
-            util.show_message(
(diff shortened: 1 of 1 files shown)


kaste:
Another variant would be, to show the restart message from within `message_view.py`. T.i. we would implement a `plugin_loaded` handler there which does basically the same thing.

braver:
> Another variant would be, to show the restart message from within message_view.py

That sounds really good to me actually.

kaste:
There is no advantage in moving this around for now, and I still had to call `timeout` to make it work btw.

braver:
Ok, let’s go with this then

kaste:
🤞

PR #1264, opened 2018-03-22
Do not invalidate regions under cursor for quiet views

### LINKED ISSUE #1262: With highlights toggled off, whitespace highlights are still flashed
- have an underline style
- toggle highlights off
- make whitespace errors
- notice temporary flashes of fallback outlines around whitespace errors

kaste:
Fixes #1262

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -22,9 +22,11 @@
     "gutter_theme": "Default",
 
     // EXPERIMENTAL: See below `demote_while_editing`.
-    "highlights.time_to_idle": 5,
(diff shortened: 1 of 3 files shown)


kaste:
Also added `demote_all`. This is an interesting pattern. If I set `demote_all` and idle_time to 500 I basically have a quiet view where on save all errors are highlighted. `demote_all` and a very short idle time makes it a bit quiet so it is unlikely you get these parse error popups while writing the if statement. Or: empty block. Dude, I was actually about to hit enter here.

kaste:
I'll just merge bc it is a trivial bug fix.

braver:
👍

PR #1266, opened 2018-03-23
Implement environment not ready workaround for osx

### LINKED ISSUE #1202: Implement new work-around for OSX-environment-not-ready-users
(This is fun to implement too.)

kaste:
Fixes #1202

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from functools import partial
 import logging
 import os
+import time
 import threading
 
 import sublime
@@ -270,6 +271,10 @@ def lint(view, view_has_changed, lock):
     if view_has_changed():  # abort early
         return
 
+    if not environment_is_ready():
+        hit(view)
+        return
+
     bid = view.buffer_id()
     events.broadcast(events.LINT_START, {'buffer_id': bid})
 
@@ -345,3 +350,44 @@ def make_view_has_changed_fn(view):
         return changed
 
     return view_has_changed
+
+
+ENV_READY_TIMEOUT = 60
+ENV_IS_READY_STATE = {
+    'ready': False if sublime.platform() == 'osx' else True,
+    'start_time': None,
+    'first_path': None
+}
+
+
+def environment_is_ready():
+    ready = ENV_IS_READY_STATE['ready']
+    if ready:
+        return True
+
+    path = os.environ['PATH']
+    if '/local/' in path:
+        logger.info('Env seems ok: PATH: {}'.format(path))
+        ENV_IS_READY_STATE['ready'] = True
+        return True
+
+    start_time = ENV_IS_READY_STATE['start_time']
+    if start_time is None:
+        ENV_IS_READY_STATE['start_time'] = time.time()
+    elif (time.time() - start_time) > ENV_READY_TIMEOUT:
+        logger.warning('Env timeout: PATH: {}'.format(path))
+        ENV_IS_READY_STATE['ready'] = True
+        return True
+
+    first_path = ENV_IS_READY_STATE['first_path']
+    if first_path is None:
+        logger.info('Env first path: PATH: {}'.format(path))
+        logger.info(
+            'Wait until env is loaded but max {}s.'.format(ENV_READY_TIMEOUT))
+        ENV_IS_READY_STATE['first_path'] = path
+    elif path != first_path:
+        logger.info('Env path loaded: PATH: {}'.format(path))
+        ENV_IS_READY_STATE['ready'] = True
+        return True
+
+    return False


braver:
Of course, these days my environment is loaded so fast I can't reproduce the problem this is supposed to fix. It does seem to work though, from just looking at the debug log.

FichteFoll:
This definitely looks like the right approach.

broox:
i'm seeing the following errors in my SublimeLinter console:
```
SublimeLinter: #1546 eslint checkins.js ERROR:

 env: node: No such file or directory
```

how can i help test this PR?

kaste:
* Enable debug mode for SublimeLinter
* Uninstall SublimeLinter
* Exit Sublime
* `git clone https://github.com/kaste/SublimeLinter3.git SublimeLinter` into your Packages folder
* `git co osx-env-workaround`
* Start Sublime, open a js file
* Open View > Console to see it in action

broox:
done. seeing the same thing ☹️ 

```
SublimeLinter: #3 eslint microblog.js ERROR:
============================================

env: node: No such file or directory
```

kaste:
Please look at and post parts of the console log.

broox:
In the sublime console
```
SublimeLinter: WARNING: Env timeout: PATH: /usr/bin:/bin:/usr/sbin:/sbin
```

In iTerm
```
$ which node
/usr/local/bin/node

$ python
Python 2.7.10 (default, Feb  7 2017, 00:08:15)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ['PATH']
'/usr/local/mysql/bin:/usr/local/mysql/bin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin'
```

kaste:
The timeout is after 60 seconds! So I think you have real issue here. Of course, since you have the code checked out you can try a different timeout. Restart Sublime and stare at the console if it eventually resolves positively.

broox:
yea, it seems like if it can't load the path within a minute, it's probably not gonna load. have we tried using: https://github.com/codexns/shellenv ?

kaste:
@broox Ensure you have the latest Sublime installed. Basically Sublime runs a login shell (like `bash -l`) to get a good env here. Something like that should not timeout. The PR here is about avoiding the startup race, what you get is a complete failure of finding a good env to work with.

broox:
@kaste I'm running **Sublime Text 3.0 (3143)** on **Mac OS 10.12.6**

Note: When I initially installed SublimeLinter and eslint, everything seemed to work until I rebooted my machine. Sublime has been failing to load the path ever since (I've rebooted several times since, as well).

braver:
Do you have any other packages installed?

kaste:
There's also setting iirc in Sublime where you can turn this off. So double check here.

Otherwise, if you have nvm, hack it so it doesn't load automatically.

broox:
In my default sublime settings, the relevant pieces I have are:

```
"shell_environment": true,
"show_full_path": true,
```

None of this is overridden by user settings.

The only packages are have installed are:

* Package Control v3.3.0; packagecontrol.io
* SublimeLinter (this PR's branch)
* SublimeLinter-eslint v4.0.5; github.com/SublimeLinter/SublimeLinter-eslint
* SublimeLinter-flow v4.5.0; github.com/SublimeLinter/SublimeLinter-flow

LeEnno:
tl;dr: works for me

Just a few things:

1. The [instructions by kaste](https://github.com/SublimeLinter/SublimeLinter/pull/1266#issuecomment-375907208)  miss a `cd SublimeLinter`. Also `co` means `checkout`. Nitpicky, I know, but some users may want to test and be confused about it not working they way they expect.
2. When restarting Sublime after checkout of `osx-env-workaround` branch, I was told that a missing dependency has been installed and I should restart Sublime again. Hope this is expected.

But finally no more missing `node` shizzle 👌

kaste:
We should merge this @braver, it's a finite state machine.

braver:
Jup, sounds good.

zertosh:
One issue I'm having with this is that since I keep `shell_environment` set to `false`, I'm forced to always wait for the 60s timeout before being able to use the linters

kaste:
Our naive heuristic just waits for `/local/` in PATH. So you're not generally forced to wait 60s. (E.g. users should not wait if they start subl from the command line.) So, how does your 'good' PATH look like, or can you share the debug log here?

braver:
Perhaps not too clear from documentation, but `paths` is only used to find the executables, it doesn't affect the env they run in. If your executable has other dependencies that's not enough and you need to path env as well, as you did.

kaste:
We have an issue to deprecate paths. And of course as soon as you want to deprecate something, nobody wants to write obsolete documentation. 🙄

PR #1267, opened 2018-03-23
Ensure a min region length of 1

### LINKED ISSUE #1168: No highlight using outline style when linter reports only line number
@kaste I get underscores with pydocstyle. 

screen shot 2018-03-08 at 17 31 51

This is with stippled_underlines. With outlines it draws nothing at all. You can basically run any py file in SL with pydocstyle to reproduce.

Example pydocstyle (@ master) output (it only reports line numbers):
```
SublimeLinter: #39 linter.py:809: pydocstyle output:
    /var/folders/cp/659wm5bs1pl2kpz8j9jvckrw0000gn/T/tmpf_z1f2.py:1 at module level:
            D100: Missing docstring in public module
    /var/folders/cp/659wm5bs1pl2kpz8j9jvckrw0000gn/T/tmpf_z1f2.py:10 in public class `TooltipController`:
            D101: Missing docstring in public class
    /var/folders/cp/659wm5bs1pl2kpz8j9jvckrw0000gn/T/tmpf_z1f2.py:28 in public class `SublimeLinterLineReportCommand`:
            D101: Missing docstring in public class
    /var/folders/cp/659wm5bs1pl2kpz8j9jvckrw0000gn/T/tmpf_z1f2.py:29 in public method `run`:
            D102: Missing docstring in public method
    /var/folders/cp/659wm5bs1pl2kpz8j9jvckrw0000gn/T/tmpf_z1f2.py:88 in public function `join_msgs`:
            D103: Missing docstring in public function
    /var/folders/cp/659wm5bs1pl2kpz8j9jvckrw0000gn/T/tmpf_z1f2.py:132 in public function `get_current_pos`:
            D103: Missing docstring in public function
```

kaste:
Fixes #1168

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -354,6 +354,10 @@ def prepare_highlights_data(view, linter_name, errors, demote_predicate):
     for error in errors:
         line_start = get_line_start(view, error['line'])
         region = sublime.Region(line_start + error['start'], line_start + error['end'])
+        # Ensure a region length of 1, otherwise we get visual distortion:
+        # outlines are not drawn at all, and underlines get thicker.
+        if len(region) == 0:
+            region.b = region.b + 1
 
         scope = get_scope(**error)
         mark_style = get_mark_style(**error)


braver:
This works, but shouldn't we remove the `no_column_highlights_line` setting now that its reason d'être has been resolved?

kaste:
No, `no_column_highlights_line` is a setting that marks the *whole* line if the linter does not output a `col`. If it is false we mark the first `col`.

PR #1269, opened 2018-03-24
Ensure you hover over something

### LINKED ISSUE #1268: Line Hover issue
I have gutter "icon": "none", set for my warnings, however entering the region still shows the line report about it (show_hover_line_report = true), while i think it should only do that when the gutter icon is visible aka not set to "none".

kaste:
Fixes #1268


braver:
Don’t forget the status bar.

kaste:
Now, you're just in a bad mood today. And still escalating. The 'quiet views' toggle is not a 'zen mode'. It disables the highlighted regions and then popping up a tooltip over a random word is a ux bug. It would be on any random website. 

The status bar is as the name suggest a completely different view component.

braver:
Well sorry about the mood. I’m not talking about quiet views. I’m talking about setting the highlight style to “none”. The region should then no longer trigger tooltips, but also not show in the status bar. Or am I incorrect here?

kaste:
The highlight style is not generally filtering errors. We show them still in the panel and of course on the status bar. 

I understand the issue here as a mouse interaction thing. As in: you must hover over something visually promoted to expect a tooltip. 

If you hover over 'nothing' in the gutter you don't expect a tooltip. I saw this issue in quiet mode first, where you leave the mouse cursor accidentally over a random word and then the tooltip pops.  

(Like on a website: Say you implement a read mode where you de-highlight all links. Of course, it is not enough to just de-emphasize all links. You have to make them dysfunctional/pure text.)

braver:
Well to me the status bar is weird too here. Without the highlight, is it clear what the message is about?
Also, should goto jump to a region even if it’s not visually highlighted? You won’t be able to predict where it will jump.

So to me all that is related and needs to be considered holistically.

“Silencing” the tooltips in quiet mode is an obvious win of course and needs to be done. It’s the other stuff I’d like to have a clear vision about and mine is seriously cloudy right now.

kaste:
(The goto command should enable the highlighting, I thought about this during last week; either we just exit quiet mode; at least we highlight the jump target; or flash (on-off) the highlights)

kaste:
(Ah, I have status bar messaging off bc I can't fucking read it. I use mostly quiet mode with the panel, which does feel completely consistent. You have a quiet editing area, but not the illusion of no errors, which would be more like a zen-mode.)

kaste:
FWIW I think it is wrong to show the messages in the status bar bc it is highly volatile info based on the current cursor pos. It is *not* the 'status' of the buffer. 

I think it is completely consistent to have a error counter in the status bar even when the view is 'quiet'. Actually I would not event need the counter, I basically *need* a red/yellow/green indicator. Red (error) will probably not compile/work, yellow (warning) can commit and test but not ready for master/PR, and green. (It is usually completely irrelevant here if I have 1 or 10 warnings; bc only zero is allowed. It is okay for dev happiness so you can say to yourself: Okay only 3 warnings left.)

braver:
> The goto command should enable the highlighting, I thought about this during last week; either we just exit quiet mode;

In the current state of the PR, I can still hover over an invisible region (invisible because I set `mark_style` to `none`).

kaste:
Oh. I didn't know that `mark_style = None` is a thing. 😄 I thought `scope = ''` would be the way to hide highlights. (Of course I also forgot that. Basically the highlights_view I wrote does not allow to not draw highlights bc I didn't knew this is specced/user configurable feature.)

braver:
Now that I think about it some more, again comes the question: if you set `mark_style` to `none`, will `goto` still jump to that region?

kaste:
I pushed a commit to skip `mark_style = none` or `scope = ''`. goto will jump to that position bc it shows up in the panel.

kaste:
Ah, well, if you're allowed to *hide* highlighted regions (again I thought you can only hide gutter marks) then highlighting the jump positions for goto becomes a bit trickier.

If I demote such regions during initial draw (hide them) I *need* a style and scope if I must highlight them when showing the jump positions.

braver:
See? We can’t cover this topic in 20 minutes. 😉 perhaps we don’t need to try to do so in one iteration, removing more stuff from quiet mode is a good start. But all the other kinds of “hiding” we need to be smart about. Can we limit this to quiet mode, so we can merge and then continue the discussion of how to deal with hidden highlights and errors?

kaste:
I actually only wanted to ensure we don't show hover tooltips over invisible regions in this PR. Just like the PR title `Ensure you hover over something`. Maybe I lost it, but I don't think I do anything else here.

kaste:
T.i. only the change in tooltips_view is relevant, user visible https://github.com/SublimeLinter/SublimeLinter/pull/1269/files#diff-7642a51d3fcb8df8b88b6490a4b8990d

braver:
Ok, that’s a good scope to focus on too 👍🏻

braver:
OK, this is 👍 👍 for me, let's merge and 🚢

kaste:
I must avoid the separate `regions_utils.py`. I probably merge tooltips and highlights into one file. I think if we deploy like that, it would break bc of these stupid upgrading issues we have with PackageControl and Sublime.

kaste:
The feature is okay.

braver:
Ok, sure, let me know when you've refactored then.

kaste:
Okay.

braver:
Excellent, LGTM.

PR #1274, opened 2018-03-27
Show active linters in the status bar

### LINKED ISSUE #1205: Show assigned linters in the status bar per view

kaste:
Fixes #1205

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -108,6 +108,8 @@
     // Set to "" to display nothing
     "statusbar.messages_template": "{message}",
 
+    "statusbar.show_active_linters": true,
+
     // Global styles for all linters.
     // - mark_style options:
     //   "none"
--- a/active_linters_view.py
+++ b/active_linters_view.py
@@ -22,7 +22,7 @@ def plugin_unloaded():
             view.erase_status(STATUS_ACTIVE_KEY)
 
 
-@events.on(events.LINT_END)
+@events.on(events.LINT_RESULT)
 def redraw_bid(buffer_id, **kwargs):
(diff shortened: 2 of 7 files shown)


braver:
If I disable and then re-enable a linter, it doesn't come back in the status bar. Tested using an html file with embedded styles and js, running the styelint, tidy and eslint linters and then disabling eslint temporarily. Eslint does run and highlight errors when it has been re-enabled, but isn't listed in the status bar anymore.

Also, this needs to be behind a setting.

Otherwise: 🌶 Never thought I'd find this useful but it's a really nice reminder of what's running actually. Just need more space in my status bar. Or a bigger screen. Or smaller pixels + better eyes.

kaste:
That's how it is if you code with your kids 😵🤗

braver:
I code with my 🐈 , does that count?

kaste:
Only Garfield like 🐈 

I had to rewrite the assign linters to view code. Also how we get the linter settings. 

An open question is what exactly an 'active' linter is. E.g. I currently have the 'php' linter installed without actually having any 'php' on my computer. So the process is SL assigns 'php' to any 'html' file and 'deactivates' it (silently fails) bc no executable can be found. Net result: it will not show up in the status bar. Probably fine. 

It could also tell you of course: Hey you're running the 'php' linter all the time. What a waste. You should disable it. So, maybe it would be better when 'php' would actually show up as active linter.

braver:
Hmmmmmmmm 🤔 

I think I would like it to display all linters that are active, not just assigned. If I expect php to be active, but it isn't for some reason, I'd want to debug that. Ie. if a linter is listed in the status bar I expect it to be able to detect errors.

kaste:
Closing in favor of #1283

PR #1278, opened 2018-03-28
Automatically close panel after save if no errors

### LINKED ISSUE #1258: Automatically hide panel when no errors
I'm currently using `"show_panel_on_save" : "view"` and `"lint_mode" : "load_save"`, which works great, showing the panel when there are errors when I save.

After I fix the errors though, the panel remains with just a message `"No lint results."`

I then have to press 'Escape' to remove the panel.

It would be great to introduce an option `"hide_panel_on_no_errors" : true` that would automatically hide the panel when the errors are resolved.

kaste:
Fixes #1258 

For this one the panel stays open with 'No lint errors' until you save.

IMO this is the less distracting/quieter implementation bc the panel (usually immediately) closes with a keystroke `<ctrl+s>`.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -26,7 +26,8 @@ OUTPUT_PANEL_SETTINGS = {
 State = {
     'active_view': None,
     'current_pos': (-1, -1),
-    'just_saved_buffers': set()
+    'just_saved_buffers': set(),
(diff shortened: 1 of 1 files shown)


braver:
Yes, I agree. I'm 👍 on this one and definitely 👎 on the other.

PR #1279, opened 2018-03-28
Automatically close panel on lint end if no errors

### LINKED ISSUE #1258: Automatically hide panel when no errors

kaste:
Fixes #1258 

For this one the panel closes immediately if you fixed the last error.

Can flicker if you have auto-fixers running on save. Bc it will first show up
the panel bc errors, then the auto-fixer runs, then the panel hides bc no errors.

I made the other PR bc of this noise and bc it felt a bit distracting maybe 
bc the panel closes after the lint process, so between fixing the error and visually closing the panel you have a significant gap.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -26,7 +26,8 @@ OUTPUT_PANEL_SETTINGS = {
 State = {
     'active_view': None,
     'current_pos': (-1, -1),
-    'just_saved_buffers': set()
+    'just_saved_buffers': set(),
+    'panel_opened_automatically': set()
 }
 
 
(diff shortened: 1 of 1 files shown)


kaste:
Okay after 15 minutes I'm 👎 on this one. 

If I save depending on the env I can expect some actions at a distance: like prettier runs, unittest run automatically or webpack recompiles. All this takes a second or more or less, but if the panel just disappears while editing, it is visual noise. The panel is huge. 1/3 of the screen changes.

braver:
Ok, that doesn't surprise me. I'll perhaps try it later too, but if you already don't like it then let's not do it.

PR #1283, opened 2018-03-29
Alt take on linter names in the status bar

### LINKED ISSUE #1205: Show assigned linters in the status bar per view

kaste:
Better prototype than talk.

Instead of our we counter, show something like `annotations(ok), flake8(2|12), php(erred)` in the status.

Obviously on top of #1274 
Fixes #1205

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -95,10 +95,9 @@
     // Set to "never" to disable this feature.
     "show_panel_on_save": "never",
 
-    // Display counters in the status bar.
-    // {warning} and {error} will be replaced with the actual numbers.
-    // Set to "" to display nothing
-    "statusbar.counters_template": "W: {warning} E: {error}",
+    // DEPRECATED: If you've modified this in your user settings,
+    // remove it now to avoid future breakage.
+    "statusbar.counters_template": "",
 
     // Show the messages for problems at your cursor position
     // {message} will be replaced by the actual messages.
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -108,6 +108,8 @@
     // Set to "" to display nothing
     "statusbar.messages_template": "{message}",
 
+    "statusbar.show_active_linters": true,
+
     // Global styles for all linters.
     // - mark_style options:
     //   "none"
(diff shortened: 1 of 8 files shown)


braver:
Dude I need this in my life now.

kaste:
🙋 🙇

braver:
Ok, this looks really good, but it raises a question:

What do we do with the `counters_template` setting? Do we need such configurability anymore? Or should we adapt it to apply to each linter's counter? I don't see the need for a setting here, so that would mean we drop the `counters_template`.

kaste:
I like this a lot. If we go with `flake8(erred)` we could demote some warnings bc we then have a subtle or small, stable visual marker. E.g. we don't need to flash the status bar with  `WARNING: 'eslint' could not find 'eslint'` (LOL). This was never really readable anyway bc the standard Sublime flashing is way too fast to read anything. 

We can probably 'solve' the eslint config error that way. We just do `eslint(erred)` and have both a visual indicator for that this linter is still running all the time (WHAT A WASTE!) and an unobtrusive indicator that it is failing. (Instead of throwing in your face.)

drop `counters_template` of course bc we don't need it.

braver:
👍 

Minor thing: `erred` sounds a bit unnatural to me, but I'm not English so it may just be me.

kaste:
The refactoring is also 👍 

Before we had in the code literally `eslint deactivated. Could not find 'eslint'`. Something like that. But in reality it was not 'deactivated' the lint task failed.

With the refactoring we have a smaller code that enables/assignes linters to views. And then after that everything else is an error of the lint task.

kaste:
I deprecated `counters_template`. Users which may have set this in their user settings, might notice that their status bar looks different now, will maybe look in the settings again, and see this message. 

TODO: If the popen call fails MUST call `linter.notify_failure()`

I'm lazy, and it would be nice if I could introduce this after #1284 bc you cannot do this without merge conflicts.

Ready to merge.

braver:
Ok, I'll merge #1284 (wanna test this first) and then this, and perhaps add a message about the new counters.

kaste:
Do NOT release! Just the master, okay.

This is one is ready, #1284 must be tested with save-only (real-file) linters, stderr only and stdout only linters. I could not find a STDOUT only linter. Otherwise I had no errors on Windows. (And the hacks I removed were all `os.name == 'nt'` hacks.)

PR #1291, opened 2018-04-04
Better delay

### LINKED ISSUE #1285: Don't check single files more than once per time
Hello, I often work on a file openned in several buffers (side by side edit). And I noticed that when I save file it spawns two linters - one per buffer. But it is all the same file (and often this file is REALLY big). This leads to huge CPU usage and as results lags and noise from coolers.

kaste:
Fixes #1285

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -165,7 +165,7 @@ class BackendController(sublime_plugin.EventListener):
         if not util.is_lintable(view):
             return
 
-        hit(view)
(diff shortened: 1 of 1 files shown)


braver:
Makes sense. I don’t really have the time to dig in and see if I notice any change in behavior.

braver:
Oh, makes even more sense and answers a question I had in the back of my mind.

Issue #1290, opened 2018-04-03
Deprecate and remove magic appending the filename or tempfile name to `cmd`


kaste:
If `code` is None (which means we want to lint an actual file, the original or a temp file) SL3 magix append the filename to `cmd` (if there is nor marker for it). 

https://github.com/SublimeLinter/SublimeLinter/blob/6d07b3653b14deadb824ecbb48c03ee9bacb3d1b/lint/linter.py#L1072-L1077

And

https://github.com/SublimeLinter/SublimeLinter/blob/6d07b3653b14deadb824ecbb48c03ee9bacb3d1b/lint/util.py#L307-L312

We want explicit behavior.

kaste:
So basically we must find the linters where `temfile_suffix` is set and add `${file}` to cmd if it is '-', otherwise we must add `${temp_file}`

kaste:
We must first agree on the name here. But 'file_on_disk' is a good one.

Note, for gotype the cmd would be `gotype -e -t .` and then `is_file_only_linter = lambda: True`

braver:
You know, we could go around all contrib plugins and ask if they want to transfer. So far everyone is happy to have someone else maintain their thing, or it’s unmaintained and we can just move it. I don’t see why anyone would want a plugin that’s not part of the org, and it would allow us to refactor where we want.

kaste:
And, no, I don't *want* more linters in the org. I want more contributors.

braver:
I want more linters in the org and then also more contributors 😉 . The reality of all open source "communities" however, is that the number of actual contributors is two orders of magnitude smaller than the open source evangelists think.

braver:
Ha, SL2 *was* a monorepo :)

Issue #1263, opened 2018-03-22
${file} variables are only replaced once


FichteFoll:
If a command intends to use `${file}` multiple times (mypy), only the first occurance is replaced. The same applies to `${real_file}` and `${temp_file}`. For `${args}` this doesn't really make sense.

kaste:
This is a bit underdeveloped right now, bc we planned to put `cmd` into the settings as well.

PR #1292, opened 2018-04-04
Iterate around popen call

### LINKED ISSUE #1201: Reduce concurrency of the backend

### LINKED ISSUE #1288: Call `notify_failure()` when popen process crashes
After #1284 and obviously #1283

kaste:
- Enable fancy logging for tmpfile
- Inline tmpfile. Move from util to linter
- Do not use STDIN for `python -V` checks

Prepare #1263 , #1290 

Replaces #1250 
Fixes #1288 
Fixes #1201

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -37,6 +37,10 @@
     // the regions. E.g. try 'comment' to get a soft gray.
     "highlights.demote_scope": "",
 
+    // Send `terminate` signal to old lint processes, if their result would
+    // be thrown away. If false we fire-and-forget processes instead.
+    "kill_old_processes": false,
+
     // Lint Modes determine when the linter is run
     // background: asynchronously on every change
     // load_save: when a file is opened and every time it's saved
--- a/lint/base_linter/python_linter.py
+++ b/lint/base_linter/python_linter.py
@@ -59,7 +59,7 @@ class PythonLinter(linter.Linter):
                     return True, None
 
                 logger.info(
-                    "{}: Using {} for given python '{}'"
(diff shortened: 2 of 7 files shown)


kaste:
I rewrite #1250 on top of here. Stay tuned. 🎧

kaste:
☺️ 🍵

kaste:
Since all these changes around popen are rather risky I would like to see a pre-release. Maybe we still have some beta testers.

braver:
> Maybe we still have some beta testers

I doubt it. I think we've had more success by actively recruiting testers for specific PRs. But we can try.

So, I'd like to round up all open PR's so we can get all the fixes an safe improvements out there, before we make life more complicated with a pre-release.

PR #1293, opened 2018-04-04
Do not flash status bar for warnings

### LINKED ISSUE #1289: Do not flash status bar on `logger.warning`s
After #1283 we have a stable 'erred' notification, so we can remove the (hardly readable) status bar flashing.

kaste:
Fixes #1289


kaste:
stylelint is not in our org? They still mark crashes in the view via regions.

braver:
I think we may need to overtake stylelint, the maintainer is awol.

PR #1295, opened 2018-04-04
Iterate args DSL

### LINKED ISSUE #1286: Allow something like `--foo` in our args DSL
Generally, `defaults` has DSL abracadabra.

We support `--foo bar` via `"--foo:"` and `--foo=bar` via `--foo=`. Now, for the `=` case this is somewhat okay, but it would be nice if we could just specify `"--foo"` (without the `:`) in the first case.

kaste:
- Make \<joiner> optional
- Remove special '@' handling
- Use the value `True` as flag
- Simplify code by checking for falsy values first

Fixes #1286

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1,5 +1,4 @@
 from collections import namedtuple, OrderedDict, ChainMap, Mapping, Sequence
-from numbers import Number
 import threading
 
 from fnmatch import fnmatch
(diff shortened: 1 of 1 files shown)


braver:
Oh boy. I know nothing about this. Does this break plugins?

kaste:
Of course not. LOL

braver:
image

braver:
No, this is me reading the diff: 

confused math gif by cbc-downsized

kaste:
So I guess you kinda figured it out

kaste:
I'll merge this. This is a BREAKING CHANGE for the `True` case only

```
before:
	'--stdin': True  => '--stdin True'

after:
	'--stdin': True   => '--stdin'
	'--stdin': 'True' => '--stdin True'
```

Note that it was not possible before to just set the flag '--stdin' bc all falsy values meant skip this arg.

kaste:
Notable: The joiner is now optional.

```
before:
	'--stdin:': True
            ^ ugly
after:
	'--stdin': True
```

PR #1296, opened 2018-04-04
Remove unused `linter.build_options`

kaste:
The last user of this method was the annotations linter. (AFAIK)


braver:
There are just 2 contrib plugins that are dead an unmaintained (and apparently unused) already anyway. 🛥 it!

PR #1297, opened 2018-04-05
Sync goto command to actual regions

kaste:
Instead of jumping to the next/previous error as stated in `persist.errors` we jump to the next/previous highlighted region. This is so that after editing a view (and before the next lint result comes in) we still jump correctly. 

For this to work, we now draw invisible regions if the user set 'mark_style' to 'none'. (Before we dropped these errors.)


kaste:
Superseded by #1299

PR #1298, opened 2018-04-05
Sync tooltips to the actual highlighted regions

kaste:
Ensure the hover tooltips show the actual errors under the cursor/line even after edits.

This is on top of #1297 

This one must be performance tested with big files, lots of errors first. The technique is basically okay.


kaste:
Notable: This should enable proper tooltips over multiline errors.

PR #1299, opened 2018-04-06
Keep error position in sync while editing

kaste:
Yes we can.

### DIFF
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -3,6 +3,8 @@ import sublime_plugin
 
 from itertools import dropwhile, takewhile
 
+from .lint import persist
+
 
 """
 Implement typical Goto Next Previous Error Commands.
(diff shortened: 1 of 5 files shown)


kaste:
Oh, hi @braver 👋

braver:
Hi 👋 

Those 4 commits work really well as far as I can tell (at least with that fix 😉 ). Wanna merge?

kaste:
Hmmmm, that's a breakthrough PR no 🎉 no 🌶 ? Did you test this with lint_mode == save

braver:
I tested with a file only linter and compared with the current state of master, and I must say it was day and night. However, I don’t like working on that TypeScript project no matter how good the linting is and that kinda limits my enthusiasm for the entire thing. With Python everything was already pretty awesome.

So, sorry. I do understand breakthrough here, so here’s some celebration for you:

🎉👌🏻🕺🏼🎆

kaste:
So this PR is a bit tested and implements the feature. I'll merge without necessary refactoring. So they will be fixups and iterate-on PRs as soon as I get my head around how this now actually works.

kaste:
Thanks for the 🕺

PR #1300, opened 2018-04-09
Showcase: Flash jump position after goto command

kaste:
Flash error jumped to after invoking goto command. Flash scope is 'markup.bold', see in the code.

### DIFF
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -2,6 +2,9 @@ import sublime
 import sublime_plugin
 
 from itertools import dropwhile, takewhile
+import uuid
+
+from .lint import queue
 
 from .lint import persist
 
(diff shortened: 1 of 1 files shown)


braver:
They don't!

So, eh, we need to test the current state of master before releasing because we probably break a lot of people's computer if we don't right?

braver:
Ok, clear. I'll look for bugs too. 👍 

This particular PR is for me a bit on the fancy side. It also overlays an outline highlight on my squiggly underlines and that looks a bit wonky. It's not bad, perhaps I need to get used to it.

kaste:
We should manage somehow that something like this lands in master bc of the many PRs I never have the time to actually see it in action and improve it.

This PR here is merely an improvement for quiet views and otherwise invisible highlights. It also helps if you jump to an error currently not in the viewport. 

If it doesn't look good with squiggles I could hide them during flashing. But generally UI experiments are difficult if you cannot try them.

FichteFoll:
Also note that some people still have pre-releases enabled and could help with testing such an unstable or develop branch.

braver:
@kaste addon-i-fy this perhaps?

kaste:
yep

kaste:
This is now here https://github.com/kaste/SublimeLinter-addon-goto-flash

PR #1301, opened 2018-04-09
Implement `on_stderr` callback

### LINKED ISSUE #1156: New API to pass captured stdout and stderr separately to find_errors
Traditionally, these two streams aren't strictly related in a time-dependant manner. Linters may output lint results on stdout and general usage warnings or errors on stderr. Usually, `find_errors` would only want to operate on stdout while still capturing stderr to look for errors and display those to the user.

There is currently no way to get both of these streams seperately since you can only specify either or have them merged (by simple concatenation).

### LINKED ISSUE #1314: If we catch BrokenPipe log it
Currently, we eat these errors bc we expect them when we terminate old processes. But this can hide other errors as reported https://github.com/SublimeLinter/SublimeLinter/issues/347#issuecomment-380415897

So, we should log the error in a way it does not confuse our users.

kaste:
Fixes #1156 
Fixes #1314 

As proposed in https://github.com/SublimeLinter/SublimeLinter/issues/1156#issuecomment-380546660

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -246,10 +246,6 @@ def communicate(cmd, code=None, output_stream=STREAM_STDOUT, env=None, cwd=None,
             startupinfo=create_startupinfo(),
             creationflags=get_creationflags()
         )
-
-        with persist.active_procs_lock:
-            persist.active_procs[bid].append(proc)
-
     except Exception as err:
         try:
             augmented_env = dict(ChainMap(*env.maps[0:-1]))
@@ -269,7 +265,14 @@ def communicate(cmd, code=None, output_stream=STREAM_STDOUT, env=None, cwd=None,
             logger.info(make_nice_log_message(
                 'Running ...', cmd, uses_stdin, cwd, view, None))
 
-        out = proc.communicate(code)
+        with persist.active_procs_lock:
+            persist.active_procs[bid].append(proc)
+
+        try:
+            out = proc.communicate(code)
+        finally:
+            with persist.active_procs_lock:
+                persist.active_procs[bid].remove(proc)
 
         if output_stream == STREAM_STDOUT:
             return _post_process_fh(out[0])
@@ -284,10 +287,6 @@ def communicate(cmd, code=None, output_stream=STREAM_STDOUT, env=None, cwd=None,
                 return stdout
             else:
                 return ''.join(_post_process_fh(fh) for fh in out)
-
-    finally:
-        with persist.active_procs_lock:
-            persist.active_procs[bid].remove(proc)
 
 
 def _post_process_fh(fh):
--- a/lint/util.py
+++ b/lint/util.py
@@ -283,7 +283,19 @@ def communicate(cmd, code=None, output_stream=STREAM_STDOUT, env=None, cwd=None,
             'Running ...', cmd, uses_stdin, cwd, view, None))
 
     with store_proc_while_running(bid, proc):
-        out = proc.communicate(code)
+        try:
+            out = proc.communicate(code)
+        except BrokenPipeError as err:
+            friendly_terminated = getattr(proc, 'friendly_terminated', False)
+            # TODO: Consider `raise err from None` for friendly_terminated
+            # Bc this is a somewhat expected failure which should NOT count
+            # as 'no errors'. E.g. it should NOT empty all highlighted regions.
+            if friendly_terminated:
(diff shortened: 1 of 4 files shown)


lint/util.py
```diff
@@ -279,7 +279,14 @@ def communicate(cmd, code=None, output_stream=STREAM_STDOUT, env=None, cwd=None,
         if output_stream == STREAM_STDERR:
             return _post_process_fh(out[1])
         if output_stream == STREAM_BOTH:
-            return ''.join(_post_process_fh(fh) for fh in out)
+            if _linter and hasattr(_linter, 'on_stderr'):
+                stdout, stderr = [_post_process_fh(fh) for fh in out]
+                if stderr:
+                    _linter.on_stderr(stderr)
```
FichteFoll: Calling this from within `util.communicate` seems bad practice. I'd rather return a `stderr, stdout` tuple always (with either as `None` if not requested) and let some code inside `Linter` take care of handling the different streams, i.e. passing them around. 

Downside is that it changes the function signature and some plugins may use it. I know mypy does.

Ideally, we would add a second parameter to `find_errors` to hand the streams separately there, but I suppose that would be too breaking?

kaste: Calling this from `util` is not an issue. I already refactored it from both sides, so that I have this ugly `_linter` arg. I can move this to `Linter` flip the arguments and call `_linter` `self`. 

Now, initially you think you want full control and everything ready in `find_errors` or something similar. But then, should every plugin implement `find_errors` to get proper failure messaging? Probably not, so the question was if we are good to go with a simple 'callback' fn. 

We can make this opt-in via `hasattr` and if we get more knowledge about the different linters, we probably can make a trivial `on_stderr` the default.

The trivial impl is 

```
def on_stderr(self, output):
	logger.error(output)
	self.notify_failure()
```

FichteFoll: Ah well, considering that usually linter output happens on stdout and stderr is for usage instructions, I guess special-casing the stderr stream like this is okay.

You know more about how linter plugins interact with the core, so I'll retract my concerns. These are just the kind of things that cause me to think "maybe I structured this suboptimally", but I'm not keen on bothering everyone with that.

kaste:
This actually looks a lot nicer if we present it as error. But that may be to risky.

kaste:
Naw, the eslint config error will be demoted to a warning bc it is a default error. But if you setup a wrong executable or arg the error popup is clear and easy to read.

braver:
Oh, that's a good point 👍

kaste:
Let's see how this one goes.

`communicate` now generally returns a proc like object and passes it to a new `parse_output(proc, virtual_view)`. The proc has `.stdout//.stderr//.pid//.returncode` members. Also a `.combined_output` which is the string we used before. 

Standard error processing occurs automatic as before if linter sets STREAM_BOTH and has a callback `on_stderr` defined.

FichteFoll:
Looks pretty good. I like this a lot more than the earlier version. (Obviously I didn't review every detail.)

kaste:
Thanks, I did enough hacks here in the last two or whatever weeks.

braver:
Ok, we can perhaps merge this and do a pre-release, see if that catches some bugs if people are still on the pre-releases

alkuzad:
ok figured out that I need "2>&1" for groovyc linter, log message about this split looks kinda like deprecation error, not something that breaks plugins :D

kaste:
You don't need 2>&1. Open an issue and show your code. Pretty sure we can help you.

alkuzad:
@kaste yes, indeed it would be best to handle real errors too, like bad paths or arguments. I've dig into code and setting `on_stderr = None` is better option.

But groovyc is compiler so any syntax error it founds goes to stderr as runtime error. There is also linter called CodeNarc but it does not check syntax errors, only predefined rules.

SublimeLinter was designed for linters, like pyflake or pylint, which takes code and errors are only raised when they make analysis impossible (like invalid paths or any other RuntimeError).

I'm pretty sure there is at least few methods to get code from on_stderr back to right track (or match code normally and use notify_failure()) but right now I'm not willing to dig into it right now. Groovyc is simple input/output program, with limited set of options and I have provided interfaces for most common ones. If there is fail, 99% of it is just code :)

kaste:
Pretty sure I would have recommended setting on_stderr to None as well. It's an easy start. :+1:

PR #1306, opened 2018-04-10
Fix for ghost regions after undo

kaste:
Generally, Sublime keeps ghost regions of length 0 if you delete text
that has a region attached. Since we invalidate highlights under the
cursor when editing, we never see them, until the user hits 'undo'
then they re-appear (out of Sublime's internal cache).

This problem was hidden by the fact that we used generalized region keys
and so a following update always fixed the ghosts. But since we now use
uuid region keys, these ghost become permanent when SL forgets about
these keys and never redraws them.

Ideally we could redraw everything all the time, but here we store the
invalidated keys in memory and restore them on 'undo'.

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -224,7 +224,6 @@ class IdleViewController(sublime_plugin.EventListener):
     def on_modified_async(self, view):
         active_view = State['active_view']
         if view.buffer_id() == active_view.buffer_id():
-            invalidate_regions_under_cursor(active_view)
             set_idle(active_view, False)
 
     def on_post_save_async(self, view):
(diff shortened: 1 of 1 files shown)


braver:
Ah, I noticed a dangling region but haven't found time to report it yet. So this would fix that? 🎉

kaste:
🤞 or 🖕

braver:
Yes, it's a shitty editor. It's also the best editor. Which probably says a lot about all the other editors.

braver:
screen shot 2018-04-11 at 15 55 05

Ok, I can't seem to reproduce dangling inline highlights so easily anymore, but I managed to produce a dangling gutter mark.

kaste:
Icon dangles should at least go away when you produce a new error with the same icon. The highlights here were complete, unreachable zombies.

PR #1307, opened 2018-04-10
Fix: Reset (failed) linters before each run

kaste:


braver:
I'm not sure how to test it, but it does seem to work ok. Merge at will IMO.

PR #1309, opened 2018-04-10
Fix: Source gutter icons from all linters of current file

### LINKED ISSUE #1302: gutter icons "protected regions" don't always work
Perhaps since recent work on master. 

screen shot 2018-04-09 at 20 55 20

They do sometimes show, and it feels like a race condition somewhere.

Also, sometimes I can get the line report tooltip even if the icon is hidden by GitGutter, and sometimes I can't get the line report tooltip no matter what. The inline tooltip for regions is more reliable however.

tooltips

kaste:
Fixes #1302

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -81,6 +81,9 @@ def on_lint_result(buffer_id, linter_name, **kwargs):
     errors = persist.errors[buffer_id]
     errors_for_the_highlights, errors_for_the_gutter = prepare_data(errors)
 
+    view = views[0]  # to calculate regions we can take any of the views
+    protected_regions = prepare_protected_regions(view, errors_for_the_gutter)
+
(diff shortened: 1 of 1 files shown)


braver:
That doesn't fix the problem for me, I can still reproduce the gif from #1302

kaste:
This one solves the second one. The animated.

kaste:
Before 



After

braver:
Ok that fixes at least one problem then, I got to find out what my problem is then. It’s nothing to do with gitgutter, but the data internally is incorrect.

kaste:
The bug here was easy to reproduce and easy to see in the code. It was clear that the var `errors_for_the_gutter` first holds *all* gutter icons and a few lines later holds only the gutter icons from the latest arriving linter. And I took the wrong one.

braver:
Excellent, works now!

⛵️

PR #1317, opened 2018-04-11
Fix: Make linters throw away objects to keep the view up-to-date

### LINKED ISSUE #1308: If you close the 'initial' view into a buffer, `linter.view` becomes invalid state
Super 🐛 

Repro

- Open first view into file
- Lint
- Open second view into file
- Lint
- Close *first* view
- Switch to second view
- Lint

In the debug you should see something like this 

image

T.i. `view.file_name() == None; view.buffer_id() == 0`

kaste:
Fixes #1308


braver:
LGTM.

What do you say? Merge this and #1301 and then ship a release?

PR #1319, opened 2018-04-16
Assign 'main' linters even to empty views

### LINKED ISSUE #1313: Empty files should show active/assigned linters as well
- Open new file
- Change syntax to Python
- Probably user expects 'flake8(ok)' in the status bar

(Technically, we don't assign linters to empty views at all. So this is not a bug of the status bar indicator.)

kaste:
Fixes #1313

### DIFF
--- a/active_linters_view.py
+++ b/active_linters_view.py
@@ -39,13 +39,17 @@ def redraw_bid(buffer_id, linter_name, errors, **kwargs):
     else:
         problems.pop(linter_name, None)
 
-    for view in views_by_buffer(buffer_id):
-        if view.buffer_id() == buffer_id:
-            draw(view, problems)
+    for view in views_into_buffer(buffer_id):
+        draw(view, problems)
 
 
-def views_by_buffer(bid):
-    return (view for window in sublime.windows() for view in window.views())
+def views_into_buffer(bid):
+    return (
+        view
+        for window in sublime.windows()
+        for view in window.views()
+        if view.buffer_id() == bid
+    )
 
 
 def count_problems(errors):
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -58,9 +58,6 @@ def get_lint_tasks(linters, view, view_has_changed):
         tasks = []
         for region in regions:
             code = view.substr(region)
-            if not code:
-                continue
-
             offset = view.rowcol(region.begin())
 
             # Due to a limitation in python 3.3, we cannot 'name' a thread when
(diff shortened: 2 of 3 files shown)


kaste:
Interesting detail. It initially shows `annotation(ok)` but not `flake8(ok)`. So it looks like we get an initial activated call, but the syntax is not ready.

Ok. What I don't like here is that it states e.g. `eslint(ok)` but it actually doesn't run `eslint`, so the 'ok' bit can be misleading.

braver:
> so the 'ok' bit can be misleading.

Good point.

This is a very transient situation though, perhaps we're overdoing it by spending effort here. What use is linting an "empty string", what expectation can a user have here? As soon as you start adding something to the file, the linters jump into action. That's good feedback too.

kaste:
The usage here is only to test if the linter is working. The easiest way to know for us is to just run it.

My usage was: setup new project, create new file. eslint says ok. You start typing ... and boom config error. So you go back to config mode and loose the focus.

braver:
What's the status here? It looks like it's a) incomplete (ie. my screenshots), and b) the linter doesn't actually run so we can't give good feedback about its status. 
Do we need a different status, (e.g. `flake8(...)`)? Do you need more info to move forward, or do you even want to still do this?

kaste:
I think the wanted UX is that the linters give early feedback via '(ok) vs (erred)' if they are correctly configured and coding can actually start. The overhead of running on empty view is a wrong optimization here. In fact we did run on empty views not long ago and just changed this behavior for eslint. 

So, the clean way here is to fix the bug for eslint in eslint. (The bug was: eslint does not accept '' the empty string via stdin. This bug https://github.com/eslint/eslint/issues/9515 has been fixed of course in eslint upstream (4.9 has bug, 4.12 runs fine), but we want to support older versions of eslint, so we need to fix it over there.)

kaste:
ADD: Fixed in eslint 4.10.0

kaste:
(The placeholder buffer bug is an edge-case, so we tackle or don't tackle it with a new issue. Yet another Sublime bug probably.)

kaste:
Related https://github.com/SublimeLinter/SublimeLinter-eslint/pull/242

kaste:
The placeholder buffer root cause is that these placeholder views are not returned in `window.views()`. So, the linter runs and reports correctly, but at the end we fire LINT_RESULT just with a buffer_id, then we search for views with that bid and there are none in `window.views()`.

braver:
The placeholder view quirk is totally acceptable for me. Let's ⛵️ this.

Issue #1315, opened 2018-04-11
Prevent dangles for highlights and icons


kaste:
As seen https://github.com/SublimeLinter/SublimeLinter/pull/1306#issuecomment-380461021

highlight dangles can still be produced if regions magically appear on 'redo'.

Also: probably not all ways to modify the buffer are handled by invalidate errors *under cursor*.

kaste:
Okay, the 'revert' command can do this.

kaste:
This is non-trivial. Interestingly, SL3 had no code for that at all. But maybe that just means a practical solution is good enough. Via #1320 we first try a practical solution for highlight zombies. 

The gutter 'dangles' aren't zombies, so this would be rather an enhancement.

PR #1320, opened 2018-04-16
Stable instead of random uid per error

kaste:
Mitigates #1315 

We can always prevent zombie dangles if we re-use a region key. So if we switch from a random `uuid4()` to a stable *id* we can mitigate #1315. But, we must be sure to **not** use the same id more than once *at the same time*. Here we just hope that an error (as dict) as reported by a linter is unique for this current run.

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -3,12 +3,13 @@ import sublime
 from concurrent.futures import ThreadPoolExecutor, as_completed, wait
 from itertools import chain, count
 from functools import partial
+import hashlib
+import json
 import logging
 import multiprocessing
 import os
 import time
 import threading
-import uuid
 
 from . import util, linter as linter_module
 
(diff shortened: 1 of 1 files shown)


braver:
Ok, so you hash all aspects of the error together to create the id? That sounds like a good idea. 

I think the way SL3 dealt with it was to remove all regions on save, highlights and gutter icons, and start over from scratch. Of course none of that was async or performant, or a nice way to do it. 

This LGTM 👍

kaste:
I think the way SL3 dealt with it was not dealing with it. Bc you had no 'styles' object it was just unlikely to get zombies, yu just had to 'make' a new lint error and the zombies would went away. Now, I have different icons for different linters, and when I tackled the last TODO annotation, then undo/redo edit, I can end up with an annotation zombie as long as I don't add a new TODO comment (which is likely).

kaste:
Ah, for the id it would of course be nice if we could take the 'message' out of it. T.i. if we could say that the pair (linter_name, line, col, code) would be sufficient. Bc then we would get more same ids. But I don't know if a linter *could* report the same 'code' on the same 'position'. E.g. flake8 reports

```
939:20   error         flake8: F821         undefined name 'settingss'
``` 

Then you edit, and it reports

```
939:20   error         flake8: F821         undefined name 'settingsss'
```

As you can see, only the message changes.

FichteFoll:
Was running into lots of zombie highlights yesterday. I hope this pr will improve things and report  back, though I probably won't be coding much in the following days.

PR #1324, opened 2018-04-18
remove the backup of SL3 settings

braver:
It's not the most robust thing in the world and by now we won't have many upgraders left I think.


kaste:
LGTM. Away with it.

PR #1326, opened 2018-04-18
Zombies

### LINKED ISSUE #1315: Prevent dangles for highlights and icons

kaste:
The Real Thing™️  🖕 🤞 🖐  

Fixes #1315 

🤹‍♂️

### DIFF
--- a/zombie_protection.py
+++ b/zombie_protection.py
@@ -119,6 +119,8 @@ def on_lint_result(buffer_id, linter_name, **kwargs):
             # this 'micro'-task. Essentially, we want to run after the
             # main side-effects of the event (here probably: drawing)
             # are done.
+            # TODO: highlights_view should fire an event 'UPDATED_REGIONS'
+            # we can subscribe to instead.
             lambda: replace_top_undo_state(view), 1)
 
 
--- a/zombie_protection.py
+++ b/zombie_protection.py
@@ -1,0 +1,171 @@
+from collections import defaultdict
+
+import sublime_plugin
+
+
+# Sublime internally stores all highlighted regions and gutter icons
+# within its 'undo' machine. These are part of the buffer state.
+# The problem herein is that even if we deleted some highlights long ago
+# Sublime will resurrect them on undo/redo. These resurrected regions
+# are in fact zombies bc otherwise SL has forgotten their keys, and thus
+# can not access them anymore. E.g. we cannot delete them again.
+# In SL we only transition from the *current* set of region keys to the
+# *next*.
+# The solution is to model a separate, own undo machine which essentially
+# stores all region keys ever drawn.
+
+# TODO Move to util and use it throughout SL
+def distinct_until_buffer_changed(method):
+    last_call = None
+
+    def wrapper(self, view):
+        nonlocal last_call
+
+        bid = view.buffer_id()
+        change_count = view.change_count()
+        this_call = (bid, change_count)
+        if this_call == last_call:
+            return
+
+        last_call = this_call
+        method(self, view)
+
+    return wrapper
+
+
+IGNORE_NEXT_MODIFIED_EVENT = set()
+
+
+class UndoManager(sublime_plugin.EventListener):
+    @distinct_until_buffer_changed
+    def on_modified_async(self, view):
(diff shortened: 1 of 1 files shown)


braver:
I've got two notes for this one:

- of course I cannot reproduce the original issue right now 🙄 

- looking at the size and seeming complexity of the solution here, my gut feeling says there should be a more simple solution. 

So, as I understand it, Sublime Text will recreate regions on undo. When it does so, the regions may or may not match regions SublimeLinter would create in the state after the undo. If they don't match for whatever reason, they'll be zombies we cannot clean up because they don't exist in our own bookkeeping. 

So, that sounds like there are 2 solutions. The first, and the goal of this PR as I understand it, is to better understand ST and follow its behaviour to perfect our bookkeeping. To me this sounds like a complicated and fragile solution. Can't we simply not trust our bookkeeping, e.g. nuke all our regions and recreate them on save or something?

It could very well be that I don't understand something here, and this is the correct way to move forward, so if you want to do it be my guest.

kaste:
Repro video master

oncvjhgeph

(But I'm typing faster than the video recording here. Essentially you hit undo, redo, edit, which happens in real life. Say you edit, buffer states a - b - c - d. Now you realize, that this is going completely wrong. You know you need to go back (undo). Usually you hit undo 3, 4, 6 times quickly and maybe adjust the correct point to jump to with redo. If your focused you begin re-editing immediately. So you go from 'd': c - b - a - b - e.)

kaste:
Cleanup PR to only contain the zombie feat

> So, that sounds like there are 2 solutions. The first, and the goal of this PR as I understand it, is to better understand ST and follow its behaviour to perfect our bookkeeping. To me this sounds like a complicated and fragile solution. Can't we simply not trust our bookkeeping, e.g. nuke all our regions and recreate them on save or something?

If the second solution is to just "trust our bookkeeping" then well we can't trust our bookkeeping bc it is not good enough as long as Sublime will resurrect regions via undo/redo at will. Our current bookkeeping only stores the *current* state as drawn by highlights_view. (And honestly all Sublime plugins I looked at do this, so this PR is probably also a genuine, novelty solution.)

Now, of course, this is an API bug by Sublime. Sublime should provide an API like `view.get_region_keys()`. Then nobody would need any bookmarking at all. We would just compute what we want to draw. Ask Sublime about the current state, and then compute a diff.

The only way afaik to erase all regions and get a clean state is to issue 'File: Revert'. But you can only call this iff the view is clean (saved) and without any linter errors. 'Revert' could also have other side-effects we don't know about. 

The pro side of this PR is that it actually doesn't add complexity bc it is self-contained in one file. If we delete that one file, SL will work like our current master. Also pro: It actually tries to solve the problem at the root, in that sense it tries to be 'The Real Thing:tm:' instead of a workaround. 

On the con: I don't know if there other commands to undo/redo. 'Redo' already has the rather strange name 'redo_or_repeat'.

braver:
Ok, another idea. How about we remember all regions we ever had in a file, ie. we store them in a “to be deleted” list and we don’t remove them until the view closes. If SL ever recreates a region, we can use our list of all regions to wipe them all out. Perhaps memory leak-ish, but probably insignificant (compared to ST’s unlimited undo history).

kaste:
What's your main concern here. Bc with this PR we already store all region keys ever used. This iteration stores in memory. File storage is usually for some sort of persistence, long consistency. We could use the views settings for that.

braver:
Ok, that helps me read the code here. I guess I’m concerned that we might miss a command, and in general I don’t like it when something that should be simple becomes complicated. It kinda feels like we’re reverse engineering ST’s behavior and compensating for it. I’m never super happy about stuff like that,  but we gotta do what we gotta do. It’s not terrible, I think the PR does a good job and it’s actually less complicated than it looked. I’m just trying to help with some ideas. I’m super far out of my depth here, but I really want to try to help you think about the problem. 

So, as, with some roundabout comments, we seem to have very similar ideas about how this should be done (mine just being awkwardly voiced, and me not able to read your code well enough). Let’s merge this then.

kaste:
Do not delete the branch yet, I need a tiny bit here.

PR #1328, opened 2018-04-19
Begin writing 4.3 changelog

kaste:
That's basically just the compilation of things. @braver can you edit this?

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -3,5 +3,6 @@
     "4.0.0": "messages/4.0.0.txt",
     "4.0.1": "messages/4.0.1.txt",
     "4.0.3": "messages/4.0.3.txt",
-    "4.1.0": "messages/4.1.0.txt"
+    "4.1.0": "messages/4.1.0.txt",
+    "4.3.0": "messages/4.3.0.txt"
 }
--- a/messages/4.3.0.txt
+++ b/messages/4.3.0.txt
@@ -1,0 +1,29 @@
+SublimeLinter 4.3
+=================
+
+
+For users
+---------
+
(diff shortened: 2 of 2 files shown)


kaste:
I like how this changelog works. https://github.com/sindresorhus/refined-github/issues/1137
WDYT?

braver:
The messages are the only way to push something out to Sublime Text users. I’d like a nicer changelog, but I don’t want to duplicate anything.

kaste:
@braver You should release `4.3.0-pre.1` as `4.3.0`. 👌

kaste:
> The messages are the only way to push something out to Sublime Text users.

Ah well, no, we can use GitHub to send changelogs or interestings news to interested people like shown in the issue above. It's like a super-transparent newsletter and I really enjoy how they do it. You can easily imagine a dev-newsletter where we post 'Attention: we chang stderr handling in the next version. Better check if you're still working.'

Otherwise the messages are not that accessible. They popup at their will, and if you close'em they're basically gone for most users. Usually you would want to read them at your will. Maybe on mobile, in free time. Or just later. If they're in your inbox you just handle them the way you want, and can refer to them later again.

braver:
Ok, I don't really have the mind-space left over for this. I just want to do as much as I can in as little time as possible. The vibe I get from our user base is that they don't mind as long as we don't break too much, and those that do know their way around Sublime packages and where to find the messages. I'm not stopping anyone from doing anything, so if you want to communicate more or better do your thing. Me, I'm just going to merge the messages, edit them, make a release and get on with my payed job or sit in the sun with a book.

PR #1342, opened 2018-04-24
Prevent icon dangles

### LINKED ISSUE #1315: Prevent dangles for highlights and icons

kaste:
Extracted from #1326 bc these two commits are easier to decide.

- First, we remove some not working code.
- Second, we actively remove icon dangles.

'Dangles' are not Zombies. They are empty regions that hang around when the user deletes code that was highlighted before. In that case Sublime will not also erase the highlight regions but keep them *visible* although the highlighted text has been gone. (Sorry, but that just is _another_ Sublime core bug.) 

For underlined errors, we erase the regions as soon as the user edits on it. For highlighted lines (gutter icon), we erase the icon as soon as the line has been emptied or removed.

Fixes #1315

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -321,15 +321,23 @@ def invalidate_regions_under_cursor(view):
     selections = view.sel()
     region_keys = get_regions_keys(view)
     for key in region_keys:
-        if '.Highlights.' not in key:
-            continue
+        if '.Highlights.' in key:
+            if any(
+                region.contains(selection)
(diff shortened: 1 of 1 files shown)


braver:
I'm having a real hard time reproducing the dangling icons. The changes here look good though, so I'm 👍

braver:
At half time I'll release the 4.3.1

kaste:
LOL

PR #1343, opened 2018-04-24
Add `linter.finalize_cmd` which expands variables found in `cmd`

### LINKED ISSUE #1263: ${file} variables are only replaced once

### LINKED ISSUE #1290: Deprecate and remove magic appending the filename or tempfile name to `cmd`

kaste:
We use our normal `substitute_variables` fn to finalize the cmd. Also we factor this feat into its own method `finalize_cmd`. 

As long as auto_appending is only deprecated but not removed users can override this method to actively opt-out of auto_appending (go linters could do this). Otherwise, this method could be used to normalize paths in the future e.g. to support `wsl` environments for Windows.

DANGERZONE: `substitute_variables` would probably fail if there are any `$` in cmd. Bc these must be afair escaped via `\$`. 🦆 🤷‍♀️ 

Fixes #1263
Fixes #1290

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1014,13 +1014,11 @@ class Linter(metaclass=LinterMeta):
 
     def communicate(self, cmd, code=None):
         """Run an external executable using stdin to pass code and return its output."""
-        if '${file}' in cmd:
-            cmd[cmd.index('${file}')] = self.filename
-        elif '@' in cmd:  # legacy SL3 crypto-identifier
-            cmd[cmd.index('@')] = self.filename
-        elif code is None:
-            cmd.append(self.filename)
+        ctx = get_view_context(self.view)
+        ctx['file_on_disk'] = self.filename
 
+        cmd = self.finalize_cmd(
+            cmd, ctx, at_value=self.filename, auto_append=code is None)
         return self._communicate(cmd, code)
 
     def tmpfile(self, cmd, code, suffix=None):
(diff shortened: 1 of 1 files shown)


braver:
Looks good I guess. 

When I've got some time later, e.g. a rainy summer Sunday afternoon, I'd like to have a list of deprecations so I can check the linters in the org, and this time around also the contrib linters. We might be able to rescue a few more and then drop the old functionality.

kaste:
👍

This PR here tries to help some go linters bc they still don't work in SL4 land. Otherwise I think I should make a bigger PR with all the internal Linter API stuff so we can get a broader picture. I will probably make an issue first that combines there open issues, so we actually see what are user issues and what are pure refactoring goals. Bc pure refactorings can be bikeshedding. We don't need to look good, we want all use cases to be handled.

PR #1346, opened 2018-04-24
Alt Zombies

### LINKED ISSUE #1315: Prevent dangles for highlights and icons

kaste:
The lean thing:tm:

@braver's idea but without the file 🤣 

We use an everstore to store all region keys ever drawn by SL
per bid. On undo/redo we just restore the current state from
this everstore.

Fixes #1315

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -303,25 +303,13 @@ def toggle_all_regions(view, show):
         view.add_regions(key, regions, scope=scope, flags=flags)
 
 
-InvalidatedRegions = defaultdict(set)
-
-
 class InvalidateEditedErrorController(sublime_plugin.EventListener):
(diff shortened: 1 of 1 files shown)


braver:
I like this!

For the record, I never wanted to propose storing stuff elsewhere than in memory. If I did I blame auto correct and my iPad’s stupid keyboard. 😉😁

Do you have a preference between this and the Original™️ Zombie Protection?

kaste:
I prefer this one of course bc I don't need to read 60 lines of comments to understand it.

braver:
Want me to make some ascii art? Will make me feel like I’m really contributing too 👨🏼‍🎨

kaste:
A Zombie in ~6 lines maybe here:

image

PR #1350, opened 2018-04-26
Hotfix: Treat user args beginning with '@' as values without keys

### LINKED ISSUE #1349: Regression for user args starting with `@`
`phpmd` defines 

```
	'@rulesets:,': 'cleancode,codesize,controversial,design,naming,unusedcode'
```

Don't know if this actually worked before, but sure, I made it non-working in SL 4.3. (As always: Remove cruft you don't understand. LOL)

This is btw the same as setting

```
	'args': 'cleancode,codesize,controversial,design,naming,unusedcode'
```

kaste:
E.g.
  `'-ruleset': 'cleancode'` becomes `-ruleset cleancode` on the cli.
  `'@ruleset': 'cleancode'` becomes `cleancode`

Fixes #1349

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -17,7 +17,7 @@ from .const import WARNING, ERROR
 logger = logging.getLogger(__name__)
 
 
-ARG_RE = re.compile(r'(?P<prefix>--?)?(?P<name>[@\w][\w\-]*)(?:(?P<joiner>[=:])?(?:(?P<sep>.)(?P<multiple>\+)?)?)?')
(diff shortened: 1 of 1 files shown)


kaste:
(That doesn't mean 'that-I-like-it' 🎶)

kaste:
So for the docs (LOL?) this would mean the `@` prefix means this is an unnamed, 'positional' arg on the cli. The other prefixes are like keyword args.

braver:
Can’t we drop the feature and let the args cover it?

kaste:
Hm, but we have positional and named args on the command line, so that makes it a complete solution. I just didn't know the feat and it wasn't documented. 

Dropping a feat should be decision not a mistake.

braver:
Sure, just wanted to entertain the thought for a while. I like dropping features too much 😉

LGTM then

PR #1354, opened 2018-04-27
Append errors if message panel is open

### LINKED ISSUE #1339: Only the last thrown error is visible in our error panel
If the message_view (error panel) is open, each `logger.error` call overwrites the previous message. It would be probably better to append here if the panel is open. (Esp. since errors are only displayed once per session, quickly overwritten errors become otherwise inaccessible)

kaste:
Fixes #1339

### DIFF
--- a/message_view.py
+++ b/message_view.py
@@ -12,14 +12,27 @@ def plugin_unloaded():
 
 class SublimeLinterDisplayPanelCommand(sublime_plugin.WindowCommand):
     def run(self, msg=""):
-        panel_view = self.window.create_output_panel(PANEL_NAME)
+        window = self.window
+
+        if is_panel_active(window):
+            panel_view = window.find_output_panel(PANEL_NAME)
+        else:
+            panel_view = window.create_output_panel(PANEL_NAME)
+
+        scroll_to = panel_view.size()
+        msg = msg.rstrip() + '\n\n\n'
+
         panel_view.set_read_only(False)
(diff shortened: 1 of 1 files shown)


message_view.py
```diff
@@ -11,14 +12,27 @@ def plugin_unloaded():
...
+
+        scroll_to = panel_view.size()
+        msg = msg.rstrip() + '\n\n\n'
+
         panel_view.set_read_only(False)
         panel_view.run_command('append', {'characters': msg})
         panel_view.set_read_only(True)
-        panel_view.show(0)
-        self.window.run_command("show_panel", {"panel": "output.{}".format(PANEL_NAME)})
+        panel_view.show(scroll_to)
```
braver: Would that make it scroll as new messages come in? I'm not sure I'd want that.

kaste: This looks and feels pretty good. I added this while testing the feature.

Without it you hardly notice the 'append'. And if you really have the error panel open, you kinda wait for a reaction.

Say you try the --foo arg on flake8. Then the panel opens and says nope, there's no foo allowed. You edit to fooz, then save, then the error panel will append the new message and show it.

If you closed before the error will be replaced, and thus will also show up directly, without scrolling.

braver: Ok, not a big deal anyway.

kaste: The pragmatic braver™️

PR #1355, opened 2018-04-27
Assign linter name to each error in backend

### LINKED ISSUE #1332: Assign linter_name in backend.finalize_errors
A user should not fill in the boilerplate linter_name when returning errors. This becomes useful if the user implements `parse_output`.

kaste:
Fixes #1332

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -85,7 +85,7 @@ def execute_lint_task(linter, code, offset, view_has_changed, settings, task_nam
     with reduced_concurrency():
         try:
             errors = linter.lint(code, view_has_changed, settings) or []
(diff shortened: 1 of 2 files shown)


lint/backend.py
```diff
@@ -109,7 +109,9 @@ def reduced_concurrency():
         yield
 
 
-def finalize_errors(view, errors, offset):
+def finalize_errors(linter, errors, offset):
+    linter_name = linter.name
```
braver: Any particular reason for the intermediate var?

kaste: That's just automatic writing style bc after that is a for loop. And you just put everything that cannot change out of it.

It's not for performance, but to keep the changing parts from the constants.

PR #1357, opened 2018-04-27
Fix brittle automatic panel

kaste:
Decide if we toggle the panel after save on 'LINT_RESULT'. Before
we used 'LINT_END' which is super brittle bc it fires from an
autonomous thread while 'LINT_RESULT's fly in on the shared Sublime
worker thread.

The panel also opens as soon as the first linter reports errors,
instead of waiting for all results. Still 🤞


braver:
LGTM

PR #1363, opened 2018-04-28
[WIP] Which executable?

### LINKED ISSUE #1335: Factor out method which returns the final executable
Currently build_cmd and in turn get_cmd returns the complete computed `cmd`. It is useful though to just get the currently configured executable of a linter (type: str or list) so that a user can use our config system to query e.g. the version of such an executable. 

This method could be named `which_executable(cmd) -> str|list`.

### LINKED ISSUE #1362: Intent to change signature of `context_sensitive_executable_path`
`context_sensitive_executable_path` had a polymorphic signature in that
it also allowed the full `cmd` to be passed in. However all
implementations then only extracted `cmd[0]` and acted upon that.

Esp. for the new `which_executable` we want a simpler signature, so that
the user can just call e.g. `self.which_executable('mypy')`. We, of
course, could transform this internally to `['mypy']`, pass it down,
and transform it back to `'mypy'` in every `csep` implementation but
that seems unnecessary. 

So the question is, is it sufficient to compute the executable from one
`str` (like all our implementations), or does the executable depend on the 
full `cmd`.

kaste:
Expose `get_executable() -> Optional[List[str]]`
Factor `linter.which_executable(str) -> Optional[List[str]]`
Change signature of `context_sensitive_executable_path` as proposed in #1362 

Fixes #1335 
Fixes #1362

### DIFF
--- a/lint/base_linter/composer_linter.py
+++ b/lint/base_linter/composer_linter.py
@@ -31,6 +31,7 @@ class ComposerLinter(linter.Linter):
         if self.manifest_path:
             self.read_manifest(os.path.getmtime(self.manifest_path))
 
+    @util.ensure_cmd_is_str
     def context_sensitive_executable_path(self, cmd):
         """
         Attempt to locate the composer package specified in cmd.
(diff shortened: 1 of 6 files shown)


braver:
Works for me. Anything in particular I should test?

braver:
@kaste Wanna merge this?

kaste:
Nope or not sure. Don't think it adds enough value right now. I have added several other commits on top of it on my computer. I want to get a clearer picture of the direction the API should go.

braver:
👍

kaste:
Bye 👋

PR #1368, opened 2018-05-01
consider the shell_environment setting

braver:
Via this comment: https://github.com/SublimeLinter/SublimeLinter/pull/1266#issuecomment-385582192

If `shell_environment` is set to false, we only set the env to "ready" after the timeout runs out because ST will never update then env.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -382,6 +382,13 @@ def environment_is_ready():
     if ready:
         return True
 
+    s = sublime.load_settings("Preferences.sublime-settings")
+    if not s.get('shell_environment', False):
+        # If false, ST will not read the env from the shell,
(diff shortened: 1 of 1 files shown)


kaste:
Although the patch looks reasonable you didn't confirm the bug in the first place. Your own comment here suggest that we would generally wait here which is not true. Our naive heuristic is that a path with /local/ in it is always good.

braver:
As I understand it, we wait until our heuristic says the env is good. If shell_environment is false, we can skip the entire thing, the env will not be updated. That’s what this PR is about.

This has been solved in ST so I’m also ok with just not bothering with it at all 🤷🏼‍♂️

braver:
3.1 is already out there, so let's drop the entire thing.

PR #1372, opened 2018-05-02
Enable reloader after install/upgrade

kaste:
I did test this with a local repository; FichteFoll explained this in the kill_switch thread.

### DIFF
--- a/log_handler.py
+++ b/log_handler.py
@@ -15,6 +15,10 @@ handler = None
 error_panel_handler = None
 
 
+def plugin_unloaded():
+    uninstall()
+
+
 def install():
     install_std_handler()
     install_error_panel_handler()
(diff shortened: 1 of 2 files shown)


kaste:
BTW, @FichteFoll `file://...` did not work. I have to serve the files via localhost.

braver:
Well, it’s worth a try at least 👍🏻

kaste:
The actual zip packages must be served via localhost. The repository hack worked with the local file.

kaste:
About the noise: My first issue here was in the range #6xx. This PR has the number #1372!

kaste:
FWIW I made some notes https://github.com/SublimeLinter/SublimeLinter/wiki/Test-upgrade-locally

PR #1382, opened 2018-05-07
Sublime Text 3.1

braver:
3.1 is out, which helps us wrt. colors and loading the env. We should probably give it a week to trickle out to users though.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -143,7 +143,7 @@
         {
             "mark_style": "outline",
             "priority": 1,
-            "scope": "region.yellowish markup.changed.sublime_linter markup.warning.sublime_linter",
+            "scope": "region.yellowish markup.warning.sublime_linter",
             "icon": "dot",
             "types": [
                 "warning"
@@ -152,7 +152,7 @@
         {
             "mark_style": "outline",
             "priority": 1,
-            "scope": "region.redish markup.deleted.sublime_linter markup.error.sublime_linter",
+            "scope": "region.redish markup.error.sublime_linter",
             "icon": "dot",
             "types": [
                 "error"
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -32,55 +32,12 @@ These scopes then receive color from the color scheme, which is kinda like a CSS
 
 SublimeLinter expects the scopes ``markup.warning`` and ``markup.error`` to get
 correct colors from most color schemes.
-We have some fallbacks, and SublimeText > 3149 helps us with scopes like ``region.redish``.
+We use scopes like ``region.redish`` for color schemes that don't provide colors for these scopes.
 
-However, sometimes that doesn't work and your icon is white.
-In this case you need to use a scope that *does* get the color you want.
+To change the colors, you can use region.colorish scopes:
+redish, orangish, yellowish, greenish, bluish, purplish, pinkish
 
-Take this code snippet as a starting point to exact colors from:
-
-.. image:: colors/1.png
-
-If I put the cursor just before the ``=`` I can then use the
-"Tools > Developer > Show Scope Name" menu item to find out what scope that is:
-
-.. image:: colors/2.png
-
-.. image:: colors/3.png
-
-So, let's hit "copy" here and put that into the "scope" for errors.
-Do the same thing for the yellowish color in ``getElementById``.
-This gets me the following styles:
-
-.. code-block:: json
-
-    {
-        "styles": [
-            {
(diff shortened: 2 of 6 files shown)


kaste:
LGTM except for the messages.txt

Either it just works for most users or we should tell em they should upgrade to 3.1

kaste:
We had reports imm after they shipped the new version, that it doesn't work. And we have reports from yesterday that SL worked with 3.0 but not with 3.1

Now, I can only guess, but since they wanted not a separate callback, but a guarantee that the env is ready before plugin loaded; you can already guess that they have short or shorter timeouts. For example, if you implement something like this, you pretty sure don't want to block `plugin_loaded` for 1 minute.

braver:
I must not be watching some of our plugin repos then.

> you can already guess that they have short or shorter timeouts

Sure, but if you need to wait a minute for your env to load, perhaps you should just configure it manually in ST. Is that really a valid use case *we* need to solve? 

Also, if it stopped working now, apparently our existing solution doesn't even work either.

kaste:
Our current solution depends on a mature implementation of that feat in ST. If they regress here, we regress as well. (So, we had to iterate on our solution.)

Now, the bug goes up to `nvm` and then to `node`. For affected users the reported **avg** wait time is 2-3 secs, but there are valid reports for wait times up to 1 minute. 

Waiting 1 minute is still a valid use case for an app. Before I started here in SL land, I (re-) started ST maybe once a month.

braver:
If you don't restart Sublime Text, there is no problem for us to solve. Remember that the problem already, without our intervention, goes away by itself. It's only immediately after starting Sublime Text that there may be a problem for some users on macOS. Right now, that number of users is reduced to a handful with environments that take very long to load. 

Either we go out of our way to solve that, or we don't and they wait (they'll be waiting anyway) or they tell ST or SL what their path is manually.

braver:
Also, do you have some reports you can linkt to? I feel like I'm guessing here. I tend to be very up to date about what happens around ST dev builds and I've seen zero reports of the env loading not working as advertised.

kaste:
> If you don't restart Sublime Text, there is no problem for us to solve. Remember that the problem already, without our intervention, goes away by itself.

That is not correct. We wait for the `env is loaded` coming from ST. If they fail more often bc of timeouts or whatever, then env auto-repairing is done for that ST session. With the new ST patch we don't have the race between our `plugin_loaded` and their `env_loaded` bc `env_loaded` always fires before `loaded`. 

Also remember that even before this new patch we had valid reported logs from users where auto-repair didn't work (again probably timeout) and so SL never worked even after an hour.

braver:
Hm, ok. Just so I understand:

- `plugin_loaded` now means the env is ready
- this is the only trigger left that we can wait for
- if it triggers too soon, e.g. because time out, executables may never be found and linting will not work


The only way to solve this automatically is to enhance our heuristic and wait, perhaps indefinitely (or at least a user configurable high timeout), for our heuristic to return true.

I believe such a heuristic should not be part of SublimeLinter. A lot of other packages have the same problem. Not very many users have the problem though, the new ST patch should fix it for the vast majority. 
So it should be an optional thing and perhaps a third party package that fires an event we can listen for.

I believe that if your env takes a minute to load, that's a very special situation. In that case you could also manually configure paths to the executables and have linting instantly and without waiting. Sublime Text is all about performance and not having to wait for things to load. We have a manual configuration fall back for some scenario's and I believe this scenario fits that solution well enough too.

That said, if you want to spend more time trying to create a better heuristic (for an operating system you don't use I might add), I'm not going to stop you. I don't really like having it in our code base, but as long as it doesn't get in the way for people that don't have the problem it's fine by me.

kaste:
> the new ST patch should fix it for the vast majority

The key take-away here is for now, that ST probably or maybe regresses with the new patch.

braver:
> that ST probably or maybe regresses with the new patch

Maybe. I would like to see evidence of that though?

braver:
Alright.

The first one is using `zsh`, I've seen various reports of issues with that and it's not certain to be a regression. Still, looks like a SublimeText bug and if so we shouldn't work around but report it. Also, he was able to work around it.

The second one is super vague and it doesn't seem like the reporter understands anything about paths etc.

So, I don't think this is our problem anymore really.

braver:
Ok, so I confirmed with wbond that loading the env now times out after 1000 ms. Which is too soon for some users. However, we also cannot wait for it anymore, it simply bails out. So, whatever we do, our existing "solution" doesn't work anymore and we may as well merge this PR. And then we think a little bit about what we can offer for users with super slow envs.

PR #1395, opened 2018-05-11
Check `is_lintable` on startup before `hit`

### LINKED ISSUE #1393: AttributeError: 'NoneType' object has no attribute 'buffer_id'
Sometimes during startup I get this exception:

```
  File "/home/code/sublime_text_3/sublime_plugin.py", line 594, in on_selection_modified_async
    callback.on_selection_modified_async(v)
  File "/home/code/.config/sublime-text-3/Packages/SublimeLinter/highlight_view.py", line 247, in on_selection_modified_async
    if view.buffer_id() == active_view.buffer_id():
AttributeError: 'NoneType' object has no attribute 'buffer_id'
```

I looked into it a little and found that the `State['active_view']` doesn't have the active view in `"/home/code/.config/sublime-text-3/Packages/SublimeLinter/highlight_view.py", line 247`. I couldn't figure out why. It happens ~35% of the time.

kaste:
As reported https://github.com/SublimeLinter/SublimeLinter/issues/1393#issuecomment-388244401

Fixes #1393

### DIFF
--- a/busy_indicator_view.py
+++ b/busy_indicator_view.py
@@ -36,7 +36,7 @@ def on_begin_linting(buffer_id):
     State['running'][buffer_id] = time.time()
 
     active_view = State['active_view']
-    if active_view.buffer_id() == buffer_id:
+    if active_view and active_view.buffer_id() == buffer_id:
         sublime.set_timeout_async(lambda: draw(**State), INITIAL_DELAY * 1000)
 
 
@@ -45,7 +45,7 @@ def on_finished_linting(buffer_id, **kwargs):
     State['running'].pop(buffer_id, None)
 
     active_view = State['active_view']
-    if active_view.buffer_id() == buffer_id:
+    if active_view and active_view.buffer_id() == buffer_id:
         draw(**State)
 
 
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -13,7 +13,9 @@ Implement typical Goto Next Previous Error Commands.
 
 class SublimeLinterGotoError(sublime_plugin.WindowCommand):
     def run(self, direction='next', count=1, wrap=False):
-        goto(self.window.active_view(), direction, count, wrap)
+        active_view = self.window.active_view()
+        if active_view:
+            goto(active_view, direction, count, wrap)
 
 
 def goto(view, direction, count, wrap):
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -234,6 +234,10 @@ class IdleViewController(sublime_plugin.EventListener):
 
     def on_selection_modified_async(self, view):
         active_view = State['active_view']
+        # Do not race between `plugin_loaded` and this event handler
+        if active_view is None:
+            return
+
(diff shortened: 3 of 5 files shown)


kaste:
I cannot verify #1393

gerardroche:
`is_lintable` won't work because the view is None, you get:

```
  File "/home/code/sublime_text_3/sublime_plugin.py", line 283, in on_api_ready
    m.plugin_loaded()
  File "/home/code/.config/sublime-text-3/Packages/SublimeLinter/sublime_linter.py", line 56, in plugin_loaded
    if util.is_lintable(view):
  File "/home/code/.config/sublime-text-3/Packages/SublimeLinter/lint/util.py", line 80, in is_lintable
    not view.window() or
AttributeError: 'NoneType' object has no attribute 'window'
```

kaste:
Okay?

braver:
@kaste I don't know how to reproduce the original issue, but in my experience it never hurts to check view existence a lot so LGTM.

kaste:
I only guarded views where I could actually create a traceback.

gerardroche:
:+1: Looks fixed in 4.5.3.

gerardroche:
Actually, no, I have another AttributeError:

```
  File "/home/gerard/sublime_text_3/sublime_plugin.py", line 594, in on_selection_modified_async
    callback.on_selection_modified_async(v)
  File "/home/gerard/.config/sublime-text-3/Packages/SublimeLinter/panel_view.py", line 91, in on_selection_modified_async
    if active_view.buffer_id() != view.buffer_id():
AttributeError: 'NoneType' object has no attribute 'buffer_id'
```

Looks like a race-condition. It happens sometimes, I don't know why.

gerardroche:
My only suggestion would be to consolidate the root package modules into one root module and load everything from there. That way you have total control of the order that things get loaded. Create a module `plugin.py` and load everything from there: one `plugin_loaded()` and one `plugin_unloaded()`. I'd go further and consolidate the events into one event listener. It will give you more control of the flow of things and will also improve performance.

kaste:
Thanks for the reports here. We first fix it. Before the exact semantics aren't clear we cannot do an abstraction. 

As I started here a couple of months ago, I introduced the classical `State` vars for views. If you do this you immediately have the problem when and where to init that state. I thought the proper place for this is `plugin_loaded` as it would be executed *before* any other event handler. But this is not true.

However, as it seems the semantics aren't totally clear here. As I first wrote this I wanted to ensure via `plugin_loaded` that the State is initialized. This could be expressed by a decorator `@ensure_initialized`. But obviously it throws here and we just do nothing. This would more look like `@ignore_if_not_initialized`. Another variant would be a lazy `@start_after_activated_once`.

So I think we must first look if aborting does do any harm here or has the desired outcome.

To consider:

- Should `active_view` become `None` again for example if you activate a pic, or a view which is not lintable or probably already detached?

There are multiple ways to code and enforce such rules. We take 'less-coupled' (over `integrated`) as a very high value.

PR #1396, opened 2018-05-11
Maintenance: Deprecate unused npm_name and composer_name

kaste:


braver:
Oh, are those unused? Good to know.

kaste:
Hope so.

PR #1398, opened 2018-05-11
Guard against `theme_file` being None

### LINKED ISSUE #1397: Be ready to use after installation/enabling without restart
When enabling SL via Package Control I get an exception:

```
  File "/home/code/sublime_text_3/sublime_plugin.py", line 183, in reload_plugin
    m.plugin_loaded()
  File "/home/code/.config/sublime-text-3/Packages/SublimeLinter/sublime_linter.py", line 48, in plugin_loaded
    style.read_gutter_theme()
  File "/home/code/.config/sublime-text-3/Packages/SublimeLinter/lint/style.py", line 23, in read_gutter_theme
    theme_file = os.path.basename(theme_path)
  File "./python3.3/posixpath.py", line 142, in basename
AttributeError: 'NoneType' object has no attribute 'rfind'
```

Ubuntu, ST 3173, SL 4.5.1

Reproduce: 

* Disable SL via PC
* Restart ST
* Enable SL via PC

kaste:
Fixes #1397

### DIFF
--- a/lint/style.py
+++ b/lint/style.py
@@ -19,11 +19,14 @@ def get_linter_style_store(name):
 
 
 def read_gutter_theme():
-    theme_path = persist.settings.get('gutter_theme')
-    theme_file = os.path.basename(theme_path)
     global COLORIZE
     COLORIZE = True
 
+    theme_path = persist.settings.get('gutter_theme')
+    if not theme_path:
+        return
+
+    theme_file = os.path.basename(theme_path)
     if not theme_file.endswith(".gutter-theme"):
         theme_file += ".gutter-theme"
 


braver:
It’s probably better to (also) prevent an empty string via the schema validation, using [minlength](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.3.2) or a pattern. The schema validation also gives users feedback about their erroneous setting.

kaste:
An empty string would not produce the reported traceback. Also, don't know if we changed here, but the empty string was always a correct setting. Not on my computer, but since I don't like theming, it is probably my setting.

kaste:
This none check here is hairy. OP clearly indicates that this is an init problem if you enable SL. So it seems the settings are not yet ready at that time.

So if this fix fixes the traceback thrown, it does not mean the whole process of just enabling SL without a need to restart is solved.

braver:
> Also, don't know if we changed here, but the empty string was always a correct setting.

No, empty string has never been a valid setting and it's definitely not what you have. If you haven't set it, we get the merged value back from Sublime Text and it'll be "Default". If not, you're icons will look like this:

screen shot 2018-05-12 at 13 53 13

Of course, it's "valid" in so much that it won't crash, but it'd definitely not what you've got on your computer.

But ok, I understand and agree that, that this is besides the point.

> So if this fix fixes the traceback thrown, it does not mean the whole process of just enabling SL without a need to restart is solved.

Could be. I also get this one:

```
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 233, in synthesize_on_activated_async
    el.on_activated_async(v)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 142, in on_activated_async
    hit(view)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 239, in hit
    delay = get_delay(reason)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/sublime_linter.py", line 414, in get_delay
    max(MIN_DEBOUNCE_DELAY, float(persist.settings.get('delay'))),
```

Personally I'm not interested in solving all problem that may occur following the steps laid out in #1397. Restart ST after enabling SL and *then* look if there are errors. You'll end up guarding against everything maybe being `None` at some point. That said, I don't have anything against this particular PR, so my opinions  are just my opinions and this seems good to merge.

kaste:
Okay, checked. This is not sufficient. It seems that the settings are not ready. ST core bug? If we fix here, we get e.g.

```
  File "C:\Dev\Sublime Text 3\sublime_plugin.py", line 611, in on_activated_async
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\sublime_linter.py", line 146, in on_activated_async
    hit(view)
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\sublime_linter.py", line 243, in hit
    delay = get_delay(reason)
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\sublime_linter.py", line 418, in get_delay
    max(MIN_DEBOUNCE_DELAY, float(persist.settings.get('delay'))),
TypeError: float() argument must be a string or a number
```

If you just enable SL in a running system, you can wait but still get

```
sublime.load_settings("SublimeLinter.sublime-settings").get('debug')
> None
```
 in the console.

kaste:
I have to actually open the settings and save to get them (re-) loaded. Do you know something here. Ask around?

kaste:
I have `gutter_theme: ""` but only use the builtin icons as well, and it works. (Now, of course, I can remove that setting bc I can still get the builtin icons with a theme. But I probably didn't knew that back then.)

kaste:
> Restart ST after enabling SL and then look if there are errors. You'll end up guarding against everything maybe being None at some point. That said, I don't have anything against this particular PR, so my opinions are just my opinions and this seems good to merge.

No, this is like a feature request. Is it possible to live enable/disable SublimeLinter without restarting.

braver:
I could look into it some more but the weather is too good 😎

kaste:
Closing here, all settings return None after enabling SL via Package Control.

PR #1403, opened 2018-05-13
`cls.disabled` takes precedence over settings

kaste:
Let `cls.disabled` have three states (None, True, False). None
being the default will ask for user settings, a boolean value
though will overwrite any other user settings.

This allows to activate a linter plugin *in-memory* which is
otherwise disabled via user settings.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -220,7 +220,7 @@ class LinterMeta(type):
             return
 
         name = attrs.get('name') or cls_name.lower()
-        setattr(cls, 'disabled', False)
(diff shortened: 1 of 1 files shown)


braver:
Can you explain the use case to me more clearly? It sounds good though.

kaste:
E g. You can install the sublime syntax linter and disable it 'globally' via the user or project settings. With this PR you can still enable it in-memory using e.g. my toggler plugin.

braver:
Oh, nifty. 👍🏻

PR #1406, opened 2018-05-14
Apply col_offset only on line 0

kaste:
Fixes https://github.com/SublimeLinter/SublimeLinter-eslint/issues/249


kaste:
It's a regression, we had this working last month

PR #1411, opened 2018-05-16
'Refactor' style.py

### LINKED ISSUE #1333: Refactor style.py to not use a `style` attribute as main key
Currently style.py computes a `style` (str) which is an intermediate value passed around. The values we're actually interested in are 'scope', 'icon', and 'marks_style'. `style` is either an implementation detail (hidden/encapsulated state) or just not necessary data transformation.

### LINKED ISSUE #1334: Make error priority available to all views
Currently we have a priority system used in highlights_view. It would be better if we could assign the error priority in `backend.finalize_errors` so that all views e.g. the statusbar view can sort errors also by priority and not just by severity.

Unfortunately we must refactor style.py for this. #1333

kaste:
Fixes #1333 
Fixes #1334

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -49,7 +49,8 @@
 
     // Linter specific settings.
     // More info: http://www.sublimelinter.com/en/stable/linter_settings.html
-    // Linter specific settings can also be changed in sublime-project settings.
+    // Linter specific settings except for "styles" can also be changed
+    // in sublime-project settings.
     // What settings are available is documented in the readme of the
     // specific linter plugin.
     // Example:
@@ -57,7 +58,12 @@
     //     "somelinter": {
     //         "disable": false,
     //         "args": [],
-    //         "excludes": []
+    //         "excludes": [],
+    //         "styles": [
+    //              {
+    //                  "icon": "diamond"
+    //              }
+    //         ]
     //     },
     //     "somepythonlinter": {
     //         "python": null,
@@ -115,7 +121,7 @@
     "statusbar.show_active_linters": true,
 
     // Global styles for all linters.
-    // Styles can also be specified per linter.
+    // Note: Styles can also be specified per linter! See above.
     // - mark_style:
     //   - "none"
     //   - "fill", "outline",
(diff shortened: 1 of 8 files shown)


braver:
I pushed some related documentation if you don't mind. Looks good this. This is always a good sign:

screen shot 2018-05-17 at 10 09 05

braver:
I've played around with the priorities a bit and noticed that setting priority to 0 kills the highlights completely for all linters, looks like a bug (a stray "falsy" check?)

Here I have a syntax error highlighted in the same position by both flake8 and pyflakes. I cycle the priority for flake8 from 0 to 2 and back. At 0, you see no highlights at all, not even for pyflakes. At some point you'll see a red icon (pyflakes) and a blue highlight (flake8). If I stop there and trigger "lint this view" manually, it sorts itself out so doesn't seem to be a big problem and may just be ST not re-rendering the gutter.

styles

kaste:
Made some changes. We now use the defaults we ship as fallback. Or you could say we merge them in. This basically means you only have to override what you actually want to change. E.g. these are my global user styles.

image

Here I don't define an icon bc I like the default, and of course I don't choose a default priority.




Regarding the bug, I have the following:

bcqk5h7euk

braver:
> We now use the defaults we ship as fallback.

Excellent!

> Regarding the bug, I have the following:

Priority 0 still takes out that highlight though. I expect 0 to just be an integer that's lower that 1.

kaste:
Okay, in the screencast I change the faster linter (pyflakes). If I change the priority setting of the always slower linter (flake8) I can reproduce a bit of it.

kaste:
That's basically bc we don't reassign priorities before we get new results from the linter. So as the faster linter arrives it compares its values to the old priority of the old results from the slow linter.

kaste:
Fixed!

kaste:
Finally: make `types` optional. If you omit `types` (and `codes` for the linter styles) it will just match all error types. This reduces invalid states (or possible user errors).

kaste:
So my new global styles look like this

image

It is also valid to just promote `pyflakes` via

```
linters:
	pyflakes:
		styles:
			priority: 2
```

braver:
Looking good then. Let's ⛵️

SublimeLinter.sublime-settings
```diff
@@ -115,31 +115,34 @@
...
-    //      "Packages/SublimeLinter/gutter-themes/Blueberry Cross/error.png"
-    //   or one provided by a gutter theme (e.g. "warning" or "error")
-    //      in theme Default: "warning", "error", "cog", "x",
-    //      and "diamond", "heart", "pointer", "square", "star", "triangle" all
-    //      also have an -outline variant.
-    // Note: At least one of these targets is required: types or codes.
-    // Styles can also be specified per linter.
+    // - priority:
+    //   Determines which mark is shown in the gutter;
+    //   error marks will still always override warning marks.
```
kaste: This is incorrect. For 'problems' with the same priority, error will win over warning. You can promote a warning by giving it a higher priority.

braver: Good catch, that's indeed no longer true.

SublimeLinter.sublime-settings
```diff
@@ -115,31 +115,34 @@
...
+    // - types:
+    //   An array which can contain "warning" and/or "error".
+    //   If omitted will match both.
+    // - codes:
+    //   An array which can contain error codes provided by a linter.
```
kaste: Only valid as a linter style!

kaste:
🎉

braver:
💪🏻

Issue #1230, opened 2018-03-18
Refactor around lint modes


braver:
Originally posted in #1222 by @kaste:

I would like to see a change here for background mode as well and change the behavior of the BackendController.

We can have three modes (background|save|manual) plus a flag `lint_on_load(true|false)`. Basically I could set background and lint_on_load to false for navigating through a code base, and linting would begin when I begin to write code. 

Also to change: we hit on save in background mode which is not necessary.

```
	on_activated => skip if lint_on_load == false or lint_mode == manual
    on_save => skip if lint_mode != save
    on_modified => skip if lint_mode != background
```

This becomes a bit more complicated for

- when saving a project file fire `config_changed` event
- on_activate check if the syntax changed. For convenience we hit here bc there is a good chance that the assigned linters changed as well

Notable: if the config changes, we currently hit every view regardless of the lint_mode

kaste:
This would help me a lot/sometimes. I don't know if we can do this as easy since we have the reason and lint_mode per project/linter thing. 

The use case is simple:

- Just browsing reading a code base. 
- E.g. Find/replace. If you do this on more files and forgot to disable SL ... you hate yourself

braver:
> Nothing changes "on_save" when "on_modified" already
triggered?

Perhaps for linters that only run on real files (ie. not temp files)?

I do like the idea of going from the predefined behaviors to letting the user specify exactly which events to listen to. Both kinds of settings could coexist IMO (and will have to for compatibility). And if the current “behaviors” already expand to those events behind the scenes, that just makes sense. You’re just letting the user control that directly.

kaste:
> > Nothing changes "on_save" when "on_modified" already
> triggered?
> 
> Perhaps for linters that only run on real files (ie. not temp files)?

That's the beauty of "named" modes.  They hide the details and encode some
useful, special knowledge.  (Looks like "on_save" should only run for these real
file linters.)

PR #1416, opened 2018-05-18
Support lint_mode option for specific linters

bongnv:
This is an attempt PR to close this task #244 . In this PR, I add method `should_lint` in the base linter to support:
- decides whether the linter should start or not based on a given reason
- check `lint_mode` with `reason` to decide whether it should check the given view or not
- allows each custom Linter to programmatically decide whether it should take action on each trigger or not
Docs also is added but not sure if it has the correct format. Please help to verify.

### DIFF
--- a/docs/linter_settings.rst
+++ b/docs/linter_settings.rst
@@ -82,6 +82,15 @@ be a string or a list.
 See :ref:`Settings Expansion <settings-expansion>` for more info on using variables.
 
 
+lint_mode
+---------
+Lint Mode determines when the linter is run.
+- background: asynchronously on every change
+- load_save: when a file is opened and every time it's saved
+- manual: only when calling the Lint This View command
+- save: only when a file is saved
+
+
 python
 ------
 This should point to a python binary on your system. Alternatively
--- a/docs/linter_settings.rst
+++ b/docs/linter_settings.rst
@@ -85,10 +85,10 @@ See :ref:`Settings Expansion <settings-expansion>` for more info on using variab
 lint_mode
 ---------
 Lint Mode determines when the linter is run.
-- background: asynchronously on every change
-- load_save: when a file is opened and every time it's saved
-- manual: only when calling the Lint This View command
-- save: only when a file is saved
+* `background`: asynchronously on every change
+* `load_save`: when a file is opened and every time it's saved
+* `manual`: only when calling the Lint This View command
+* `save`: only when a file is saved
 
 
 python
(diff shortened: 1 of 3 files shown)


kaste:
Very cool! Thank you. 

As you can see I already prepared implementing that feature with the `reason` arg we can pass around. Now, the idea was to finally decide if we actually *will* lint at this time in `skip_linter` within `sublime_linter.py`. 

I think, we cannot use `get_linters_for_view` which then asks `can_lint_view` bc `get_linters_for_view` has enormous side-effects (You may have read the comment there "Yeah, it's a *getter* LOL.") and if it doesn't return a stable set of linters we erase all collected lint errors all the time. Also the status bar indicating which linter has been assign to the view (e.g. showing 'eslint(ok)') would probably flicker, change all the time.

Within `skip_linter(linter, view) -> bool` you can get the settings using the function `get_linter_settings(linter, view)` from `linter.py`.

bongnv:
Hi @kaste,
Thanks for reviewing. I've updated the PR. Basically, it renames func `skip_linter` to `should_lint` and move it base linter implementation. By doing this, it allows each custom linter to programmatically decide whether it should lint the attached view based on a trigger or not.

kaste:
That's perfect. :+1:

kaste:
Thanks again. 

Do you want to further iterate on this topic? Your PR allows 'demoting' the globally `lint_mode`. T.i. if the user is in background-mode globally, he can now declare a specific linter to be 'only' in 'on_save' mode. But not vice-versa. 

So, the next iteration would be that a specific 'lint_mode' setting will overrule (take precedence) over the global user settings. 

The idea I had is to just remove the prechecks in the `BackendController` in `sublime_linter.py`, and then take the global 'lint_mode' in `should_lint` as a fallback.

WDYT?

kaste:
Related #1230

We want to change lint_mode to only (background|save|manual). This matches the reasons (None|on_save|on_user_request) and in turn the events (on_modified|on_save|LintViewCommand).

We introduce a new setting lint_on_load(true|false) which we check in the event handler on_activated which will then in turn hit with the reason on_user_request. 

So the complete feature set is in the end a breaking change.

This is just to give a broader perspective. Take whatever you want and ask questions if you have any.

braver:
@bongnv we got some donations recently and I would like to share some of it with you. Do you have a PayPal address?

bongnv:
Hi,
I appreciate that. I'm just good. Please give it to who need. Thanks.
Bong

braver:
👍🏻

Issue #1312, opened 2018-04-10
Optimimize what to do when `config_changes` fires


kaste:
Just linting all views makes me want to not change the settings bc it makes my fan sing. I know we inherited this, but if we write down what we actually want to achieve we probably can code it (more efficient).

```
if `gutter_theme, highlights.*, show_marks_in_minimap, styles` change => redraw highlights
if `linters, no_column_highlights_line` changes => re-lint
otherwise => do_nothing
```

- [ ] If the project file config triggers it is enough to only re-lint views from the current related window.  
- [ ] `lint_all_views` is completely misnamed bc it doesn't do that, although an actual command to lint all view (of the current window) would be probably nice.

FichteFoll:
Well, the settings affect a lot of aspects of SL wrt highlighting, linter settings, etc., so just re-running all linters is definitely the easiest way to pick up those changes.

braver:
Exactly. You may have changed the styles, or the rules (excludes, args, etc) with which the linters run. Especially if you’re running file-only linters, their reports probably don’t match your settings anymore. You could try to interpret the impact of each setting individually and update the state accordingly, or you simply tear down and rebuild the state. 

User stories:

- I lint a file, get red highlights.
- Change the style to blue.
- Switch back to the file and I expect the highlights to be blue.

- I lint a file with some args.
- Change the args in settings.
- Get some coffee, talk to a colleague.
- Get back to the file, and want to be sure the way it was linted matches the current settings.

The assumption is that there is no need to be efficient/performant here usually, except perhaps if you have lots of files open and you don’t want your laptop to heat up. Another assumption is that you set your settings, perhaps tweak it later, and you’re done with them. Unless you’re a SL dev and are testing various settings, and have way too many linters installed, of course. 😉

kaste:
Hm, 

```
if `gutter_theme, show_marks_in_minimap, styles` change => redraw highlights
if `linters, no_column_highlights_line` changes => re-lint
otherwise => do_nothing
```

Of course it is relatively likely that I change the project linter settings `args, executable`. I don't know if these settings can be observed at all, but of course we do nothing here. So the user usually does the `<space>,<backwards>`-is-it-working dance.

kaste:
EDIT: we already do a lint all when a project file changes.

braver:
@kaste can you phrase this issue in a way that’s actionable, or is there more you’d like to discuss?

kaste:
[...]

PR #1421, opened 2018-05-23
lint_mode in linter settings should overwrite global one

bongnv:
Following the conversation here https://github.com/SublimeLinter/SublimeLinter/pull/1416, this diff centralizes logic of `lint_mode` to function `should_lint`. It aims to increase the visibility of how `lint_mode` work in order to make it easier to maintain or refactor (including add `lint_on_load ` settings).
The diff also updates documentation render errors about `linter_mode` settings.
It's expected to have some changes in existing behaviour of how linters are created for each view. However, the change should be minimal or none to end users.


kaste:
👍 That looks really good.

lint/linter.py
```diff
@@ -847,12 +858,8 @@ def should_lint(self, reason=None):
             "checking lint mode {} vs reason {}"
             .format(lint_mode, reason)
         )
-        if lint_mode == 'manual':
-            return reason == 'on_user_request'
-        elif lint_mode in ('save', 'load_save'):
-            return reason in ('on_save', 'on_user_request')
 
-        return True
+        return reason in _ACCEPTABLE_REASONS_MAP[lint_mode]
```
kaste: You actually forgot to lookup the global `lint_mode`.

sublime_linter.py
```diff
@@ -234,7 +223,7 @@ def lint_all_views():
     for window in sublime.windows():
         for view in window.views():
             if view.buffer_id() in persist.view_linters:
-                hit(view)
+                hit(view, "on_load")
```
kaste: This one only gets called when the config changes. (Which in itself is not optimized yet #1312) And we want 'on_user_request' here.

bongnv:
Hi @kaste ,
All comments should be resolved. Please take a look when you get a chance.

PR #1422, opened 2018-05-24
Nicer cmd printing

kaste:
Super little detail here. We use `subprocess.list2cmdline(cmd)` for fromatting on Windows bc it handles quoting.


braver:
👍🏻

PR #1424, opened 2018-05-24
Showcase: Allow post filters via setting `filter_errors`

kaste:
`filter_errors` takes a list of regexes which will get matched against
the string "{code}: {message}".

Typical usage e.g. `"filter_errors": ["W291"]` for flake8

This is a 15 line implementation.

Closes https://github.com/SublimeLinter/SublimeLinter-golint/pull/8
Fixes https://github.com/SublimeLinter/SublimeLinter-html-tidy/issues/39
Fixes https://github.com/SublimeLinter/SublimeLinter-html-tidy/issues/52

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -81,6 +81,18 @@
 
             "excludes": [],
 
+            // Either a string or a list of strings. Each input string is
+            // handled as a case-insensitive regex pattern and matched against
+            // the error_type, code, and message of a particular lint error.
+            // If it matches, the error will be thrown away.
+            // E.g. "warning: " will remove all warnings but keep errors.
+            // "no-trailing-spaces: " will suppress this rule known from eslint.
+            // "W3\d\d: " will suppress some whitespace errors known from flake8
+            // and pyflakes.
+            // "missing <!DOCTYPE> declaration" will suppress that error known
+            // from htmltidy.
+            "filter_errors": "",
+
             // Lint mode determines when the linter is run. The linter setting
             // will take precedence over the global setting.
             "lint_mode": "manual",
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -53,7 +53,7 @@
 
     // Linter specific settings.
     // More info: http://www.sublimelinter.com/en/stable/linter_settings.html
-    // Linter specific settings except for "styles" can also be changed
+    // Linter specific settings except for 'styles' can also be changed
     // in sublime-project settings.
     // What settings are available is documented in the readme of the
     // specific linter plugin.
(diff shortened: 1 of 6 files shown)


braver:
So, this would be the return of the `ignore_match` setting? We removed this from SublimeLinter 4 because people were using this instead of the configuration options already available from the linters themselves (e.g. setup.cfg). 

I don't understand why you would want such a feature for flake8 which is already fully configurable.

For something like html-tidy I do get it, it has no configuration features by itself and a lot of what it reports is garbage while the important stuff is actually useful. That said, it's not our job to fix the stupid linter and fixing it at the front-end is not the right place to do so.

So, I have my reservations here, regardless of how many lines it is. On the other hand I don't care if people want to make a crazy mess of their configuration as long as I don't have to know these people exist. 🤷‍♂️ 

In any case, if we do ship it, it needs to be documented.

kaste:
A bit strawman/showcase here, but effort-impact-balance here is extremely important. 

- Now the more general problem is, that we have lint rules which are enforced via CI or pre-merge hooks etc. but between the code we write and the code the CI checks are probably auto-fixers. So the lint conf between CI and IDE should be different.

- The lone developer also cannot change the lint conf bc it is enforced from above or by the team.

We don't have to decide immediately. I just thought (effort-impact) let's see what can you do in really little code. (Actually bc I saw the old implementation for SL3. You removed 180 (!) lines back then.)

braver:
Yeah, I do see that case where you want control over the info you get in your editor vs config that may be enforced or different for CI. I personally deal with that via styles. But you were able to do this in 15 lines, so I’m ok with it. I don’t like adding features, but 🤷🏼‍♂️

That said, the feature has been gone for months by now. It’s mostly because tidy is stupid that we’re seeing demand for it at all.

braver:
Ok, so, I made my point and I’ll let the decision to you. If you’re happy with it and want to write a few lines in the docs about it, be my guest. If not, that’s cool too.

kaste:
I just move this to the filter-addon, no? Or yet another plugin? WDYT, @braver ?

braver:
If it’s easy to do as an addon, go for it. I don’t mind merging it either though, it’s a good feature to have as long as people understand what it’s for.

braver:
@kaste https://packagecontrol.io/packages/SublimeLinter-addon-filter covers this now right?

kaste:
Not really. Then I would have closed this immediately. The add-on is dynamic, in memory and global.

This one is just an additional setting. It's per linter and mostly static.

I think I use the same technique in the black-for-flake add-on. It was meant as a way to configure a linter basically for fixers where you can't change the linter config itself.

braver:
Ok, so, let's merge this then?

kaste:
No, don't think so. 

* We don't have upvotes, reactions yet.
* Since then, I have shown for flake8 (unreleased PR 🙄) how to suppress what Sublime fixes (trailing ws etc)
* I have made black-for-flake (auto-configure instead of manual config)
* I have shown (unreleased PR) for eslint how we could mute what eslint can fix.

I've not tested (bc bugs I fixed today) to visually mute codes via styles. Muting via styles has the advantage that the errors are still in the panel. Whereby for black I completely filter some errors bc black will fix them. 

E.g. if you see this ugly screenshot. https://github.com/SublimeLinter/SublimeLinter-eslint/issues/212#issuecomment-372770596

I think it would be better to visually mute the 'fit' error, which means the buffer is still red and it will show up in the error panel. If we just filter the error, the buffer becomes 'green', ready to commit, which is not what the user wants. 

The PRO is: I really like the implementation. 

So, we still wait, if we can have less noise/distraction without this PR. If we **need** a solution for html tidy (I don't use it) then maybe we should just merge it.

kaste:
(The most controversial 15 lines in SL history.)

braver:
Ok 👍🏻

About the ugly screenshot though. It’s not super nice to look at, but it’s an honest visualisation of the error report. I find that suppressing the highlights while editing really works in these situations. 

Hard filtering shouldn’t be used because you don’t like how an error is presented, only because you’re really not interested in the error.

kaste:
LGTM. However, I did not do a functional test, I only used the tests to verify.

kaste:
@braver? 😴

braver:
Sorry, I've been busy with the SCSS syntax highlighting that I broke for a couple thousand users... It's what I'm good at, breaking things.

So, I tested this with some JS and eslint and it works really well. I was able to filter on the code and the message, or parts of the message. The work I'm doing right now probably doesn't mean I can do really intensive testing. But it's a small changeset, we've got test coverage... what's not to like?

So, IMO all this needs is an entry in the default settings and a word or two about it in the docs, and we can ship it. I think it's a good feature to have, HTMLtidy users will rejoice.

kaste:
C'mon, some team work. Added some doc stuff for you to fine tune. 🎄

braver:
Ok, I've got most of the other stuff off my plate. @kaste I'll review the docs etc., merge and then 🛥

kaste:
A SL 🎅 edition

SublimeLinter.sublime-settings
```diff
@@ -64,34 +64,31 @@
...
-            // If it matches, the error will be thrown away.
-            // E.g. "warning: " will remove all warnings but keep errors.
-            // "no-trailing-spaces: " will suppress this rule known from eslint.
-            // "W3\d\d: " will suppress some whitespace errors known from flake8
-            // and pyflakes.
-            // "missing <!DOCTYPE> declaration" will suppress that error known
-            // from htmltidy.
-            "filter_errors": "",
+            // Suppress errors that match the the given pattern(s).
+            // Either a 'string' or an 'array'. Each input string is handled as
```
kaste: Haha, you removed the quotes everywhere, and then added them here.

braver: Oh damn 😂

PR #1426, opened 2018-05-24
Update default icon

mathieug:
x looks better to show a warning or error in the gutter. It also stands out more than Modific's icons.


braver:
We default to one of Sublime Text's built-in icons to match the default look and feel. Wether one wants the icon to stand out more or less is a personal preference. Also, we can't suddenly change everyone's icons even if we decide the x looks better.

PR #1427, opened 2018-05-24
Showcase: Live filter errors

### LINKED ISSUE #1255: Should we allow filtering the panel?
The filtered panel (e.g.
```json
  { "keys": ["ctrl+k", "ctrl+a"], "command": "sublime_linter_panel_toggle",
    "args": {
      "exclude_codes": [],
      "exclude_linter": ["annotations"]
    },
  },
```
) does

1. not hide views that have all errors fitered and don't show anything,
2. show the current position indicator in ... weird places.

Interestingly, the position indicator even updates when I move the caret past a filtered report.

img of panel

kaste:
It's 🤹‍♂️ 🤹‍♀️ time again.

Provide command 'sublime_linter_filter' which takes one optional
argument 'pattern'. If omitted a TextInput will open instead (ST >3.1).

Search for 'SublimeLinter: Filter Errors'.

E.g. '-W\d+', '-annotations', 'flake E999'

Has lint errors for easy testing. Implementation doesn't matter.

Closes #1255

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -25,6 +25,11 @@
 
     },
     {
+        "caption": "SublimeLinter: Filter Errors",
+        "command": "sublime_linter_filter"
+
+    },
+    {
         "caption": "SublimeLinter: Reload SublimeLinter and its Plugins",
         "command": "sublime_linter_reload"
 
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -302,20 +302,125 @@ def kill_active_popen_calls(bid):
         proc.friendly_terminated = True
 
 
+import os, re
+
+
+PASS_PREDICATE = lambda _: True
+LIVE_FILTER = {
+    'store': defaultdict(list),
+    'filter': PASS_PREDICATE,
+    'user_value': ''
+}
+
+
 def update_buffer_errors(bid, view_has_changed, linter, errors):
     """Persist lint error changes and broadcast."""
     if view_has_changed():  # abort early
         return
 
-    all_errors = [error for error in persist.errors[bid]
+    all_errors = [error for error in LIVE_FILTER['store'][bid]
                   if error['linter'] != linter.name] + errors
-    persist.errors[bid] = all_errors
+    LIVE_FILTER['store'][bid] = all_errors
+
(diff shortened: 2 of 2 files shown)


braver:
Oh, so that’s what you can do with that new feature. Now *that’s* cool bro.

kaste:
@FichteFoll Can you check this for UX. Code review is not relevant bc I just did this as fast as possible.

braver:
Man this is *nice*. I kinda want to be able to filter for "err" and only see the errors though. Also because I'm looking at the panel to figure out how to filter and "error|warning" is part of the string there too. 

It's not 100 clear how the filtering works... E.g. perhaps I'd like to be able to use a regex and be super specific, but I'm not sure what the string it's checking against and whether regexes are supported at all. I mean, from the UI I don't get that, but perhaps it's ok if it's in the docs somewhere. It's easy enough to reduce the list to find what I'm looking for though.

So, I really like how well this works. I personally don't think I'll ever use it, so I do get my "I don't like adding features" thing, but dang it's nice 🍷

kaste:
Maybe this is an add-on again. Errors/warnings filter would be nice. Inline help could be done I think via the preview thingy built into these new input handlers.

What's nice is that you can also add key bindings to filter all annotations etc. So you don't have to type in the filter string all the time. Maybe regexes are overkill.

braver:
> What's nice is that you can also add key bindings to filter all annotations etc.

👍🏻👍🏻

> Maybe regexes are overkill.

Probably.

I like spinning things off into addons because it makes them less official and we can drop support later on more easily. Niche features like this are really nice to have, but I don’t get a sense that people need them and will use them a lot. Here, I think again annotations is the big use case and not much else. 

And I’m kinda scared of increasing the complexity and maintenance load of SL. Especially with so few maintainers and outside contributors.

FichteFoll:
Just a heads up that I saw this but didn't have any time whatsoever to spend on testing, let alone developing on itself. I'll report back once I do.

kaste:
This is now a plugin. https://github.com/kaste/SublimeLinter-addon-filter

braver:
Cool! 🧚🏼‍♂️

PR #1430, opened 2018-05-24
Do not show error panel on unhandled exceptions

### LINKED ISSUE #1420: Updating a plugin can crash running lint processes
As reported https://github.com/SublimeLinter/SublimeLinter-eslint/issues/251#issuecomment-390678923

Package Control updates a SL-plugin. Bc of live-reload running threads (lint processes) become orphaned in way they don't even see their top-level imports. So they crash, and SL prompts an error panel although after the reloading everything should work okay.

We don't handle this right now. We only listen for SL updates and silent the logger during that time. 

https://github.com/SublimeLinter/SublimeLinter/blob/95fa803aa2eb3eea586dd79c2ccd0f212ccdf399/sublime_linter.py#L72-L81

Instead we should probably listen for 'in_process' changes (coming from Package Control) and silent the root logger as long as some of our plugins are 'in_process'.

### LINKED ISSUE #1428: Yet another crash: list.remove(x): x not in list
OS: Windows 10
Sublime: 3.1.1
SublimeLinter: 4.6.2

Before posting this stack trace, I would like to offer my unfortunate opinion that I've become very disappointed with this project in the last two months. It seems that every other day, the linter is crashing or upgrading itself (requiring a restart) or both. SublimeLinter used to be a stable product and recently it's gone down the toilet. I'm considering ditching Sublime Text altogether because of it. Sorry for the harsh comments but I think you need to hear how your users feel.

Anyway, I opened ST this morning, opened a file, and the linter crashed immediately:

```
  File "C:\...\SublimeLinter.sublime-package\lint/backend.py", line 87, in execute_lint_task
    errors = linter.lint(code, view_has_changed, settings) or []
  File "C:\...\SublimeLinter.sublime-package\lint/linter.py", line 755, in lint
    output = self.run(cmd, code)  # type: util.popen_output
  File "C:\...\SublimeLinter-eslint.sublime-package\linter.py", line 115, in run
    return super().run(cmd, code)
  File "C:\...\SublimeLinter.sublime-package\lint/linter.py", line 1079, in run
    return self.communicate(cmd, code)
  File "C:\...\SublimeLinter.sublime-package\lint/linter.py", line 1090, in communicate
    return self._communicate(cmd, code)
  File "C:\...\SublimeLinter.sublime-package\lint/linter.py", line 1216, in _communicate
    raise TransientError('Friendly terminated')
  File "./python3.3/contextlib.py", line 55, in __exit__
  File "C:\...\SublimeLinter.sublime-package\lint/linter.py", line 1244, in store_proc_while_running
    persist.active_procs[bid].remove(proc)
ValueError: list.remove(x): x not in list
```

kaste:
Although they are a bit useful for developers, these scare and annoy users.

Esp. during live-reload we cannot guarantee to not throw, at least atm. So
popping up the panel in your face is not even appropriate bc the reloading
can still succeed. (These are really transient errors.)

If we demote them, we still log them to the console. Hopefully, users still
see and report non transient errors this way. However, edge-case errors often
*appear* transient to the end user, and we will probably cannot fix those as
quickly or ever. E.g. we actually fixed year old bugs bc of this offensive
throwing.

Closes #1420
Closes #1428

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -93,7 +93,7 @@ def execute_lint_task(linter, code, offset, view_has_changed, settings, task_nam
         except Exception:
             linter.notify_failure()
             # Log while multi-threaded to get a nicer log message
-            logger.exception('Linter crashed.\n\n')
+            logger.exception('Unhandled exception:\n', extra={'demote': True})
             return []  # Empty list here to clear old errors
 
 
--- a/log_handler.py
+++ b/log_handler.py
@@ -120,6 +120,9 @@ shown_error_messages = defaultdict(set)
 
 class ErrorPanelHandler(logging.Handler):
     def emit(self, record):
+        if getattr(record, 'demote', False):
+            return
+
         try:
             msg = self.format(record)
             lines = msg.splitlines()


braver:
Totally agree here 👍🏻👍🏻

braver:
The pull request they do not stop!

kaste:
> The pull request they do not stop!

image

Really wonder how SL3 was developed. It had just no commit flow. It's a spike and then mostly nothing.

braver:
Just one person’s manic month it seems.

PR #1435, opened 2018-05-28
Inject settings on init

### LINKED ISSUE #1425: Change linter_class.__init__ signature
Currently the signature is `linter_class.__init__(view, syntax)`. However, `syntax` is unused and should be removed here. The new sig should be 

```
	def __init__(self, view, *, settings)
```

Since linter instances are now throw-away objects, we can inject `settings` immediately. 

PRO

- No need for `linter_context` (threadLocal) anymore.
- The settings are available as soon as we meet an instance. (Currently, the settings are only avail after the backend started `lint()`.
- settings are *consistent* for the life-time of the instance

CON

- Breaking change, although I never met a plugin overriding `__init__`

kaste:
- Change `__init__` sig from `(view, syntax)` -> `(view, settings)`
- Change `can_lint_view` sig from `(view)` -> `(view, settings)`
- Change `lint` sig from `(code, view_has_changed, settings)` -> `(code, view_has_changed)`
- Remove `lint_context` threading local

Overall slight refactorings

Fixes #1425

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -54,7 +54,7 @@ def run_tasks(tasks, next):
 
 
 def get_lint_tasks(linters, view, view_has_changed):
-    for (linter, settings, regions) in get_lint_regions(linters, view):
+    for (linter, regions) in get_lint_regions(linters, view):
         tasks = []
         for region in regions:
             code = view.substr(region)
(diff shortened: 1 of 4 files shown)


braver:
I'm not sure I'm doing a test with good coverage of this PR, but it seems to work and looks good. Ship it?

kaste:
This is a good prerelease candidate bc you just can't be sure I don't fuck it up somewhere for someone. Even though I now have seen a lot of linters.

braver:
#1427 too... I'll try and put out a pre-release later with these features.

kaste:
Sure #1427. The showcase thingy?

PR #1449, opened 2018-06-11
Use standard Sublime settings for the linter settings

### LINKED ISSUE #1446: Use standard view settings where appropriate
The [docs](http://www.sublimelinter.com/en/stable/settings.html#settings-stack) state:

> SublimeLinter merges settings from several sources to calculate the value. Settings are merged in the following order:
> 1. Default settings
> 2. User settings
> 3. Project settings (only “linters” settings)

Would it be possible to include "syntax-specific" settings between User settings and Project settings?

My use case is the following:
- I'm using https://github.com/asbjornenge/Docker.tmbundle/pull/24 to embed ShellScript in Dockerfiles
- I'm using https://github.com/SublimeLinter/SublimeLinter-shellcheck for Shell linting
- ShellCheck requires a shell to be specified in order to provide suggestions, and complains with: _"Tips depend on target shell and yours is unknown. Add a shebang. [SC2148]"_
- Adding a shebang to a Dockerfile does not make sense

--- 

I tried to solve this by adding the following to the Dockerfile syntax-specific settings:
```
  "linters": {
    "shellcheck": {
      "args": "--shell=sh"
    }
  }
```

But it had no effect. I'm assuming this is because syntax-specific overrides are ignored?

kaste:
Fixes #1446

### DIFF
--- a/lint/settings.py
+++ b/lint/settings.py
@@ -197,8 +197,8 @@ def print_deprecation_message(settings):
     import json
 
     message = """
-    Project settings for SublimeLinter have a new, flat format following the
-    Sublime standard. The old-style has been deprecated.
+    Project settings for SublimeLinter have a new, flat format following
+    Sublime Text conventions. The old format has been deprecated, use this instead:
 
     {}
     """
--- a/lint/settings.py
+++ b/lint/settings.py
@@ -126,13 +126,37 @@ def validate_project_settings(filename):
     except ValueError:
         return False
 
-    settings = obj.get('SublimeLinter', {})
-    if len(settings.keys()) >= 2:
-        logger.error(
-            "Invalid settings in '{}':\n"
-            "Only the key 'linters' is allowed. "
-            "Got {}.".format(filename, ', '.join(map(repr, settings.keys())))
-        )
-        return False
+    if 'SublimeLinter' in obj:
+        print_deprecation_message(obj.get('SublimeLinter', {}))
+    else:
+        util.clear_message()
 
-    return validate_settings([(filename, settings)])
+    return True
+
+
+def print_deprecation_message(settings):
+    import json
+
+    message = """
+    Project settings for SublimeLinter have a new, flat format following the
+    Sublime standard. The old-style has been deprecated.
+
+    {}
+    """
+
+    new_settings = {}
+    for linter_name, linter_settings in settings.get('linters', {}).items():
+        for key, value in linter_settings.items():
+            new_settings['.'.join(('SublimeLinter', 'linters', linter_name, key))] = value
+
+    if not new_settings:
+        # User has an empty SublimeLinter obj in their project file. So we
+        # make up an example
+        new_settings['SublimeLinter.linters.eslint.disable'] = True
+
+    formatted_settings = json.dumps(
+        {'settings': new_settings}, sort_keys=True, indent=4
+    )[1:-1]
+    util.show_message(
+        message.format(formatted_settings)
+    )
--- a/lint/settings.py
+++ b/lint/settings.py
@@ -82,7 +82,7 @@ def validate_global_settings():
     return validate_settings(get_settings_objects())
 
 
-def validate_settings(filename_settings_pairs):
+def validate_settings(filename_settings_pairs, flat=False):
     status_msg = "SublimeLinter - Settings invalid!"
     schema_file = "resources/settings-schema.json"
     schema = util.load_json(schema_file, from_sl_dir=True)
@@ -95,11 +95,16 @@ def validate_settings(filename_settings_pairs):
                 validate(settings, schema, format_checker=FormatChecker())
             except ValidationError as error:
                 good = False
-                path_to_err = (' > '.join(
-                    repr(part)
-                    for part in error.path
-                    if not isinstance(part, int)  # drop array indices
-                ) + ': ') if error.path else ''
+                if flat:
+                    path_to_err = '"{}": '.format(
+                        'SublimeLinter.' + '.'.join(error.path)
+                    )
+                else:
+                    path_to_err = (' > '.join(
+                        repr(part)
+                        for part in error.path
+                        if not isinstance(part, int)  # drop array indices
+                    ) + ': ') if error.path else ''
 
                 logger.warning("Invalid settings in '{}'".format(name))
                 util.show_message(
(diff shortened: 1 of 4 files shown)


braver:
Not sure,  but are the view settings already the aggregate of user, project and syntax settings? i.e. if we just read the view settings we’re done and we don’t have to merge them ourselves?

kaste:
Only how SL does the project settings is non-standard.  So we must deprecate or at least strongly discourage

```
{
  "SublimeLinter":
  {
    "linters":
    {
      "flake8":
      {
        "python": 3.6
      }
    }
  },
  "folders":
  [
    {
      "path": "."
    }
  ],
}
```

which is also a lot to write, or to get right because of the deep nesting.  Instead the user should do

```
{
  "folders":
  [
    {
      "path": "."
    }
  ],
  "settings":
  {
    "SublimeLinter.linters.flake8.python": 3.6,
  }
}
```

In Sublime we have static/global settings, you can get them without a view context via `sublime.load_settings("<Package>.sublime-settings")`. Sublime aggreates here all files with the same name.

Or, you have what I call view settings, bc you access them through the view context. Here Sublime aggregates `Preferences > Preferences (Platform) > Syntax > Project Settings > view overrides (persisted in the workspace)`.

For our use-case we need to manually aggregate `defaults (set by the plugin author) > Global Settings (from SublimeLinter.sublime-settings) > View Settings`. For backwards compat we still read our non-standard project settings by evaluating the project file directly. 

However, any view setting will override a (non-standard) project-setting.
So if the user defines `"SublimeLinter.linters.flake8.python": 3.6` in `Preferences` it will mask any non-standard project setting.

braver:
Oh, nice, I didn't know that, at least not in such detail. 

So basically for 4.x we're stuck with our custom project settings thing, but we can already discourage that (at least with this PR) and update our docs to start moving people to the ST standard way of setting things in different contexts?

kaste:
We're not stuck, we just don't remove the compat layer. 

This can get really tricky for users if they don't convert their settings. Say `luae` from last week adds `SublimeLinter.linters.luacheck.selector` to their syntax setting. The idea is probably okay, but it can frustrate users if the old-school project settings don't work anymore. 

The transition/deprecation period can be surprising and frustrating for some users.

But:

- It's #usetheplatform
- The toggler can e.g. disable/enable linters per view
- Users can specify in e.g. their vue syntax settings that eslint should use a different selector

kaste:
To make the transition easier I show a message with the new settings. Feel free to adjust the wording here... 🤞 it should be copy-paste to adjust the project settings.

lint/settings.py
```diff
@@ -126,13 +131,91 @@ def validate_project_settings(filename):
...
+
+    return validate_settings([(filename, deep_settings)], flat=True)
+
+
+def print_deprecation_message(settings):
+    import json
+
+    message = """
+    Project settings for SublimeLinter have a new, flat format following the
+    Sublime standard. The old-style has been deprecated.
```
braver: Suggestion:

"following Sublime Text conventions. The old format has been deprecated, please use this format instead:"

braver:
🛥 it!

kaste:
There is always a bummer. It is not

```
Preferences > Syntax > Project Settings > view overrides (persisted in the workspace)
```
but
```
Preferences > Project Settings > Syntax > view overrides (persisted in the workspace)
```

which means *a plugin* overriding a setting in a Syntax-settings-file is hostile in the sense the user cannot easily overrule anymore. The user would have to overrule in an own Syntax-settings file. (That being said, we did not had syntax specific settings before. )

Just let that sink for a minute. I don't think that's a big problem, it's just usetheplatform, and maybe we're just surprised how the platform works.

zx8:
I may be missing something, but I'm not sure this solves the original problem described in #1446?

I've updated to SublimeLinter 4.6.6 and added the following under `Preferences → Settings – Syntax-specific`: 

```sh
# ~/Library/Application Support/Sublime Text 3/Packages/User/Dockerfile.sublime-settings
{
  "SublimeLinter.linters.shellcheck.args": ["--shell", "sh"]
}
```

Yet I'm still seeing: `Tips depend on target shell and yours is unknown. Add a shebang. [SC2148]`.

Have I done something wrong?

braver:
Works for me, using shellcheck and setting the `--shell` arg. Both in projects as in syntax specific settings. 

Note that when changing syntax specific settings we don't re-lint automatically.

```json
"SublimeLinter.linters.shellcheck.args": "--shell=bash"
```

Changing the setting for a project:

shellcheck-project

zx8:
Can't seem to get it to work & I completely exited & restarted Sublime between every change...

Content of `~/Library/Application Support/Sublime Text 3/Packages/User/Dockerfile.sublime-settings`:
```json
{
  "SublimeLinter.linters.shellcheck.args": "--shell=sh",
}
```

Content of `Dockerfile`
```
FROM scratch

RUN echo foobar
```

Debug output:
```
SublimeLinter: sublime_linter.py:250: Delay buffer 88 for 0.1s
SublimeLinter: sublime_linter.py:250: Delay buffer 88 for 0.0s
SublimeLinter: linter.py:887: Checking lint mode background vs lint reason on_save
SublimeLinter: #24 linter.py:907: 'shellcheck' is linting 'Dockerfile'
SublimeLinter: #24 linter.py:1263: Running ...

  /Users/zx8/Desktop  (working dir)
  $ cat Dockerfile | /usr/local/bin/shellcheck --format=gcc -

SublimeLinter: #24 linter.py:955: shellcheck output:
    -:1:1: error: Tips depend on target shell and yours is unknown. Add a shebang. [SC2148]
SublimeLinter: sublime_linter.py:444: Linting buffer 88 took 0.06s
```

zx8:
I'm using the Dockerfile syntax from this PR: https://github.com/asbjornenge/Docker.tmbundle/pull/24

The statusbar says `Dockerfile`.

braver:
Well you should at least see the args in the debug log.

```
$ cat README.md | /usr/local/bin/shellcheck --format=gcc - --shell=bash
```

Works on my machine:

screen shot 2018-06-25 at 08 56 47


Double check your installed packages, the location of said packages (don't put them in User), etc.

zx8:
> don't put them in User

This was the issue, thanks! Sorry for wasting your time.

braver:
Yeah, that's one of the not-so-documented pitfalls of working with packages manually. Glad we sorted it out 😄

PR #1450, opened 2018-06-13
Various fixes

### LINKED ISSUE #1312: Optimimize what to do when `config_changes` fires

kaste:
Small fixes + optimize config changed actions

Fixes #1312

### DIFF
--- a/lint/settings.py
+++ b/lint/settings.py
@@ -63,11 +63,26 @@ class Settings:
         if not validate_global_settings():
             return
 
-        if self.has_changed('gutter_theme'):
+        gutter_theme_changed = self.has_changed('gutter_theme')
+        if gutter_theme_changed:
             from . import style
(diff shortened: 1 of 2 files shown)


braver:
Nice. Looks good, seems to work good too.

PR #1457, opened 2018-06-18
Ignore orphaned files

### LINKED ISSUE #1454: Proposal: Do not lint orphaned files
If files become orphaned (a network mapping lost, or git checkout different branch) - t.i. they no longer reachable - Sublime shows their views as dirty. At that point, view.file_name() still has the old correct filename, and if the files become reachable again Sublime will remove the dirty flag. 

Other than that, you shouldn't do something to these files. E.g. you shouldn't edit and save unless you really know what you're doing.

I propose to skip linting orphaned files to reduce the risk of total failures (folders or config files are unreachable) or irritating results (e.g. config file falls back to the default or user folder). 

We would need to do this relatively early before the backend, so that these files also loose the `flake8(ok)` marker in status bar. Doing so would clearly indicate that no linter is attached to this view atm.

Alt: we show `flake8(erred)` and console.log why.

Related: https://github.com/SublimeLinter/SublimeLinter-flake8/issues/97

kaste:
Closes #1454

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -2,8 +2,9 @@
 
 from collections import defaultdict, deque
 from contextlib import contextmanager
-from functools import partial
+from functools import lru_cache, partial
 import logging
+import os
 import time
 import threading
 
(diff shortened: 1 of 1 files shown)


sublime_linter.py
```diff
@@ -400,3 +412,20 @@ def remember_runtime(bid):
...
+    try:
+        return fn()
+    except Exception as exc:
+        if 'NoneType' in str(exc):
+            return None
+        else:
+            raise
+
+
+@lru_cache()
```
braver: I'm always scared when I see one of those. I'm scared of caching in general. Is this one safe?

kaste: This is so that we don't spam users with the logging. We show each unique message only once in the status bar. (cache size by default is 128 items/messages). 👍 ?

braver: 👍 👍

kaste:
Ok?

sublime_linter.py
```diff
@@ -312,11 +313,22 @@ def get_linters_for_view(view):
...
-        if linter_class.can_lint_view(view, settings):
-            wanted_linters.append(linter_class(view, settings))
+    filename = view.file_name()
+    # Unassign all linters from orphaned views
+    if filename and not os.path.exists(filename):
+        logger.info(
+            "Skipping buffer {}; '{}' is unreachable".format(bid, filename))
+        flash_once(
+            none_for_none(lambda: view.window().id()),
+            "'{}' has become unreachable.".format(filename)
```
braver: For status_messages skip the quotes around the path and the period around the end.

kaste: Okidoki, you're in detail mode. 😁

braver: Gotta feel like I contribute something 😀

PR #1459, opened 2018-06-23
Use CWD for pipenv check

theFroh:
The incoming `cwd` respects projects with relative paths configured, though if it is None, resort to the existing `get_project_path` guess.

### DIFF
--- a/lint/base_linter/python_linter.py
+++ b/lint/base_linter/python_linter.py
@@ -145,7 +145,7 @@ def get_project_path():
 def ask_pipenv(linter_name, cwd):
     """Ask pipenv for a virtual environment and maybe resolve the linter."""
     # Some pre-checks bc `pipenv` is super slow
-    project_path = get_project_path()
+    project_path = cwd or get_project_path()
     if not project_path:
         return
 


theFroh:
Issue with `/.` is that it'll guarantee a failure of the pipenv check outright:

https://github.com/theFroh/SublimeLinter/blob/6045dd72ef89cf819ba3385d2aec6f8958ab09a6/lint/base_linter/python_linter.py#L152-L154

The actual working directory of Sublime Text is, as far as I can tell, assumed to be arbitrary, so anything involving a relative path is going to suffer from that.

To that end, I don't think there's a sensible default for the case where `cwd` (as determined by the set of heuristics inside of the linter base' `get_working_dir`) is `None`, other than bailing on the pipenv check then and there. That is the current behaviour, and would be the same if `/.` is optionally returned here too.

Now, what might be a better point to consider -- if I have a project structured like:
```
.
├── deep
│   └── test.py
├── Pipfile
├── Pipfile.lock
└── main.py
```

I can be within `/deep` and **not** in a pipenv "shell", yet still run `pipenv --venv` and successfully fetch the correct virtual environment. It appears that pipenv will walk up to find Pipfiles.

Now, consider the case of opening `test.py` in Sublime Text. No project file. If the pipenv check could still run (it wouldn't, both pre- and post- the tweaks made in this PR, as "{CWD}/Pipfile" will not exist -- CWD would validly be `deep/`), then linting would behave completely as expected. How to solve this comes down to how expensive you consider the operation, I suppose!

kaste:
Ah well. Maybe then just none for none: `if cwd is None: return`.

You're of course correct that pipenv searches up, but this is too expensive.

kaste:
If you want to you could fix 

https://github.com/SublimeLinter/SublimeLinter/blob/7dae0eb6c25a7e00a5a14db6b8db0cb3b075796e/lint/base_linter/python_linter.py#L156-L159

As long as `_ask_pipenv` throws if not successful, it will not be cached. This basically means `util.check_output` should not catch any Exception (which is the better choice anyway), and instead the exception will naturally bubble up/unwind and will be catched in `ask_pipenv` and `get_python_version` which are the only callers I think in our code base.

braver:
@theFroh Could you check #1466, see if that works out for you?

kaste:
Sorry, @theFroh, but today was patch-day and I just had to move on. Really thanks for finding and reporting that bug bc that was really old code.

PR #1464, opened 2018-07-13
Iterate demote predicates

kaste:
* Remove 'ws_regions'
* Add 'ws_only', 'some_ws', and 'multilines'

For a short period, we use 'some_ws' when the user set 'ws_regions'
previously.

Note that 'some_ws' will match all errors for which we currently use the
fallback style.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -23,7 +23,9 @@
 
     // Demote the visibility of a class off errors while you're editing.
     // The errors will re-appear after `time_to_idle` and immediately on save.
-    // - ws_regions: erroneous regions that contain whitespace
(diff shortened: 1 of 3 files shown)


braver:
Ehm ok I guess. I always wonder if anyone uses this. I set it to `all` and really like it that way and don't really see the need for all these options. But the implementation seems really easy so 🤷‍♂️

kaste:
The guy who literally pukes all over his laptop when he sees outlines 'mentioned' a possible bug here. So I thought he deserves it.

braver:
Oi, that’s gotta be expensive. I’m totally ok with merging this, if only to help that guy!

kaste:
Serve the people

PR #1465, opened 2018-07-13
Fix: Do not provide default 'types' if user set 'codes'

kaste:
Bascically something like this did not work

image

(All errors were gone. 👍 😁)

After the patch, you can do

image

e.g. to match a specific code *and* all warnings.

### DIFF
--- a/lint/style.py
+++ b/lint/style.py
@@ -45,7 +45,17 @@ def get_value(key, error, default=None):
             except KeyError:
                 ...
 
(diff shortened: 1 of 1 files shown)


kaste:
Every time you post a screenshot, you have a completely different look. And they all look great. 🤔

kaste:
Green comments, wtf?

braver:
Yeah, it’s Apple’s default code style. Weird thing is, it works.

Summertime means light themes!

kaste:
Patch day!

PR #1466, opened 2018-07-13
Fix pipenv support

### LINKED ISSUE #1458: PythonLinter pipenv check assuming cwd
Just finished poking around trying to find why automatic retrieval of the project's current environment via pipenv wasn't happening when trying to set up flake8.

The primary issue appears to be that while the correct project `cwd` (as determined by SublimeLinter's own heuristics) is fetched, this is not used when bailing early in pipenv guessing portion.

https://github.com/SublimeLinter/SublimeLinter/blob/4cb7f74ff181b535f41f5dba20a4b9bfa0d5fe31/lint/base_linter/python_linter.py#L133-L154

`project_path` will be typically relative (unless the user has edited the `folders` block in their settings), meaning the Pipfile will not be found unless Sublime Text was launched, or had its actual CWD changed to, the same directory as the project file.

I think `get_project_path` is redundant here, as this should (correctly) already be retrieved in the `cwd` variable by this point via `self.get_working_dir(settings)`.

kaste:
Fixes #1458 
Closes #1459

### DIFF
--- a/lint/base_linter/python_linter.py
+++ b/lint/base_linter/python_linter.py
@@ -140,21 +140,18 @@ def ask_pipenv(linter_name, cwd):
     if not os.path.exists(pipfile):
         return
 
-    # Defer the real work to another function we can cache.
-    # ATTENTION: If the user has a Pipfile, but did not (yet) installed the
-    # environment, we will cache a wrong result here.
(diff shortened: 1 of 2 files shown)


braver:
I don't know anything about pipenv, but looks good so 🚢  if you're happy with it!

PR #1467, opened 2018-07-13
Use shlex.split() variant for Windows support

### LINKED ISSUE #1442: `shlex.split` is not Windows safe
We use `shlex.split` for `cmd` and `args` if they're provided as strings. This works okay for simple stuff, but in the end it does not split confirming to Windows rules. (E.g. never split a Windows path with it bc it doesn't like `\`s.)

kaste:
Fixes #1442

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -61,10 +61,8 @@
             "disable": false,
 
             // Additional arguments for the command line. Either a 'string'
-            // or an 'array'. If set to a string, we 'shlex.split' it*.
+            // or an 'array'. If set to a string, we split it.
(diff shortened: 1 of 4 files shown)


braver:
So, does this take into account things like wsl, mingw etc.? Other than that I'm not set up to test this.

kaste:
Erm, no. It's about how we split cmd and args given as strings into lists.

So, if I messed this up, we can call the 🚑 bc they 🔫 me.

braver:
Works for me!

kaste:
Now, let's see if we have a working master.

braver:
Looks ok for me

kaste:
✌️

Issue #1475, opened 2018-08-02
Add a "NOTE" severity level


vengin:
Some linters have not only **ERROR**/**WARNING** severity levels, but also **NOTE** message types. So, it might make sense to add a third **NOTE** severity support level, so that it can be scoped and marked up.
Thank you.

braver:
Do you have some examples of this? Ie. what kind of message would be a note?

vengin:
Yes. First of all, I'm using SublimeText for HDL development (i.e. using VHDL, Verilog, SystemVerilog languages)  and most of linters/compiler there have some sort of Note/Info severity level messages. Just as an example the ModelSim might generate a NOTE message like this:
`# ** Note: (vcom-143) Detected '1' FSM/s in design unit 'serial_mul.rtl'.`

braver:
Generally my position on info/note level severity, ie. a severity level below warning is that I haven't seen a good use case for it and therefore don't feel like it's worth it complicate everything so support it.

A linter, especially one integrated into an editor, is to warn the programmer about possible mistakes, risky constructions etc. When a linter reports and *error*, that should mean the programmer has to take action because the code will not work. When the linter reports a *warning* that should mean the programmer should probably at some point take a look at it because there might be a problem.

Info is either:
- a warning in that sense, but the linter is not sure enough that there actually is a problem and so drops the severity level
- the linter just being chatty and telling the programmer about little interesting facts it discovered, but the programmer doesn't have to pay attention to it

Error and warning are a logical distinction, it's either one or the other or it isn't a problem and we shouldn't bother the programmer about it. As soon as you open it up for other levels of severity, "maybe a problem?", "possibly might be something?", you basically open it up for arbitrary levels. Indeed there are linters that have [more than 3 levels](https://pylint.readthedocs.io/en/latest/user_guide/output.html#source-code-analysis-section), so what about those?

I believe it's SublimeLinter's job to translate linter output into clear, contextual and actionable feedback to the end user that is the same for all possible plugins. While I don't believe we should interpret the linter output too much, I do believe severity should come down to either "do something about it now" and "perhaps do something about it later".

One way to deal with this is to categorise the notes as warning, but giving them a style that's less obtrusive.

vengin:
[...]

PR #1478, opened 2018-08-11
separate out error code in panel

braver:
Error codes, especially when they're more texty than your average pythonesque numbered code, kinda get lost in the panel sometimes.

Before:
screen shot 2018-08-11 at 08 52 41

After:
screen shot 2018-08-11 at 08 52 08

### DIFF
--- a/panel/panel.sublime-syntax
+++ b/panel/panel.sublime-syntax
@@ -63,12 +63,12 @@ contexts:
 
   expect-linter-type:
     - include: pop-at-end
-    - match: (\S+)(:)(\S*)
+    - match: ([^\s:]+)(?:(:)(\S+))?
       captures:
         0: meta.linter-code.sublime_linter
-        1: entity.type.linter-name.sublime_linter
+        1: entity.name.class.linter-name.sublime_linter
         2: punctuation.separator.sublime_linter
-        3: storage.type.error-code.sublime_linter
+        3: entity.name.label.error-code.sublime_linter
       pop: true
 
   expect-linter-message:
--- a/panel/panel.sublime-syntax
+++ b/panel/panel.sublime-syntax
@@ -15,12 +15,14 @@ contexts:
     - match: '^\s+(?=[0-9: ]+error)'
       push:
         - ensure-error-meta-scope
+        - expect-linter-message
         - expect-linter-type
         - expect-linter-severity
         - expect-line-maybe-column
     - match: '^\s+(?=[0-9: ]+warning)'
       push:
         - ensure-warning-meta-scope
+        - expect-linter-message
         - expect-linter-type
         - expect-linter-severity
         - expect-line-maybe-column
@@ -52,6 +54,12 @@ contexts:
         1: entity.type.linter-name.sublime_linter
         2: punctuation.separator.sublime_linter
         3: storage.type.error-code.sublime_linter
+      pop: true
+
+  expect-linter-message:
+    - include: pop-at-end
+    - match: .*$
+      scope: meta.linter.message
       pop: true
 
   expect-line-maybe-column:
--- a/panel/panel.sublime-syntax
+++ b/panel/panel.sublime-syntax
@@ -35,8 +35,11 @@ contexts:
 
   expect-linter-type:
     - include: pop-at-end
-    - match: \S+
-      scope: comment.line.sublime_linter
+    - match: (\S+)(:)(\S*)
+      captures:
+        1: comment.line.sublime_linter
+        2: punctuation.separator.sublime_linter
+        3: storage.type.sublime_linter
       pop: true
 
   expect-line-maybe-column:
--- a/panel/panel.sublime-syntax
+++ b/panel/panel.sublime-syntax
@@ -52,13 +52,13 @@ contexts:
   expect-linter-severity:
     - include: pop-at-end
     - match: \berror\b
-      scope: meta.linter.error-type.sublime_linter markup.error.sublime_linter
+      scope: entity.name.label.error-type.sublime_linter markup.error.sublime_linter
       pop: true
     - match: \bwarning\b
-      scope: meta.linter.error-type.sublime_linter markup.warning.sublime_linter
+      scope: entity.name.label.error-type.sublime_linter markup.warning.sublime_linter
       pop: true
     - match: \b(\S+)\b
-      scope: meta.linter.error-type.sublime_linter
+      scope: entity.name.label.error-type.sublime_linter
       pop: true
 
   expect-linter-type:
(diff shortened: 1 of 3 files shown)


kaste:
Subtle. I would have to customize my scheme to like it.

Before: 

image


After:

image


As you can see, I have an almost monochrome panel. (I don't even color highlight the error type.) 

Maybe:

image

With custom colors:

image

image

Again yours:

image

image


I think I would go without the ':'. It looks too close.

kaste:
Looking at it again, I think I prefer the ':'. 🤷‍♂️

kaste:
It is good to go, but the scopes here just suck. `storage.type` is probably good, but `comment.line` for the linter name?

Unfortunately, I don't know enough about all of this, but shouldn't this be 
```
storage.type.linter.linter_name.sublime-linter
storage.type.linter.error_type

```

kaste:
Can you group the `flake8:E302` tokens, so that I can match the ':' as different from the ':' in the 'line:col' token?

Can you give the message a scope?

Can you actually mark lines as errors/warnings so that e.g. warnings can be a bit desaturated?

braver:
@kaste try it with the last commit!

kaste:
That's really interesting. While playing with it, I changed 

```
meta.linter.error.body.sublime_linter
meta.linter.body.error.sublime_linter

(same for warnings)
```

so that I can match `meta.linter.body` to match both.

Also: Shouldn't we use `entity.type.linter_name` -> `entity.type.linter-name`? Currently, we have `line-number` and `col-number`, but `linter_name` and `error_code`.

braver:
> Shouldn't we use entity.type.linter_name -> entity.type.linter-name? Currently, we have line-number and col-number, but linter_name and error_code.

Indeed!

panel/panel.sublime-syntax
```diff
@@ -37,13 +44,21 @@ contexts:
...
+  ensure-warning-other-scope:
+    - meta_scope: meta.linter.body.sublime_linter
+    - match: (?=$)
+      pop: true
+
   expect-linter-severity:
     - include: pop-at-end
     - match: \berror\b
-      scope: markup.error.sublime_linter
+      scope: meta.linter.error-type.sublime_linter markup.error.sublime_linter
```
braver: the meta is actually usually intended to indicate you're matching a group of other, more specific scopes: https://www.sublimetext.com/docs/3/scope_naming.html#meta

> Meta scopes are used to scope larger sections of code or markup, generally containing multiple, more specific scopes.

kaste: Suggestion? 

Usually you would go for `x.linter.problem.error`, `x.linter.problem.warning`, and for other types just `x.linter.problem`. 

I really don't know why we use `entity.type.linter-name` but `storage.type.error-code`. Why do we change the namespace from _entity_ to _storage_?

kaste:
Ah, well, it's git and we can still take whatever we want. ...

- Match the actual error message with `meta.linter.message`
- Ensure compatibility with any error type (related #1475) (via `meta.linter.error-type`)
- Added tests for fun

kaste:
What I don't get to work. error_code is optional, and I would like to omit the ':' if there is no code. Otherwise:

image

which is totally ugly.

panel/panel.sublime-syntax
braver: > Why do we change the namespace from entity to storage?

To differentiate them in a way that will result in a different color in most color schemes. *No* color schemes, now or in the future, will have specified something for `linter-name`. So if they're both just `entity.type` both will look exactly the same. There should be something to differentiate them in a way that makes sense semantically, and there should be something in there that helps existing color schemes assign colors.

braver:
> error_code is optional, and I would like to omit the ':' if there is no code. 

The `:` is baked into [the template](https://github.com/SublimeLinter/SublimeLinter/blob/master/panel_view.py#L351). We'd have to make that more flexible.

kaste:
> The : is baked into the template.

Sure 🙄 I changed that, but I could not get the syntax highlighting right after that.

kaste:
It's in `expect-linter-type:`. I had no luck playing with `- match: (\S+)(:)(\S*)`.

braver:
Oh right. I’ll play with it a bit later/tomorrow

braver:
@kaste to make the report without `:` work I merged the type and message sections because more of the line has to be taken into consideration to make the regex work. I updated the syntax and added a test. Can you try if it works with your dynamic message template?

panel/panel.sublime-syntax
braver: I could go with `entity.name.class` for linter-name, but the default color schemes don't have anything for `entity.name.label` so the code would look just like the name. Which isn't bad, but I like some differentiation there. `entity.name.function` would work though. `entity.name.tag` would work for the error types, but we already have `markup.error/warning` there ensuring that we get signal colors by default.

kaste: That's not what I'm seeing. 

image

Most default schemes actually **don't** have anything for `*.class`, but for `label` and for `tag`. (Basically except 'Celeste')

**None** of the default schemes define `markup.error` or `markup.warning`. Here I wanted the scope to be `entity.name.tag markup.error`, in case someone has that defined.

Now, I think for our purpose class, label, tag can be used interchangeably. So, we have at most 3 different default colors, we can choose from. We can use all three or make it less Kindergarten. E.g. `entity.name.label.error-type`, `entity.name.label.linter-name` and `entity.name.label.error-code`

braver: > None of the default schemes define markup.error or markup.warning

Dang, you're right!? What crap editor is this anyway?

Ok, so I won't really get the control that I want. Let's keep it simple then.

We do have the colorish scopes to get the signal colors for error/warning like we do elsewhere. Let's use the same scopes there as our defaults: `region.yellowish markup.warning.sublime_linter` and `region.redish markup.error.sublime_linter`. I don't believe we need an additional `entity.name` scope there because there are already plenty of hooks to customize it if you want. 

For the others let's use `entity.name.label.linter-name` and `entity.name.label.error-code`. I'd like to have the differentiation, but we don't have to color all the things.

braver: Oh, right, we can't use the colorish scopes because they also do the background color. Dang.

braver: Ok, you've got the overall idea and I'm apparently lacking the sharpness to really add anything of use here. Final thing I'd like to say is that I think the error-type should have a really different scope than the linter:error block so it doesn't all look the same in all the default color schemes. Beyond that I make my own color schemes anyway.

braver:
Alright, @kaste let's ship it?

Issue #1463, opened 2018-07-11
Support for URL's in linter outputs


fatbasstard:
Hi,

I would like to add URL's to linter output message to help people with information about the error/warning. This is a feature that "the" Atom linter supports and works pretty sweet: 

https://steelbrain.me/linter/types/linter-message-v2.html
Example:
![image](https://user-images.githubusercontent.com/35613489/42558170-b92c29aa-84f0-11e8-9112-3e2bdd7f3ce3.png)
 
I'd love to see this feature in SublimeLinter 😄 

If it's already possible, documentation and/or help about it would be great 😉

Sublime itself supports something for this: https://www.sublimetext.com/docs/3/api_reference.html#sublime.Phantom

p.s.
Bonus context: We're working on a CloudFormation linter (https://github.com/awslabs/cfn-python-lint) and every linter rule contains a source url now, explaining where the linter output originates from. Would be great if we can make this information directly available for the users

kaste:
Hi @fatbasstard, 

the blocker here is of course that none of the relevant linters (afaik) provide such an url. (But these urls are anyway probably a simple, static map from error type -> url.) 

Of course, you can just do and prove it! You just need a clear idea of the desired UI/UX to begin with. (E.g. you could add an entry to the right click context menu.)

braver:
Which probably mostly works, until they change something about their website...

Anyway, the tooltips we show are html-ish: https://www.sublimetext.com/docs/3/minihtml.html. It's pretty easy to add a clickable link to them. I guess a linter plugin could set a url attribute using a template string where replace part of that template with the error code. It's probably not even that complicated.

kaste:
This is of course super trivial to implement. #1480 for a strawman, really don't know how this should look like, which is the important thing here. And, again, none of the relevant linters actually output urls to their rules.

kaste:
Via #1480 the UI would look like this.

![image](https://user-images.githubusercontent.com/8558/44166552-eb393400-a0cb-11e8-9c30-e20f5c3f7658.png)

I think this is relatively distracting, but more important it is not accessible. So, we need a command/action for that additionally. I still think that the approach using the (right-click) context menu could work out nicely.

- it hides (useful, additional) functionality you rarely use and keeps the main UI clean
- it is accessible (bc command driven)
- it can be developed as a plugin (on top of the current SL)
- we can use text like "Lookup 'eslint no-unused-vars'", "Show 'no-unused-vars' documentation"

[...]

PR #1480, opened 2018-08-14
WIP: Show a link in the tooltip if url is given

kaste:
Related #1463 

We can afaik only show text links in Sublime. So we have unicode points or ... well text. @braver for the design.

Bc we don't have linters which provide a url right now, we construct a google search for testing.

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -622,7 +622,6 @@ TOOLTIP_STYLES = '''
     }
     a {
         text-decoration: none;
-        font-size: 0.8em;
     }
 '''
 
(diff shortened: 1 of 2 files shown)


braver:
Can we simply make the linter name and error code clickable (and make it look like a link, i.e. `bluish` and underlined?

kaste:
The links are blueish by default or with my color scheme at least.

kaste:
Some pics:

image

image

image

image

image

image

image

braver:
So, assuming linters that provide URL’s also have error codes (because what would you link to otherwise), the first image looks best to me. In the context of a website I would remove the underline, but here it’s needed because ST highlights things in blue that aren’t links and it should be really obvious you can click on them. I personally don’t feel like adding icons or anything.

PR #1483, opened 2018-08-20
Add basic formatting for the message view

### LINKED ISSUE #1340: Apply basic styling to message_view
The fanciest possible styling for our error panel would be possible if we did them in Markdown, but a super simple styling that just styles the header would be IMO enough. This is so that end users see that this is not 'the console' that pops up but a stylized error message.

Our header currently looks like

```
SublimeLinter ....
==================

The content

```

So colorizing the first two lines would be dope I think.

kaste:
Fixes #1340


kaste:
Some pics 

image

image

kaste:
I mean, that's just awesome 🙄

braver:
I have no words, just...

✨ 💃 

🛥

PR #1484, opened 2018-08-21
Syntax dev

kaste:
This is the combination of #1478 and #1483 plus small fixes.

I'll merge this, but keep in mind @braver that this is a "breaking update". Bc of how ST loads syntaxes and its settings, I think users must restart ST. (Otherwise they have an unstyled panel.)


braver:
Yeah, moving the syntaxes probably breaks stuff. Moving the panel.sublime-settings into the syntaxes dir isn't 100% what I'd do either. Besides, the messages are shown in a panel too, so why not keep all that stuff in the "panel" directory?

kaste:
Yes exactly. I revert the rename.

kaste:
This should be okay for PC installs. Installs via git checkout needs a restart bc Sublime will not pick up the settings files. 

I could work around, probably.

braver:
As long as PC users are ok it’s ok with me 🚀

PR #1498, opened 2018-09-30
Remove `util.distinct_until_selection_changed` abstraction

### LINKED ISSUE #1496: Status bar message does not update on secondary views into one buffer
Looks like `def on_selection_modified_async(self, view)` gets always called with the primary view, therefore the `@util.distinct_until_selection_changed` blocks.

kaste:
Fixes #1496

`distinct_until_selection_changed` as a decorator is a mistake bc
different views into the same buffer all have their own selection
(`view.sel()`). So we really have to readout and evaluate
`active_view-sel()` somewhere in these event handler.

Note that the all selection modified event handlers did this already
(for them adding the decorator is double the work) except the one in
`highlight_view.py`.

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -52,6 +52,7 @@ def get_regions_keys(view):
 
 State = {
     'active_view': None,
+    'current_sel': tuple(),
     'idle_views': set(),
     'quiet_views': set()
 }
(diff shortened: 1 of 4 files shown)


braver:
I don't know how to reproduce the bug. If I have 2 views into 1 file edits in either view result in correct updates of both statusbars, even if I don't switch between the views or whatever. This branch also works though, so that's good... ?

kaste:
Editing a file results in calling `on_lint_result` and is just fine. The bug is that the status bar did not update when moving the cursor (aka changing the selection, view.sel()). 

E.g. open two views into the same buffer with some errors in it. Then jump around using `ctrl-k, p`. In the primary view, the status bar tells you on which error you sit. The secondary view maybe updates once.

braver:
Hm I guess I don't get it? A change of selection in one view doesn't do anything for the other view? Perhaps a screen capture to help me out? Or just ignore me and merge it 🤷‍♂️ 😉

braver:
@kaste lets just 🛳 this?

kaste:
Yeah sure.

PR #1499, opened 2018-09-30
Small logging improvements

kaste:
- The log messages are a bit better aligned
- Log the filename where we just logged the buffer id


braver:
LGTM, ship ship!

Issue #1418, opened 2018-05-19
Improve message when executable cannot be found


kaste:
Currently we output something like `SublimeLinter: #2 linter.py:573: WARNING: lua cannot locate 'luac'` if the executable cannot be found using a global which. 

```
SublimeLinter: WARNING: (lua) Linter non-functional: Cannot find a binary 'luac' in PATH.
```

Maybe we should then print PATH as well. Or ask "Did you install 'luac' globally?"

As reported here https://github.com/SublimeLinter/SublimeLinter-lua/issues/7

braver:
Ideally an error message contains a hint of how to solve it. So, if we’re sure globally installing something is the right solution we can suggest that.

Printing PATH feels like debugging to me. I would prefer something like “if this problem persists, please refer to our troubleshooting guide: (url)” and the readme of this plugin. 

Perhaps we should do that anyway, so the plugin readme can tell you what the right approach to installation etc is.

ew4321:
I hope the changes I made are satisfactory.

braver:
closed via #1502

PR #1501, opened 2018-10-01
improved error message shown to user.

ew4321:
The error message now gives a hint on how to solve the issue by directing them to the troubleshooting guide in SublimeLinter's docs and to the readme. #1418

### DIFF
--- a/lint/base_linter/composer_linter.py
+++ b/lint/base_linter/composer_linter.py
@@ -55,7 +55,9 @@ class ComposerLinter(linter.Linter):
         if global_cmd:
             return True, global_cmd
         else:
-            logger.warning("{} cannot locate '{}'".format(self.name, cmd[0]))
+            logger.warning('{} cannot locate \'{}\'\n'
+                           'Please refer to the readme of this plugin and our troubleshooting guide: '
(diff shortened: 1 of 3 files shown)


braver:
Thanks! Two things:

1. It needs a line break in there, because it comes out like so:
```
SublimeLinter: WARNING: markdownlint cannot locate 'markdownlint'Please refer to the readme of this plugin and our troubleshooting guide: http://www.sublimelinter.com/en/stable/troubleshooting.html```

2. The composer_linter and node_linter have similar messages and need an update there too. The file you modified is just the base linter, most plugins use a more specific version.

ew4321:
Hi. I'm new to this sort of stuff, and I messed something up so I made the changes in a [new branch](https://github.com/SublimeLinter/SublimeLinter/pull/1502).

p.s. How can I simulate an error msg like you did above, so i can find my own mistakes next time?

braver:
What I did was install a plugin of which I don't have the actual linter installed (like markdownlint) and then open a file it can lint (like a README).

kaste:
Thank you.

Interesting duplication in composer and node_linter. We probably should just `return False, None` before trying to find the global cmd. Finding a global cmd is the default behavior of SublimeLinter anyway.

PR #1502, opened 2018-10-01
Error msg hotfix#1418

ew4321:
Improved error msg that directs user to readme and troubleshooting guide when executable cannot be found for base linter, node linter, and composer linter. #1418


braver:
Looks good, thanks! Also for me good enough to close #1418 with this, if someone has more fancy ideas that's cool but I don't see any real reason to do it more complicated than this.

ew4321:
Thanks!

PR #1509, opened 2018-10-18
[TEST] Patch user environment

kaste:
This is not to merge!

INSTRUCTIONS
------------

You can either checkout this PR or maybe even easier: just grab the file https://github.com/SublimeLinter/SublimeLinter/blob/fix-environment/_fix_user_env.py 

Copy paste the raw file, and put it in your Sublime User Folder. File name doesn't matter. It should run immediately, but at most once per restart. 

Now, open the Sublime Console and see what it prints. Look for something like this

```
<> Running: '/bin/bash -ilc env'
<> Finished in 0.7s
<> Environment already ok!
```

If the environment 'is not already ok' it will print which keys are added/changed. Closely look at these! If you share them in a comment in an issue: **DO NOT POST SECRET KEYS!**

TEST: If possible test this starting Sublime from different working dirs. E.g. will it select different rubies or node versions depending on the last working dir you started Sublime from. (E.g. a different `GEM_HOME` or `PATH`.)

TEST: Will it print different things if you start Sublime from the command line vs. from a graphical thing (like Dock)?

TEST: In case it modified the environment and you modified `paths` or `env` through the SublimeLinter settings. Undo these settings and check if the linter will now work automatically t.i. without manual configuration. 

Also NOTE if this process takes a rather long time. More than _one_ second already counts as slow here.

RATIONALE
---------

Users see wrong execution environments even after Sublime 3.1 which ships https://github.com/SublimeTextIssues/Core/issues/1877. (Typically they see an error message with 'No such file or directory'.) This implementation runs after `plugin_loaded` so it doesn't race with Sublime's readiness. Ideally and expected, we only ever see 'Environment already ok!'. 

The implementation here follows Atom and VSCode.

Since I don't have a Mac I had to code this blind without testing, 🤞. It is a good starting point but ...  **at some point one of the people that do use Macs needs to care enough about this issue to actively participate here.**

### DIFF
--- a/_fix_user_env.py
+++ b/_fix_user_env.py
@@ -1,0 +1,148 @@
+from contextlib import contextmanager
+from functools import partial, wraps
+import os
+import re
+import subprocess
+import time
+import threading
+import traceback
+
+import sublime
+
+
+DEFAULT_SHELL = '/bin/bash'
+DELIMITER = '_SHELL_ENV_DELIMITER_'
+ANSI_COLOR_RE = re.compile(r'\033\[[0-9;]*m')
+
+State = {
+    'ready': False,
+    'callbacks': []
+}
+
+
+def synchronized(lock):
+    def decorator(fn):
+        @wraps(fn)
+        def wrapper(*a, **kw):
+            with lock:
+                return fn(*a, **kw)
+
+        return wrapper
+    return decorator
+
+
+atomic = synchronized(threading.RLock())
+
+
+def eat_errors_and_resolve(fn):
+    @wraps(fn)
+    def wrapper(sink):
+        try:
+            return fn(sink)
+        except Exception:
+            traceback.print_exc()
+            sink({})
+            return
+
+    return wrapper
+
+
+def strip_ansi(str):
+    return ANSI_COLOR_RE.sub('', str)
+
+
+def plugin_loaded():
+    read_and_update_env_async()
+
+
+def read_and_update_env_async():
+    threading.Thread(target=partial(read_env, update_env)).start()
+
+
+@eat_errors_and_resolve
+def read_env(sink):
+    if sublime.platform() == 'windows':
+        sink({})
+        return
+
+    executable = os.environ.get('SHELL') or DEFAULT_SHELL
+    cmd = [
+        executable,
+        '-ilc',
+        'echo -n "{0}"; env; echo -n "{0}"; exit'.format(DELIMITER)
+    ]
+
+    print("<> Running: '{} -ilc env'".format(executable))
+    with print_runtime('<> Finished in {:.2f}s'):
+        stdout = subprocess.check_output(cmd)
+
+    output = stdout.decode('utf8').replace('\r\n', '\n').replace('\r', '\n')
+    output = strip_ansi(output)
+
+    user_env = {}
+    for line in output.split(DELIMITER)[1].split('\n'):
+        try:
+            key, value = line.split('=', maxsplit=1)
+        except ValueError:
+            continue
+
+        user_env[key] = value
+
+    sink(user_env)
+
+
+@atomic
+def update_env(user_env):
+    print_dict_changes(user_env, os.environ)
+    os.environ.update(user_env)
+
+    for callback in State['callbacks']:
+        safe_call(callback)
+
+    State.update({'ready': True, 'callbacks': []})
+
+
+@atomic
+def on_ready(fn):
+    if State['ready']:
+        safe_call(fn)
+    else:
+        State['callbacks'].append(fn)
+
+
+def safe_call(fn):
+    sublime.set_timeout_async(fn)
+
+
+def print_dict_changes(current, previous):
+    common_keys = current.keys() & previous.keys()
+
+    added = current.keys() - common_keys
+    changed = [k for k in common_keys if current[k] != previous[k]]
+
+    if added or changed:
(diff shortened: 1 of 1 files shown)


kaste:
Implementation notes

- This uses the same technique found in Atom and VSCode
- I run the shell command after `plugin_loaded` callback, so there is no race between Sublime's own readiness and this one. (Ideally, we never expect anything else than `Environment already ok!` in Sublime > 3.1.)

cgat:
This seems to be working for me. Mac OS (High Sierra) + fish shell + nvm 

Here's the output from the console:
```
reloading plugin User.fix_env
<> Running: '/usr/local/bin/fish -ilc env'
<> Finished in 0.82s
<> Environment will change!
-> Added:
NVM_CD_FLAGS=
OMF_CONFIG=/Users/cgat/.config/omf
rvm_prefix=/Users/cgat
LC_CTYPE=en_US.UTF-8
GOPATH=/Users/cgat/go
RUBY_VERSION=ruby-2.5.1
rvm_path=/Users/cgat/.rvm
OMF_PATH=/Users/cgat/.local/share/omf
SHLVL=1
GEM_PATH=/Users/cgat/.rvm/gems/ruby-2.5.1:/Users/cgat/.rvm/gems/ruby-2.5.1@global
GEM_HOME=/Users/cgat/.rvm/gems/ruby-2.5.1
NVM_DIR=/Users/cgat/.nvm
rvm_bin_path=/Users/cgat/.rvm/bin
rvm_version=1.29.4latest
NVM_BIN=/Users/cgat/.nvm/versions/node/v9.11.2/bin
PWD=/Applications/Sublime Text.app/Contents/MacOS
-> Changed:
PATH=/usr/local/opt/qt@5.5/bin:/Users/cgat/go/bin:/Users/cgat/.rvm/gems/ruby-2.5.1/bin:/Users/cgat/.rvm/gems/ruby-2.5.1@global/bin:/Users/cgat/.rvm/rubies/ruby-2.5.1/bin:/Users/cgat/.nvm/versions/node/v9.11.2/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/go/bin:/Users/cgat/.rvm/bin
    |/usr/bin:/bin:/usr/sbin:/sbin (previous value)
XPC_SERVICE_NAME=0
```

Thanks @kaste

jon-cooke-rentalcars:
Works for me. Thanks :)

braver:
Does this affect the environment of other plugins? We’re not the only plugin with env problems, so perhaps we can solve this in a way that can support the eco system?

braver:
> a) Should/Must we call this per window/project dir or is one call enough? (AFAIK calls this per window, VSCode once.) Which working_dir should we use if we just call this once?

IIRC, Atom windows are separate instances. ST is just one instance, one shared pool as it were. You can attach things to windows, but you don't have to. If ST is already running, which on a Mac it usually is because apps don't exit after the last window closes, "starting" it from a different working dir probably doesn't have the intended result. I'm not sure what it does in that case, what happens with the env?

kaste:
I just thought that these sophisticated scripts like *rvm* or *nvm* etc. change the env *vastly* different depending on which working dir you're in. But maybe they just 'install' them-self. 

Now, as far as I understand these things, after we patched the env they should choose e.g. different `node`s or `ruby`s *automatically* depending on the working/project dir. If they do, patching globally once is probably okay. If they always choose the same ruby/node binary, we probably need to run this 'per window'.

alevine26:
Hello, can you clarify what `put it in your Sublime User Folder` means? Putting this as a python file in `~/Library/Application Support/Sublime Text 3/` does not seem to run the script on startup, my console output remains the same.

matthewmorek:
It looks like there's a race condition, as SublimeLinter runs *before* the updated script can update the `env`. SublimeLinter still doesn't recognise the path. Take a look at my log:

```
plugins loaded
<> Running: '/usr/local/bin/zsh -ilc env'
SublimeLinter: ERROR: env: node: No such file or directory

reloading settings Packages/User/ApplySyntax.ext-list
reloading settings Packages/User/Bash.sublime-settings
Emmet: No need to update PyV8
<> Finished in 0.93s
<> Environment will change!
-> Added:
EDITOR=nano
SSH_KEY_PATH=~/.ssh/rsa_id
LC_CTYPE=
SHLVL=1
PWD=/Applications/Sublime Text.app/Contents/MacOS
ZSH=/Users/matthewmorek/.oh-my-zsh
_=/usr/bin/env
PAGER=less
LESS=-R
LSCOLORS=Gxfxcxdxbxegedabagacad
-> Changed:
PATH=/usr/local/opt/node@10/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
    |/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin (previous value)
```

My `env`:
- Node@10 from Homebrew, available from `/usr/local/opt/node@10/bin`
- Yarn from Homebrew, installed using `--ignore-dependencies`, available from `/usr/local/bin`
- Eslint installed via `yarn global add eslint` and available from `/usr/local/bin`

kaste:
@matthewmorek Yes sure. The code of the patch actually would support a non-racing `on_ready(callback)` mode which would allow SublimeLinter to wait for readiness, but here we just test this method.

kaste:
Closing bc there is just no momentum which makes merging very unlikely.

PR #1513, opened 2018-10-25
Fix lint errors

kaste:
`flake8` updated. We ignore W504 for now, new pep recommendation is to put the binary operators (and/or) on the new line. E.g.:

```
    maybe_toggle_panel_automatically = (
        persist.settings.get('lint_mode') == 'manual' 
		or buffer_id in State['just_saved_buffers']  # <======
    )
```


braver:
👍

Issue #1379, opened 2018-05-04
Nice error message if the `VirtualView` throws


kaste:
If the linter reports errors referring lines not in the `view`, we get a traceback here

```
Linter crashed.

Traceback (most recent call last):
  File "C:\Users\marce\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\backend.py", line 87, in execute_lint_task
    errors = linter.lint(code, view_has_changed, settings) or []
  File "C:\Users\marce\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\linter.py", line 726, in lint
    return self.parse_output(output, virtual_view)
  File "C:\Users\marce\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\linter.py", line 747, in parse_output
    return self.parse_output_via_regex(output, virtual_view)
  File "C:\Users\marce\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\linter.py", line 767, in parse_output_via_regex
    error = self.process_match(m, virtual_view)
  File "C:\Users\marce\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\linter.py", line 848, in process_match
    start, end = vv.full_line(m.line)
  File "C:\Users\marce\AppData\Roaming\Sublime Text 3\Packages\SublimeLinter\lint\linter.py", line 83, in full_line
    start = self._newlines[line]
IndexError: list index out of range
```

Usually, this means the linter reports errors from other files as well. (Or, the regex is broken.)

We should show a nicer error message here.

mheinzler:
I'm having the same problem with clang-tidy. There are a lot of warnings that can appear in header files which sometimes results in this error when the header file has more lines than the linted file.

A workaround for me is to change the mentioned line to the following. At least then the warnings are still shown.
```py
start = self._newlines[min(line, len(self._newlines) - 1)]
```

Of course, ultimately it would be great if SublimeLinter was able to handle warnings/errors from multiple files, but I have no idea of where to start implementing something like this.

kaste:
Shouldn't the linter `clang-tidy` filter messages coming from other files? How can the reported 'line:col' pairs be of any use if applied to the wrong buffer. 

Be forgiving here is usually correct. However, I always thought we need to inform the plugin author that there is something fishy going on here. 

I'll take and would appreciate a PR with your line, but keep in mind that writing the unit tests for such a line is the actual effort.

[...]

PR #1528, opened 2018-12-12
WIP: Support linter messages from multiple files

mheinzler:
This enables support to show messages from multiple files correctly. For example [clang-tidy](https://github.com/rwols/SublimeLinter-contrib-clang-tidy) can be configured to also show messages from files that are included in the linted file. Right now these messages are treated as coming from the linted file which is confusing and can lead to errors like #1379.

These changes try to format such messages nicely and make them behave just like messages from the linted file. It is already possible to jump to the mentioned line. Currently it looks like this:
grafik

There are some issues I have found so far:
- [ ] Columns in other files are always shown as `1`
- [ ] The panel is not updated when navigating other files
- [ ] Tests are missing

Some parts of the code are still a bit ugly. For example to get the filename a message refers to it accesses the regex match directly because I couldn't find a practical way to extend the `LintMatch` tuple.

It would be great if someone else would be also able to test this or even provide some improvements.

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -58,13 +58,13 @@ def get_lint_tasks(linters, view, view_has_changed):
         tasks = []
         for region in regions:
             code = view.substr(region)
-            offset = view.rowcol(region.begin())
+            offsets = view.rowcol(region.begin()) + (region.begin(),)
 
             # Due to a limitation in python 3.3, we cannot 'name' a thread when
             # using the ThreadPoolExecutor. (This feature has been introduced
             # in python 3.6.) So, we do this manually.
             task_name = make_good_task_name(linter, view)
-            task = partial(execute_lint_task, linter, code, offset, view_has_changed)
+            task = partial(execute_lint_task, linter, code, offsets, view_has_changed)
             executor = partial(modify_thread_name, task_name, task)
             tasks.append(executor)
 
@@ -106,10 +106,10 @@ def reduced_concurrency():
 
 
 @reduced_concurrency()
-def execute_lint_task(linter, code, offset, view_has_changed):
+def execute_lint_task(linter, code, offsets, view_has_changed):
     try:
         errors = linter.lint(code, view_has_changed) or []
-        finalize_errors(linter, errors, offset)
+        finalize_errors(linter, errors, offsets)
 
         return errors
     except linter_module.TransientError:
@@ -129,10 +129,10 @@ def error_json_serializer(o):
     return o
 
 
-def finalize_errors(linter, errors, offset):
+def finalize_errors(linter, errors, offsets):
     linter_name = linter.name
     view = linter.view
-    line_offset, col_offset = offset
+    line_offset, col_offset, pt_offset = offsets
 
     for error in errors:
         line, start, end = error['line'], error['start'], error['end']
@@ -142,11 +142,23 @@ def finalize_errors(linter, errors, offset):
 
         line += line_offset
 
+        try:
+            region = error['region']
+        except KeyError:
+            line_start = view.text_point(line, 0)
+            region = sublime.Region(line_start + start, line_start + end)
+            if len(region) == 0:
+                region.b = region.b + 1
+
+        else:
+            region = sublime.Region(region.a + pt_offset, region.b + pt_offset)
+
         error.update({
             'line': line,
             'start': start,
             'end': end,
-            'linter': linter_name
+            'linter': linter_name,
+            'region': region
         })
 
         uid = hashlib.sha256(
@@ -156,16 +168,6 @@ def finalize_errors(linter, errors, offset):
             'uid': uid,
             'priority': style.get_value('priority', error, 0)
         })
-
-        if 'region' not in error:
-            line_start = view.text_point(line, 0)
-            region = sublime.Region(line_start + start, line_start + end)
-            if len(region) == 0:
-                region.b = region.b + 1
-
-            error.update({
-                'region': region
-            })
 
 
 def get_lint_regions(linters, view):
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -121,6 +121,14 @@ def execute_lint_task(linter, code, offset, view_has_changed):
         return []  # Empty list here to clear old errors
 
 
+def error_json_serializer(o):
+    """Return a JSON serializable representation of error properties."""
+    if isinstance(o, sublime.Region):
+        return (o.a, o.b)
+
+    return o
+
+
 def finalize_errors(linter, errors, offset):
     linter_name = linter.name
     view = linter.view
@@ -142,18 +150,22 @@ def finalize_errors(linter, errors, offset):
         })
 
         uid = hashlib.sha256(
-            json.dumps(error, sort_keys=True).encode('utf-8')).hexdigest()
-
-        line_start = view.text_point(line, 0)
-        region = sublime.Region(line_start + start, line_start + end)
-        if len(region) == 0:
-            region.b = region.b + 1
+            json.dumps(error, sort_keys=True, default=error_json_serializer).encode('utf-8')).hexdigest()
 
         error.update({
             'uid': uid,
-            'region': region,
             'priority': style.get_value('priority', error, 0)
         })
+
+        if 'region' not in error:
+            line_start = view.text_point(line, 0)
+            region = sublime.Region(line_start + start, line_start + end)
+            if len(region) == 0:
+                region.b = region.b + 1
+
+            error.update({
+                'region': region
+            })
 
 
 def get_lint_regions(linters, view):
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -143,11 +143,11 @@ def finalize_errors(linter, errors, offsets):
                 belongs_to_main_file = False
 
         line, start, end = error['line'], error['start'], error['end']
-        if line == 0:
-            start += col_offset
-            end += col_offset
-
         if belongs_to_main_file:  # offsets are for the main file only
+            if line == 0:
+                start += col_offset
+                end += col_offset
+
             line += line_offset
 
         try:
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -135,12 +135,20 @@ def finalize_errors(linter, errors, offsets):
     line_offset, col_offset, pt_offset = offsets
 
     for error in errors:
+        # see if this error belongs to the main file
+        belongs_to_main_file = True
+        if 'filename' in error:
+            if (os.path.normcase(error['filename']) != os.path.normcase(view.file_name() or '') and
+                    error['filename'] != "<untitled {}>".format(view.buffer_id())):
+                belongs_to_main_file = False
+
         line, start, end = error['line'], error['start'], error['end']
         if line == 0:
             start += col_offset
             end += col_offset
 
-        line += line_offset
+        if belongs_to_main_file:  # offsets are for the main file only
+            line += line_offset
 
         try:
             region = error['region']
@@ -151,7 +159,8 @@ def finalize_errors(linter, errors, offsets):
                 region.b = region.b + 1
 
         else:
-            region = sublime.Region(region.a + pt_offset, region.b + pt_offset)
+            if belongs_to_main_file:  # offsets are for the main file only
+                region = sublime.Region(region.a + pt_offset, region.b + pt_offset)
 
         error.update({
             'line': line,
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -104,6 +104,12 @@ class VirtualView:
     # def text_point(self, row, col) => Point
     # def rowcol(self, point) => (row, col)
 
+    @staticmethod
+    def from_file(filename):
+        """Return a VirtualView with the contents of file."""
+        with open(filename, 'r') as f:
+            return VirtualView(f.read())
+
 
 class ViewSettings:
     """
@@ -1066,6 +1072,15 @@ class Linter(metaclass=LinterMeta):
     def process_match(self, m, vv):
         error_type = self.get_error_type(m.error, m.warning)
 
+        # discard a linter provided filename if it points to the linted file itself
+        filename = m.match.groupdict().get("filename", None)
+        if filename and self.filename and os.path.samefile(filename, self.filename):
+            filename = None
+
+        # if this is a match for a different file we need its contents for the below checks
+        if filename:
+            vv = VirtualView.from_file(filename)
+
         col = m.col
         line = m.line
 
@@ -1088,6 +1103,7 @@ class Linter(metaclass=LinterMeta):
 
         line, start, end = self.reposition_match(line, col, m, vv)
         return {
+            "filename": filename,
             "line": line,
             "start": start,
             "end": end,
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1103,11 +1103,19 @@ class Linter(metaclass=LinterMeta):
             col = max(min(col, (end - start) - 1), 0)
 
         line, start, end = self.reposition_match(line, col, m, vv)
+
+        # find the region to highlight for this error
+        line_start, _ = vv.full_line(line)
+        region = sublime.Region(line_start + start, line_start + end)
+        if len(region) == 0:
+            region.b += 1
+
         return {
             "filename": filename,
             "line": line,
             "start": start,
             "end": end,
+            "region": region,
             "error_type": error_type,
             "code": m.error or m.warning or '',
             "msg": m.message.strip(),
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1073,7 +1073,11 @@ class Linter(metaclass=LinterMeta):
         error_type = self.get_error_type(m.error, m.warning)
 
         # determine a filename for this match
-        filename = m.match.groupdict().get("filename", None)
+        try:
+            filename = m.match.groupdict().get("filename", None)
+        except AttributeError:
+            pass
+
         if filename:
             # ensure that the filename is absolute by basing relative paths on
             # the working directory
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -613,6 +613,8 @@ class Linter(metaclass=LinterMeta):
         if self.defaults is not None:
             self.defaults = self.defaults.copy()
 
+        self.temp_filename = None
+
     @property
     def filename(self):
         """Return the view's file path or '' if unsaved."""
@@ -1120,14 +1122,19 @@ class Linter(metaclass=LinterMeta):
         # determine a filename for this match
         filename = m.filename
 
+        main_filename = self.view.file_name() or "<untitled {}>".format(self.view.buffer_id())
         if filename and not self.is_stdin_filename(filename):
             # ensure that the filename is absolute by basing relative paths on
             # the working directory
             filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
 
+            # when the command was run on a temporary file we need to compare
+            # this filename with that temporary filename
+            cmd_filename = self.temp_filename or self.filename
+
             # if this is a match for a different file we need its contents for
             # the below checks
-            if os.path.normcase(filename) != os.path.normcase(self.filename):
+            if os.path.normcase(filename) != os.path.normcase(cmd_filename):
                 try:
                     vv = VirtualView.from_file(filename)
                 except OSError as err:
@@ -1135,9 +1142,12 @@ class Linter(metaclass=LinterMeta):
                     logger.warning('Exception: {}'.format(str(err)))
                     self.notify_failure()
                     return None
+            else:
+                # replace the temporary filename
+                filename = main_filename
         else:
             # use the filename of the current view
-            filename = self.view.file_name() or "<untitled {}>".format(self.view.buffer_id())
+            filename = main_filename
 
         col = m.col
         line = m.line
@@ -1310,6 +1320,9 @@ class Linter(metaclass=LinterMeta):
             suffix = self.get_tempfile_suffix()
 
         with make_temp_file(suffix, code) as file:
+            # store this filename to assign its errors to the main file later
+            self.temp_filename = file.name
+
             ctx = get_view_context(self.view)
             ctx['file_on_disk'] = self.filename
             ctx['temp_file'] = file.name
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -150,7 +150,7 @@ class VirtualView:
     @staticmethod
     def from_file(filename):
         """Return a VirtualView with the contents of file."""
-        with open(filename, 'r') as f:
+        with open(filename, 'r', encoding='utf8') as f:
             return VirtualView(f.read())
 
 
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1075,6 +1075,10 @@ class Linter(metaclass=LinterMeta):
         # determine a filename for this match
         filename = m.match.groupdict().get("filename", None)
         if filename:
+            # ensure that the filename is absolute by basing relative paths on
+            # the working directory
+            filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
+
             # if this is a match for a different file we need its contents for the below checks
             if not self.filename or not os.path.samefile(filename, self.filename):
                 vv = VirtualView.from_file(filename)
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1076,7 +1076,7 @@ class Linter(metaclass=LinterMeta):
         try:
             filename = m.match.groupdict().get("filename", None)
         except AttributeError:
-            pass
+            filename = None
 
         if filename:
             # ensure that the filename is absolute by basing relative paths on
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1118,10 +1118,7 @@ class Linter(metaclass=LinterMeta):
         code = m.code or m.error or m.warning or ''
 
         # determine a filename for this match
-        try:
-            filename = m.match.groupdict().get("filename", None)
-        except AttributeError:
-            filename = None
+        filename = m.filename
 
         if filename:
             # ensure that the filename is absolute by basing relative paths on
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1120,34 +1120,21 @@ class Linter(metaclass=LinterMeta):
         code = m.code or m.error or m.warning or ''
 
         # determine a filename for this match
-        filename = m.filename
+        filename = self.normalize_filename(m.filename)
 
-        main_filename = self.view.file_name() or "<untitled {}>".format(self.view.buffer_id())
-        if filename and not self.is_stdin_filename(filename):
-            # ensure that the filename is absolute by basing relative paths on
-            # the working directory
-            filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
-
-            # when the command was run on a temporary file we need to compare
-            # this filename with that temporary filename
-            cmd_filename = self.temp_filename or self.filename
-
-            # if this is a match for a different file we need its contents for
+        if filename:
+            # this is a match for a different file so we need its contents for
             # the below checks
-            if os.path.normcase(filename) != os.path.normcase(cmd_filename):
-                try:
-                    vv = VirtualView.from_file(filename)
-                except OSError as err:
-                    # warn about the error and drop this match
-                    logger.warning('Exception: {}'.format(str(err)))
-                    self.notify_failure()
-                    return None
-            else:
-                # replace the temporary filename
-                filename = main_filename
-        else:
+            try:
+                vv = VirtualView.from_file(filename)
+            except OSError as err:
+                # warn about the error and drop this match
+                logger.warning('Exception: {}'.format(str(err)))
+                self.notify_failure()
+                return None
+        else:  # main file
             # use the filename of the current view
-            filename = main_filename
+            filename = self.view.file_name() or "<untitled {}>".format(self.view.buffer_id())
 
         col = m.col
         line = m.line
@@ -1195,6 +1182,24 @@ class Linter(metaclass=LinterMeta):
             return WARNING
         else:
             return self.default_type
+
+    def normalize_filename(self, filename):
+        """Return an absolute filename if it is not the main file."""
+        if filename and not self.is_stdin_filename(filename):
+            # ensure that the filename is absolute by basing relative paths on
+            # the working directory
+            filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
+
+            # when the command was run on a temporary file we need to compare
+            # this filename with that temporary filename
+            cmd_filename = self.temp_filename or self.filename
+
+            # only return a filename if it is a different file
+            if os.path.normcase(filename) != os.path.normcase(cmd_filename):
+                return filename
+
+        # must be the main file
+        return None
 
     @staticmethod
     def is_stdin_filename(filename):
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1084,7 +1084,7 @@ class Linter(metaclass=LinterMeta):
                 # for the below checks
                 if not self.filename or not os.path.samefile(filename, self.filename):
                     vv = VirtualView.from_file(filename)
-            except (FileNotFoundError, OSError) as err:
+            except OSError as err:
                 # warn about the error and drop this match
                 logger.warning('Exception: {}'.format(str(err)))
                 self.notify_failure()
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1190,13 +1190,17 @@ class Linter(metaclass=LinterMeta):
             # the working directory
             filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
 
-            # when the command was run on a temporary file we need to compare
-            # this filename with that temporary filename
-            cmd_filename = self.temp_filename or self.filename
-
             # only return a filename if it is a different file
-            if os.path.normcase(filename) != os.path.normcase(cmd_filename):
-                return filename
+            normed_filename = os.path.normcase(filename)
+            if normed_filename == os.path.normcase(self.filename):
+                return None
+
+            # when the command was run on a temporary file we also need to
+            # compare this filename with that temporary filename
+            if self.temp_filename and normed_filename == os.path.normcase(self.temp_filename):
+                return None
+
+            return filename
 
         # must be the main file
         return None
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1072,14 +1072,15 @@ class Linter(metaclass=LinterMeta):
     def process_match(self, m, vv):
         error_type = self.get_error_type(m.error, m.warning)
 
-        # discard a linter provided filename if it points to the linted file itself
+        # determine a filename for this match
         filename = m.match.groupdict().get("filename", None)
-        if filename and self.filename and os.path.samefile(filename, self.filename):
-            filename = None
-
-        # if this is a match for a different file we need its contents for the below checks
         if filename:
-            vv = VirtualView.from_file(filename)
+            # if this is a match for a different file we need its contents for the below checks
+            if not self.filename or not os.path.samefile(filename, self.filename):
+                vv = VirtualView.from_file(filename)
+        else:
+            # use the filename of the current view
+            filename = self.view.file_name() or "<untitled {}>".format(self.view.buffer_id())
 
         col = m.col
         line = m.line
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1079,16 +1079,16 @@ class Linter(metaclass=LinterMeta):
             # the working directory
             filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
 
-            try:
-                # if this is a match for a different file we need its contents
-                # for the below checks
-                if not self.filename or not os.path.samefile(filename, self.filename):
+            # if this is a match for a different file we need its contents for
+            # the below checks
+            if os.path.normcase(filename) != os.path.normcase(self.filename):
+                try:
                     vv = VirtualView.from_file(filename)
-            except OSError as err:
-                # warn about the error and drop this match
-                logger.warning('Exception: {}'.format(str(err)))
-                self.notify_failure()
-                return None
+                except OSError as err:
+                    # warn about the error and drop this match
+                    logger.warning('Exception: {}'.format(str(err)))
+                    self.notify_failure()
+                    return None
         else:
             # use the filename of the current view
             filename = self.view.file_name() or "<untitled {}>".format(self.view.buffer_id())
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1120,7 +1120,7 @@ class Linter(metaclass=LinterMeta):
         # determine a filename for this match
         filename = m.filename
 
-        if filename:
+        if filename and not self.is_stdin_filename(filename):
             # ensure that the filename is absolute by basing relative paths on
             # the working directory
             filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
@@ -1185,6 +1185,10 @@ class Linter(metaclass=LinterMeta):
             return WARNING
         else:
             return self.default_type
+
+    @staticmethod
+    def is_stdin_filename(filename):
+        return filename in ["stdin", "<stdin>", "-"]
 
     def maybe_fix_tab_width(self, line, col, vv):
         # Adjust column numbers to match the linter's tabs if necessary
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1188,7 +1188,8 @@ class Linter(metaclass=LinterMeta):
         if filename and not self.is_stdin_filename(filename):
             # ensure that the filename is absolute by basing relative paths on
             # the working directory
-            filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
+            cwd = self.get_working_dir(self.settings) or os.path.realpath('.')
+            filename = os.path.normpath(os.path.join(cwd, filename))
 
             # only return a filename if it is a different file
             normed_filename = os.path.normcase(filename)
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -955,7 +955,7 @@ class Linter(metaclass=LinterMeta):
         return [
             error
             for error in errors
-            if not any(
+            if error is not None and not any(
                 pattern.search(': '.join([error['error_type'], error['code'], error['msg']]))
                 for pattern in filters
             )
@@ -1079,9 +1079,16 @@ class Linter(metaclass=LinterMeta):
             # the working directory
             filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
 
-            # if this is a match for a different file we need its contents for the below checks
-            if not self.filename or not os.path.samefile(filename, self.filename):
-                vv = VirtualView.from_file(filename)
+            try:
+                # if this is a match for a different file we need its contents
+                # for the below checks
+                if not self.filename or not os.path.samefile(filename, self.filename):
+                    vv = VirtualView.from_file(filename)
+            except (FileNotFoundError, OSError) as err:
+                # warn about the error and drop this match
+                logger.warning('Exception: {}'.format(str(err)))
+                self.notify_failure()
+                return None
         else:
             # use the filename of the current view
             filename = self.view.file_name() or "<untitled {}>".format(self.view.buffer_id())
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -347,6 +347,11 @@ def group_by_filename_and_update(bid, view_has_changed, linter, errors):
             view = window.find_open_file(filename)
             if view:
                 this_bid = view.buffer_id()
+
+                # ignore errors of other files if their view is dirty
+                if this_bid != bid and view.is_dirty() and errors:
+                    continue
+
                 update_buffer_errors(this_bid, view_has_changed, linter, errors)
 
                 if this_bid == bid:
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -3,7 +3,6 @@
 from collections import defaultdict, deque
 from contextlib import contextmanager
 from functools import lru_cache, partial
-from itertools import groupby
 import logging
 import os
 import time
@@ -309,6 +308,9 @@ def kill_active_popen_calls(bid):
         proc.friendly_terminated = True
 
 
+affected_filenames_per_bid = defaultdict(set)
+
+
 def group_by_filename_and_update(bid, view_has_changed, linter, errors):
     """Group lint errors by filename and update them."""
     if view_has_changed():  # abort early
@@ -317,15 +319,35 @@ def group_by_filename_and_update(bid, view_has_changed, linter, errors):
     window = linter.view.window()
 
     # group all errors by filenames to update them separately
-    errors.sort(key=lambda e: e.get('filename', None) or '')
-    for filename, grouped_errors in groupby(errors, key=lambda e: e.get('filename', None)):
-        if filename is None:  # linted file
-            update_buffer_errors(bid, view_has_changed, linter, list(grouped_errors))
+    grouped = defaultdict(list)
+    for error in errors:
+        grouped[error.get('filename')].append(error)
+
+    # The contract for a simple linter is that it reports `[errors]` or an
+    # empty list `[]` if the buffer is clean. For linters that report errors
+    # for multiple files we collect information about which files are actually
+    # reported by a given `bid` so that we can clean the results. Basically,
+    # we must fake a `[]` response for every filename that is no longer
+    # reported.
+
+    lint_id = (bid, linter.name)
+    current_filenames = set(grouped.keys())  # `set` for the immutable version
+    previous_filenames = affected_filenames_per_bid[lint_id]
+    clean_files = previous_filenames - current_filenames
+
+    for filename in clean_files:
+        grouped[filename]  # For the side-effect of creating a new empty `list`
+
+    for filename, errors in grouped.items():
+        if not filename:  # backwards compatibility
+            update_buffer_errors(bid, view_has_changed, linter, errors)
         else:
             # search for an open view for this file to get a bid
             view = window.find_open_file(filename)
             if view:
-                update_buffer_errors(view.buffer_id(), view_has_changed, linter, list(grouped_errors))
+                update_buffer_errors(view.buffer_id(), view_has_changed, linter, errors)
+
+    affected_filenames_per_bid[lint_id] = current_filenames
 
 
 def update_buffer_errors(bid, view_has_changed, linter, errors):
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -176,6 +176,7 @@ class BackendController(sublime_plugin.EventListener):
             persist.view_linters.pop(bid, None)
 
             guard_check_linters_for_view.pop(bid, None)
+            affected_filenames_per_bid.pop(bid, None)
             buffer_syntaxes.pop(bid, None)
             queue.cleanup(bid)
 
@@ -308,7 +309,7 @@ def kill_active_popen_calls(bid):
         proc.friendly_terminated = True
 
 
-affected_filenames_per_bid = defaultdict(set)
+affected_filenames_per_bid = defaultdict(lambda: defaultdict(set))
 
 
 def group_by_filename_and_update(bid, view_has_changed, linter, errors):
@@ -330,9 +331,8 @@ def group_by_filename_and_update(bid, view_has_changed, linter, errors):
     # we must fake a `[]` response for every filename that is no longer
     # reported.
 
-    lint_id = (bid, linter.name)
     current_filenames = set(grouped.keys())  # `set` for the immutable version
-    previous_filenames = affected_filenames_per_bid[lint_id]
+    previous_filenames = affected_filenames_per_bid[bid][linter.name]
     clean_files = previous_filenames - current_filenames
 
     for filename in clean_files:
@@ -363,7 +363,7 @@ def group_by_filename_and_update(bid, view_has_changed, linter, errors):
     if not did_update_main_view:
         update_buffer_errors(bid, view_has_changed, linter, [])
 
-    affected_filenames_per_bid[lint_id] = current_filenames
+    affected_filenames_per_bid[bid][linter.name] = current_filenames
 
 
 def update_buffer_errors(bid, view_has_changed, linter, errors):
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -338,6 +338,7 @@ def group_by_filename_and_update(bid, view_has_changed, linter, errors):
     for filename in clean_files:
         grouped[filename]  # For the side-effect of creating a new empty `list`
 
+    did_update_main_view = False
     for filename, errors in grouped.items():
         if not filename:  # backwards compatibility
             update_buffer_errors(bid, view_has_changed, linter, errors)
@@ -345,7 +346,17 @@ def group_by_filename_and_update(bid, view_has_changed, linter, errors):
             # search for an open view for this file to get a bid
             view = window.find_open_file(filename)
             if view:
-                update_buffer_errors(view.buffer_id(), view_has_changed, linter, errors)
+                this_bid = view.buffer_id()
+                update_buffer_errors(this_bid, view_has_changed, linter, errors)
+
+                if this_bid == bid:
+                    did_update_main_view = True
+
+    # For the main view we MUST *always* report an outcome. This is not for
+    # cleanup but functions as a signal that we're done. Merely for the status
+    # bar view.
+    if not did_update_main_view:
+        update_buffer_errors(bid, view_has_changed, linter, [])
 
     affected_filenames_per_bid[lint_id] = current_filenames
 
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -3,6 +3,7 @@
 from collections import defaultdict, deque
 from contextlib import contextmanager
 from functools import lru_cache, partial
+from itertools import groupby
 import logging
 import os
 import time
@@ -289,7 +290,7 @@ def lint(view, view_has_changed, lock, reason=None):
     with remember_runtime(
         "Linting '{}' took {{:.2f}}s".format(util.canonical_filename(view))
     ):
-        sink = partial(update_buffer_errors, bid, view_has_changed)
+        sink = partial(group_by_filename_and_update, bid, view_has_changed)
         backend.lint_view(linters, view, view_has_changed, sink)
 
     events.broadcast(events.LINT_END, {'buffer_id': bid})
@@ -306,6 +307,25 @@ def kill_active_popen_calls(bid):
     for proc in procs:
         proc.terminate()
         proc.friendly_terminated = True
+
+
+def group_by_filename_and_update(bid, view_has_changed, linter, errors):
+    """Group lint errors by filename and update them."""
+    if view_has_changed():  # abort early
+        return
+
+    window = linter.view.window()
+
+    # group all errors by filenames to update them separately
+    errors.sort(key=lambda e: e.get('filename', None) or '')
+    for filename, grouped_errors in groupby(errors, key=lambda e: e.get('filename', None)):
+        if filename is None:  # linted file
+            update_buffer_errors(bid, view_has_changed, linter, list(grouped_errors))
+        else:
+            # search for an open view for this file to get a bid
+            view = window.find_open_file(filename)
+            if view:
+                update_buffer_errors(view.buffer_id(), view_has_changed, linter, list(grouped_errors))
 
 
 def update_buffer_errors(bid, view_has_changed, linter, errors):
--- a/tests/test_filter_results.py
+++ b/tests/test_filter_results.py
@@ -46,7 +46,7 @@ class _BaseTestCase(DeferrableTestCase):
 
 VIEW_UNCHANGED = lambda: False  # noqa: E731
 execute_lint_task = partial(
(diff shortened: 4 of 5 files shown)


kaste:
Hi @mheinzler! That's a lot for the effort ... but 😁 

Generally, all the views can handle multiple files very well, we probably don't need to change them. 

Now, you correctly start in `linter:process_match`. `vv` stands for `virtual_view` and you cannot use that bc it points to the main file. With it you cannot use `reposition_match` anymore. Instead just grab the filename. Use `col` for `start` and for `end` maybe `col +1`.

Return the dict, adding 'filename'. Note that filename at this point should be an absolute path. 

Next step is in `backend:finalize_errors`. Unfortunately, we access `linter.view` in here to compute a `sublime.Region` per error. This will get tricky bc you can only compute proper regions for open files. (Maybe just throw away the errors for closed files to keep it going.)

Next, we get to `sublime_linter:update_buffer_errors` which will broadcast the errors. Now, I wouldn't change `update_buffer_errors` bc it works. Instead change the `sink` on line 292 in `def lint` to a new function. Maybe `group_by_filename_and_update(view_has_changed, linter, errors)`, go over the errors, group them by filename, note the correct `bid` (buffer_id) per filename, and then broadcast them by calling `update_buffer_errors(bid, view_has_changed, linter, errors)`. 

Cross fingers 🤞 and see if the views play nice.

(To get a view for a filename, `window.find_open_file(file_name)` is probably the fastest API.)

mheinzler:
Hey @kaste,
thanks for the suggestions 😄.

If I understand correctly, you would want to handle these messages just like all others (as if they would have come from linting these files directly).

But I don't like to drop errors just because there is no open view for the file, because then you will miss a lot of information. What would you think about adding some special handling for when a `bid` is not available (e.g. skip marking the error regions in the non-existent view)? 

Another approach could be to simply open the files in Sublime automatically. This would resolve all of the issues at once, but of course could be annoying in some situations.

kaste:
> If I understand correctly, you would want to handle these messages just like all others (as if they would have come from linting these files directly).

I think that makes sense. The errors you show in the screenshot really belong to the '.hpp' files, these are not e.g. additional notes to the '.cpp' file. Your design suggest a parent/child relationship between these, but actually the '.hpp' have these errors, they still have the same errors if you included them in a different file. It is not the way you're using them in 'foo.cpp' that produces these errors.

For the rest: You don't get everything in a first iteration.

mheinzler:
> I think that makes sense. The errors you show in the screenshot really belong to the '.hpp' files, these are not e.g. additional notes to the '.cpp' file.

Yes, you're right.

> For the rest: You don't get everything in a first iteration.

I will try to open the files automatically. This should have the least impact on existing code and behavior.

braver:
> I will try to open the files automatically.

I don’t think the UX will be great if you open files as lint errors for them roll in. If we don’t have views for the files that have errors, we could instead  just list them (“file x has n errors”). You can then navigate to the file, which should kick off the linter for that file so we can properly display the errors in full detail.

mheinzler:
> I don’t think the UX will be great if you open files as lint errors for them roll in. If we don’t have views for the files that have errors, we could instead just list them (“file x has n errors”). 

Yes, this sounds a lot better.

>You can then navigate to the file, which should kick off the linter for that file so we can properly display the errors in full detail.

The problem with this is that at least the `clang-tidy` linter can normally only run on `.cpp` files and not the included `.hpp` files. This would mean you never see the errors because the linter always fails to run on the `.hpp` files.

Maybe we could store these errors with a `bid` of `None` to signal there is no open view yet. And when it has been opened we assign the errors by comparing the filenames and update everything.

braver:
But if the linter can’t lint the file directly we cannot update it as you edit, unlike the other files. It’ll be weird that `.hpp` behave differently. To me that sounds like we should treat them separately, e.g. “... and the linter also found this in file b while Iinting file a”. If a linter does pick up those secondary results, that’s fine. If it doesn’t that’s ok too because we don’t pretend that they’re normal lint results.

mheinzler:
That would leave us with something like in the above screenshot where all errors belong to the linted file. For the `clang-tidy` linter this would probably be the clearest behavior because it shows that the `cpp` file needs to be linted again to update the errors.

But for other linters it may make more sense to show the errors separately as kaste explained.

In the end both approaches will have their pros and cons. For now I would vote for the parent/child relationship I already implemented in some part :grin:.

kaste:
I don't know but both of you focus on later, nextnext problems. I outlined the _first_ steps towards a solution. We will of course not throw away errors of closed files, and we will not open them automatically. 

But we need to get our data repo in the right shape. Here I would only add 'filename' per error, then we need to compute the correct start, end, and region.  

The other thing looks completely different. Maybe more like a linter would return `(bid, errors, additional_notes)`, whereby _additional_notes_ is probably a free form text field we can print out in the panel. I don't think this is very useful, at least I don't remember it being requested in the last year or so. The classic multi filename feature *is* useful for many linters though - all go linters do that, probably all type checker _could_ do it.

mheinzler:
But how would you compute the regions without an open view?

How about we do it like [braver described](https://github.com/SublimeLinter/SublimeLinter/pull/1528#issuecomment-447364066) and then update the errors later when we receive an event that one of those views has been opened?

braver:
Just trying to think a few steps ahead and figure out the user perspective. I don’t know the data architecture by heart, but can respond to your suggestions, see if we can find something that fits. I already totally agree with the current PR, it’s a good starting point.

Ideally, coming from TypeScript, a typechecking linter can also lint the files that are imported. Although I guess it’s not uncommon that a file can only be analyzed after it’s imported. I think we can deal with that nicely if we think it trough (e.g. maintain a parent-child map, compute regions the child when it’s opened, and run the linter on the parent after a save on the child). But only if no linter is registered for the child file? Perhaps that’s next-next problems.

Would this work? `(bid, errors, [(additional_filename, errors)...])`

kaste:
It is probably way easier to do it this way.

The broader picture is. We do what I have outlined before. I do not like what 'linter.process_match' does right now because I want to make json linters also benefit from that code, but 'process_match' is right now exlusively for regex based linters. 

Now, generally speaking I want that a 'Linter' returns all information it has, and the backend then fills the holes, aka everything else. So, everything that involves the 'vv' (virtual_view) should be moved into 'backend.finalize_errors'. I also don't think that I need 'linter.view' in there for the Region because that's just a math op on what we have. 

Now, the key to compute everything is the 'VirtualView'. We can make a 'VirtualView.from_file' which reads the file and constructs a 'VirtualView' out of it. With that we can compute basically what we need, we can for example still call 'linter.reposition_match', and of course pin the line and column to valid ranges.

After that we should have valid errors, we can report back. 

What we probably do is that we slowly switch 'errors[bid]' to 'errors[unified_filename]'. The panel will then show errors from all open buffers plus all filenames that belong to that window.

kaste:
Overlapping with @braver here.

The common thing is for go linters that they only lint whole directories. For mypy to lint the whole project etc. 

Basically this is the same as running `flake8 .`. What the linters report is not a nested parent-child thing, they concat the errors from multiple files.

braver:
> The common thing is for go linters that they only lint whole directories

Yes, thanks for that. Got it now.

What I don’t get though, is why do we need to compute stuff like regions if we don’t have a real visible view for a file? Can’t we just report line and col as we get them from the linter, and compute on the view if and when it exists, i.e. lazily?

kaste:
Computing something when a view opens is more complicated because right now 'Linter' instances are not long lived but throw-away objects. We use them for the current on-going process and then throw'em away.

We also don't have paint pressure during the lint process, and we're already multi-threaded here. Also note that we actually use VirtualViews for most computations, the one instance we use 'linter.view' in the backend is a mistake I think.

Now, in the repo (our database) we need proper errors that have a common structure. If we allow incomplete errors e.g. without an 'end' or 'region' we need to make all views safe from this uncertainty. (Which we can do.) But then, if we update the errors (our main store), we need to re-broadcast them. (Which we can do.) I think this gets more complicated rather quickly.

This is probably expected because 'lazy' is usually an optimization strategy, and these always make the code more complicated, and often are premature.

braver:
Oh definitely, it would introduce a completely new data structure, “errors still in need of processing”. Cool, let’s not do that.

This raises other questions though. When a linter runs on more than one file, e.g. on a dir, when do we run it again? Obviously not on keypress, so on save? Tslint I can feed a project file, that’s easy enough to set up, but how do we detect that we need the mode to change (luckily Tslint is save-only... but flake8 can do dirs and lint live).

kaste:
The go linters all lint on save. Otherwise, we already have lint_reason which we pass around so we could probably teach flake8 to lint the project on save, and the buffer on modified.

mheinzler:
OK, all that sounds reasonable. We will see how this works out in practice and bigger projects later. I will get on it.

> I do not like what 'linter.process_match' does right now because I want to make json linters also benefit from that code, but 'process_match' is right now exlusively for regex based linters.

How would a linter pass the filename if not via regex? I'm no expert in python but it seems we can't simply extend `split_match` because of the tuple it returns. Do we need another method which a linter can overwrite in case it supports filenames?

kaste:
JSON based linters do not reach `split_match`, they should just return a dict per error from `parse_output`, just what we return at `process_match`.

kaste:
Aside: The 'unified' filename for a view is `view.file_name() or "<untitled {}>".format(view.buffer_id())`. This returns the correct (internal) filename for unsaved, unnamed buffers.

mheinzler:
Hey everyone,
happy new year and sorry for the delay.

Here is a first version of the new approach. It works well apart from renderding wrong regions. Of course this only works for already opened views. Any errors mentioning closed files are silently dropped.
grafik

Let me know what you think.

kaste:
That looks great! 

Two things, I would like the `filename` in the error map for all errors. Ideally, this simplifies the grouping in `sublime_linter`. If possible use `view.file_name() or "<untitled {}>".format(view.buffer_id())` as the current filename. 

For the regions, you could compute the region in `process_match` using the `VirtualView` like so:

```
line_start, _ = vv.full_line(line)
region = sublime.Region(line_start + start, line_start + end)
if len(region) == 0:
	region.b += 1
```

The only problem here is that the uid generator in `backend:finalize_errors` will then crash bc `json.dumps` doesn't understand `Regions` by default. (Probably `json.dumps(..., default=lambda o: (o.a, o.b) if isinstance(o, sublime.Region) else o` or something like that) In the same function we actually compute the regions right now, and of course we should only do this if not already provided.
 
🤞 ✋

kaste:
(Tip: We have a `SublimeLinter: Reload SublimeLinter` command in the `ctrl+p` thing. So you don't have to restart Sublime after every change. Just saying.)

mheinzler:
What should we do when a file does not exist? Is it OK to use the logger here with an error message?

Computing the regions early works but breaks some tests that call directly into the backend and use an offset. Maybe it is easier to also add the views to the error object (at least until the backend is fixed)?

One big problem I found is that errors will not disappear from the panel when they are corrected in the files. For the main file this can be easily fixed but for any other file we would again need a connection to the main file so that we know when to update these errors.

kaste:
I noticed the cleanup problem today and already had an almost ready patch here. 

For the offsets I applied a patch to the backend. 

We have 6 tests failing but I think they were wrong before and correct now. These come from the pre-Christmas patch for the out-of-bounds lines thing. I hate that.

So the expected is e.g. `[{'end': 10, 'line': 1, 'region': (0, 10), 'start': 0}]`. I don't think that makes sense. The input has two lines. So line '1' is actually valid. The point (line, col) (1, 0) then is at 11. The first line has 10 chars plus the `\n`. So the second line begins at 11. 

In this test I use `execute_lint_task` as if it were pure. But it isn't since it actually accesses the `view`. As I wrote these test I probably knew that and always used line `0` where this fact doesn't count bc line 0 always starts at pt 0. But as soon as you start testing for a line `1` you must stub/mock the view which I probably forgot about. I do this in the other multi-line tests e.g. `test_if_col_and_on_a_word_apply_offset_first_line` using the method `set_buffer_content`.

> What should we do when a file does not exist?

How can a linter report an error for a file that actually does not exists. 🤔

kaste:
Okay, I will check the failing tests against master tomorrow. We need to make the `filename` rock solid though. 

- The reported file does not exist, or we cannot read it. 
- The linter reports `main.go` which probably means `./main.go`.
- The linter actually reports `./foo/bar/thing.py` like python linters.

The latter two are likely relative paths compared to `self.get_working_dir(self.settings)`.

If everything fails we issue a `logger.warning`? `logger.error` is probably too much bc we can still handle useful output. We could or should drop such lint errors? We also have `self.notify_failure`.

kaste:
I fixed the tests in master and rebased. (Use `git fetch` + `git rest --hard` to get the new version.)

mheinzler:
How is this?

I based it on the implementation of `os.path.abspath`. The call to `normpath` should also have the advantage that we get the same path as Sublime Text uses for the view and so `window.find_open_file` should always work.

`OSError` is for when the path contains invalid characters. And as I just now found out it will also catch `IOError` (and `FileNotFoundError` which I will remove again :smile: ).

lint/linter.py
```diff
@@ -1066,6 +1072,27 @@ def split_match(self, match):
...
+        filename = m.match.groupdict().get("filename", None)
+        if filename:
+            # ensure that the filename is absolute by basing relative paths on
+            # the working directory
+            filename = os.path.normpath(os.path.join(self.get_working_dir(self.settings), filename))
+
+            try:
+                # if this is a match for a different file we need its contents
+                # for the below checks
+                if not self.filename or not os.path.samefile(filename, self.filename):
```
kaste: - Avoid `samefile` bc it does a `stat` call
- try/except clause is then too wide
- Can we just do a simple `self.filename != filename`. Do you have concerns about symlinks? Does `realpath` instead be okay?

kaste:
We need tests @mheinzler.

- Ensure that *no* new VirtualView gets used/created for 
	* `filename` is not given in regex
	* points to the same file as abs path e.g. '/foo/bar/main.go'
 	* points to e.g. 'main.go'
	* points to e.g. './main.go'

- Ensure that 'filename' is in the error map with the correct value.
	* same as above 

- Test that for errors from other files we have correct error maps. Esp. correct filename and region. 

- Ensure that reported files which cannot be found are evicted from the result

- Ensure backend does not add the offsets for other files. Offsets are only applicable for the current buffer.

...

lint/linter.py
mheinzler: Is this a performance issue for you or more general?

I mainly chose `samefile` because I wanted to make sure that paths like `a/../a/b` are considered equal to `a/b`. But now that we call `normpath` on `filename` a simple comparison should be OK for this.

But on Windows paths are case insensitive and so there is a (small) chance that Sublime opens a file as `C:\file` and we receive it as `c:\file`. I once ran into such a problem with a different plugin. What about this?
```suggestion
                if os.path.normcase(filename) != os.path.normcase(self.filename):
```

I think we can neglect symlinks but if we want to support them, I don't see an advantage of `realpath` over `samefile`.

mheinzler: I will switch to `normcase` then. Sublime Text and python (e.g. `abspath`) don't seem to care what the _correct_ path is. For example they open/return lower case drive letters unchanged.

mheinzler:
Here are the first few tests. I'm not sure how to test real files so I just used `test_regex_parsing.py` as the filename for now. Maybe we could add one or more test files which could be used for this? I guess with that all these tests could be made simpler and more practical.

kaste:
❤️ the tests! 

I added a test. I think just faking a VirtualView should be okay?

kaste:
We can also do 

```
        spy2(builtins.open)
        when(builtins).open(other_file, 'r').thenReturn(io.StringIO(OTHER_FILE_CONTENT))
```

🤔 	We must return universal newlines though. Sublime and the VirtualView only work with '\n's. `s.replace('\r\n', '\n').replace('\r', '\n')`. Don't think python3.3 has anything for it.

kaste:
The offset check must be implemented bc it is clearly a bug. Ideally, we do a filename check. This should lead to duplicated code between `finalize_errors` and `process_match` (similar to the region code we now have in both functions) which is a 👍 bc it _leads_ our way to future refactorings. (I always thought that what we do in `process_match` should be done in the backend.)

I don't need tests for `sublime_linter.py`. I don't have a good idea for the unit-under-test here. (For `sublime_linter.py` it should probably only matter that the correct events are sent through the system.)

We should already think about moving the `vv` generation out of `process_match` one level up into `parse_output_via_regex`. There we have 

```
	...
	yield self.process_match(m, virtual_view)
```

so we could inject the correct `vv` into `process_match`. The main thing here is to reuse the `vv` per filename so that we don't have the side-effect of opening or raising multiple times per reported filename.  

We had to think about what arguments really determines the VirtualView. Say we have an inlined function in there 

```
def parse_output_via_regex(...):
	@lru_cache()
	def get_virtual_view(filename, x, y ...):
		...
```

so that we can actually use a simple `lru_cache` on that fn. 

I could write an extensible `LintMatch` so that we can store `filename` on it. 👍 ?

kaste:
(Don't know why the except clause appears as uncovered. Clearly a coverage bug bc I just have to remove that logger statement and get a failing test.)

mheinzler:
What would you think about moving everything related to `VirtualView` (`start`, `end`, `region`) to the backend right away (maybe in a different PR)? This way `linter.py` would only be responsible for the frontend and everything else would be in the backend (all the stuff the plugin authors don't need to care about). At least if my understanding of the code is correct so far.

Extending `LintMatch` would certainly be great if this is what linters should return when they do the parsing themselves.

kaste:
Yes, the idea was to move VirtualView stuff more towards `finalize_errors`. The main problem here is probably backwards compatibility. `reposition_match` can be implemented by plugin authors and so we must call it or something that works very similar. This is a super important function since most linters only report `line:col` pairs but we want to highlight a word, or the whole line, t.i. we transform it to a region.

lint/backend.py
```diff
@@ -135,12 +135,20 @@ def finalize_errors(linter, errors, offsets):
...
+                    error['filename'] != "<untitled {}>".format(view.buffer_id())):
+                belongs_to_main_file = False
+
         line, start, end = error['line'], error['start'], error['end']
         if line == 0:
             start += col_offset
             end += col_offset
 
-        line += line_offset
+        if belongs_to_main_file:  # offsets are for the main file only
```
kaste: You forgot the col_offset above. 😁

mheinzler:
> The main problem here is probably backwards compatibility.
Just a thought: We could call these functions from the backend (and have them call a backend function again). So basically a simple redirection. Not very nice, but that way all that code could be moved and the plugins would still be able to intervene if they want.

kaste:
@braver can you check this already? We don't have linters that support it, so only check if it is backwards compatible. We could eventually push it to users although we don't have the feature correct. 

Possible roadmap:

- Testing this branch for compatibility
- I make LintMatch extensible in another branch
- I make a dirty flake8 plugin that will lint the whole working dir instead of the buffer.
- *After* the LintMatch PR we can move stuff from Linter to backend
- @mheinzler you can start with the panel in another branch. As always in ask anything mode.

braver:
Sorry, I just started a new job and didn’t keep up with everything. I’ll try and be of value here ASAP!

mheinzler:
I assumed `match` would always be a regex match, but the `eslint` linter returns a `dict` here. It should be fixed now.

This can be removed again when the filename is a part of `LintMatch`.

kaste:
Yes, I reviewed a dozen plugins. People use `match` to pass around additional information bc LintMatch and the former tuple are not extensible. They also just use the value `True` if they actually construct the LintMatch by themself (without using `super().split_match` at all).

I will add tests to the split-match contract TestCase.

braver:
onto the next it seems:

```
SublimeLinter: log_handler.py:74      Logging installed; log level INFO
SublimeLinter: sublime_linter.py:50   debug mode: on
SublimeLinter: sublime_linter.py:51   version: unknown
SublimeLinter: sublime_linter.py:252  Delay linting 'index.js' for 0.0s
SublimeLinter: linter.py:897          eslint: checking lint mode 'background' vs lint reason 'on_load'
SublimeLinter: #1 linter.py:914       eslint: linting 'index.js'
SublimeLinter: #1 util.py:132         PATH:
    /usr/local/bin
    /usr/bin
    /bin
    /usr/sbin
    /sbin
SublimeLinter: #1 linter.py:1345      Running ...

  /Users/koenlageveen/Sites/programmingfonts  (working dir)
  $ cat index.js | /usr/local/bin/eslint --format json --stdin --stdin-filename /Users/koenlageveen/Sites/programmingfonts/index.js

SublimeLinter: #1 linter.py:993       eslint: output:
  [{"filePath":"/Users/koenlageveen/Sites/programmingfonts/index.js","messages":[{"ruleId":"vars-on-top","severity":1,"message":"All 'var' declarations must be at the top of the function scope.","line":87,"column":3,"nodeType":"VariableDeclaration","endLine":87,"endColumn":71},{"ruleId":"no-console","severity":2,"message":"Unexpected console statement.","line":93,"column":3,"nodeType":"MemberExpression","messageId":"unexpected","endLine":93,"endColumn":16},{"ruleId":"curly","severity":1,"message":"Expected { after 'if' condition.","line":98,"column":4,"nodeType":"IfStatement","messageId":"missingCurlyAfterCondition","fix":{"range":[3239,3249],"text":"{return -1;}"}},{"ruleId":"curly","severity":1,"message":"Expected { after 'if' condition.","line":99,"column":4,"nodeType":"IfStatement","messageId":"missingCurlyAfterCondition","fix":{"range":[3306,3315],"text":"{return 1;}"}},{"ruleId":"curly","severity":1,"message":"Expected { after 'if' condition.","line":100,"column":4,"nodeType":"IfStatement","messageId":"missingCurlyAfterCondition","fix":{"range":[3368,3378],"text":"{return -1;}"}},{"ruleId":"curly","severity":1,"message":"Expected { after 'if' condition.","line":101,"column":4,"nodeType":"IfStatement","messageId":"missingCurlyAfterCondition","fix":{"range":[3431,3440],"text":"{return 1;}"}},{"ruleId":"no-unused-vars","severity":2,"message":"'toggleFavorite' is defined but never used.","line":130,"column":10,"nodeType":"Identifier","endLine":130,"endColumn":24},{"ruleId":"vars-on-top","severity":1,"message":"All 'var' declarations must be at the top of the function scope.","line":132,"column":3,"nodeType":"VariableDeclaration","endLine":132,"endColumn":71},{"ruleId":"semi","severity":2,"message":"Missing semicolon.","line":136,"column":6,"nodeType":"ExpressionStatement","fix":{"range":[4505,4505],"text":";"}},{"ruleId":"no-undef","severity":2,"message":"'Set' is not defined.","line":140,"column":67,"nodeType":"Identifier","endLine":140,"endColumn":70},{"ruleId":"no-console","severity":2,"message":"Unexpected console statement.","line":142,"column":3,"nodeType":"MemberExpression","messageId":"unexpected","endLine":142,"endColumn":16},{"ruleId":"no-useless-return","severity":1,"message":"Unnecessary return statement.","line":248,"column":5,"nodeType":"ReturnStatement","endLine":248,"endColumn":12,"fix":{"range":[7022,7761],"text":"function(event) {\n\t\tif (\n\t\t\tevent.target === document.querySelector(\".select-list\")\n\t\t\t&& ! event.ctrlKey\n\t\t\t&& ! event.altKey\n\t\t\t&& ! event.metaKey\n\t\t\t&& ! event.shiftKey\n\t\t) {\n\t\t\tif (event.key === \"ArrowUp\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tpreviousFont();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === \"ArrowDown\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tnextFont();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\n\t\tif (event.ctrlKey || event.metaKey) {\n\t\t\tif (event.key === \"-\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tdecreaseFontSize();\n\t\t\t\t\n\t\t\t} else if (event.key === \"=\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tincreaseFontSize();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\t}"}},{"ruleId":"no-useless-return","severity":1,"message":"Unnecessary return statement.","line":253,"column":5,"nodeType":"ReturnStatement","endLine":253,"endColumn":12,"fix":{"range":[7022,7761],"text":"function(event) {\n\t\tif (\n\t\t\tevent.target === document.querySelector(\".select-list\")\n\t\t\t&& ! event.ctrlKey\n\t\t\t&& ! event.altKey\n\t\t\t&& ! event.metaKey\n\t\t\t&& ! event.shiftKey\n\t\t) {\n\t\t\tif (event.key === \"ArrowUp\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tpreviousFont();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === \"ArrowDown\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tnextFont();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\n\t\tif (event.ctrlKey || event.metaKey) {\n\t\t\tif (event.key === \"-\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tdecreaseFontSize();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === \"=\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tincreaseFontSize();\n\t\t\t\t\n\t\t\t}\n\t\t}\n\t}"}}],"errorCount":5,"warningCount":8,"fixableErrorCount":1,"fixableWarningCount":6,"source":"/* global CodeMirror window document $ */\n/* eslint-disable no-implicit-globals */\n\n// CodeMirror\nvar editor = CodeMirror.fromTextArea(document.getElementById(\"code\"), {\n\tlineNumbers: true,\n\tstyleActiveLine: true,\n\tmatchBrackets: true,\n\ttheme: \"pastel-on-dark\",\n\tlineWrapping: true\n});\n\n// CodeMirror theme selector\nvar input = document.getElementById(\"select-theme\");\n\nfunction selectTheme() {\n\tvar theme = \"monokai\";\n\n\tif (input.selectedIndex > -1) {\n\t\ttheme = input.options[input.selectedIndex].innerHTML;\n\t}\n\teditor.setOption(\"theme\", theme);\n\tdocument.cookie = \"theme=\" + theme + \";max-age=172800\";\n}\n\n// ProgrammingFonts font selector\nfunction selectFont() {\n\tvar font = window.location.hash.substring(1);\n\n\tif (font === \"\") {\n\t\tfont = \"input\";\n\t}\n\n\tif (font === \"input\") {\n\t\t$(\"pre\").css({ fontFamily: \"Input Mono, monospace\" });\n\t\t$(\"textarea\").css({ fontFamily: \"Input Mono, monospace\" });\n\t\t$(\".CodeMirror\").css({ fontFamily: \"Input Mono, monospace\" });\n\t} else {\n\t\t$(\"pre\").css({ fontFamily: font + \", monospace\" });\n\t\t$(\"textarea\").css({ fontFamily: font + \", monospace\" });\n\t\t$(\".CodeMirror\").css({ fontFamily: font + \", monospace\" });\n\t}\n\t$(\"#select-font a[data-value]\").removeClass(\"active\");\n\t$(\"#select-font a[data-value='\" + font + \"']\").addClass(\"active\");\n\n\tdocument.cookie = \"font=\" + font + \";max-age=172800\";\n}\n\nwindow.onhashchange = selectFont;\n\nfunction setSize() {\n\tvar size = $(\"#size\").val();\n\n\t$(\".CodeMirror\").css({ fontSize: size + \"px\" });\n\tdocument.cookie = \"size=\" + size + \";max-age=172800\";\n}\nfunction setSpacing() {\n\tvar spacing = $(\"#spacing\").val();\n\n\t$(\".CodeMirror\").css({ lineHeight: spacing });\n\tdocument.cookie = \"spacing=\" + spacing + \";max-age=172800\";\n}\nfunction setAntialiasing() {\n\tif ($(\"#aliasing\").is(\":checked\")) {\n\t\t$(\".CodeMirror\").removeClass(\"no-smooth\");\n\t\tdocument.cookie = \"antialiasing=smooth;max-age=172800\";\n\t} else {\n\t\t$(\".CodeMirror\").addClass(\"no-smooth\");\n\t\tdocument.cookie = \"antialiasing=no-smooth;max-age=172800\";\n\t}\n}\nfunction selectLanguage() {\n\tvar lang = $(\"#select-language\").val();\n\n\teditor.setOption(\"mode\", lang.toLowerCase());\n\tdocument.cookie = \"language=\" + lang + \";max-age=172800\";\n}\n\nfunction renderSelectList() {\n\tvar icon = '<svg class=\"octicon\" viewBox=\"0 0 12 14\" version=\"1.1\" width=\"12\" height=\"14\" aria-hidden=\"true\"><path fill-rule=\"evenodd\" d=\"M11 10h1v3c0 .55-.45 1-1 1H1c-.55 0-1-.45-1-1V3c0-.55.45-1 1-1h3v1H1v10h10v-3zM6 2l2.25 2.25L5 7.5 6.5 9l3.25-3.25L12 8V2H6z\"></path></svg>';\n\tvar pinIcon = '<svg class=\"octicon octicon-pin\" viewBox=\"0 0 16 16\" version=\"1.1\" width=\"16\" height=\"16\" aria-hidden=\"true\"><path fill-rule=\"evenodd\" d=\"M10 1.2V2l.5 1L6 6H2.2c-.44 0-.67.53-.34.86L5 10l-4 5 5-4 3.14 3.14a.5.5 0 0 0 .86-.34V10l3-4.5 1 .5h.8c.44 0 .67-.53.34-.86L10.86.86a.5.5 0 0 0-.86.34z\"></path></svg>';\n\tvar favoritesMap = {};\n\n\t$(\"#select-font\").empty();\n\n\ttry {\n\t\tvar favorites = JSON.parse(localStorage.getItem('favorites')) || [];\n\t\tfavoritesMap = favorites.reduce(function(acc, alias) {\n\t\t\tacc[alias] = true;\n\t\t\treturn acc;\n\t\t}, {});\n\t} catch (err) {\n\t\tconsole.error('could not render favorites', err);\n\t}\n\n\t$.getJSON(\"fonts.json\", function(data) {\n\t\tdata.sort(function(a, b) {\n\t\t\tif (favoritesMap[a.alias] && !favoritesMap[b.alias]) return -1;\n\t\t\tif (!favoritesMap[a.alias] && favoritesMap[b.alias]) return 1;\n\t\t\tif (a.name.toLowerCase() < b.name.toLowerCase()) return -1;\n\t\t\tif (a.name.toLowerCase() > b.name.toLowerCase()) return 1;\n\t\t\treturn 0;\n\t\t});\n\t\t$.each(data, function(k, v) {\n\t\t\tvar liga_info = \"\";\n\t\t\tvar render_info = \"\";\n\n\t\t\tif (v.ligatures) {\n\t\t\t\tliga_info = \", ligatures\";\n\t\t\t}\n\t\t\tif (v.rendering === \"bitmap\") {\n\t\t\t\trender_info = \", bitmap\";\n\t\t\t}\n\t\t\t$(\"#select-font\").append(\n\t\t\t\t\"<div class='entry'>\" +\n\t\t\t\t\"<a href='#\" + v.alias + \"' data-value=\\\"\" + v.alias + \"\\\">\" +\n\t\t\t\t\t\"<span class='name'>\" + v.name + \"</span>\" +\n\t\t\t\t\t\"<span class='details'>\" + v.author + \" (\" + v.year + \") — \" + v.style + render_info + liga_info + \"</span>\" +\n\t\t\t\t\"</a>\" +\n\t\t\t\t\"<a class=\\\"favoritelink\" + (favoritesMap[v.alias] ? ' pinned' : '') + \"\\\" onclick=\\\"toggleFavorite('\" + v.alias + \"')\\\">\" +\n\t\t\t\t\tpinIcon +\n\t\t\t\t\"</a>\" +\n\t\t\t\t\"<a class='website' href='\" + v.website + \"' rel=external> <span>Info & Download</span>\" + icon + \"</a></div>\"\n\t\t\t);\n\t\t});\n\t\tselectFont();\n\t});\n}\n\nfunction toggleFavorite(alias) {\n\ttry {\n\t\tvar favorites = JSON.parse(localStorage.getItem('favorites')) || [];\n\t\tif (favorites.indexOf(alias) > -1) {\n\t\t\tfavorites = favorites.filter(function(v) {\n\t\t\t\treturn v !== alias;\n\t\t\t})\n\t\t} else {\n\t\t\tfavorites.push(alias);\n\t\t}\n\t\tlocalStorage.setItem('favorites', JSON.stringify(Array.from(new Set(favorites))));\n\t} catch (err) {\n\t\tconsole.error('could not save favorite', err);\n\t}\n\trenderSelectList();\n\treturn false;\n}\n\nfunction nextFont() {\n\tvar activeLink = document.querySelector(\".entry .active\");\n\tvar activeEntry = activeLink.parentNode;\n\tvar next = activeEntry.nextSibling;\n\tif (next && next.matches(\".entry\")) {\n\t\tnext.querySelector(\"a\").click();\n\t\tnext.scrollIntoView();\n\t}\n}\n\nfunction previousFont() {\n\tvar activeLink = document.querySelector(\".entry .active\");\n\tvar activeEntry = activeLink.parentNode;\n\tvar next = activeEntry.previousSibling;\n\tif (next && next.matches(\".entry\")) {\n\t\tnext.querySelector(\"a\").click();\n\t\tnext.scrollIntoView();\n\t}\n}\n\nfunction increaseFontSize() {\n    var sizeEl = document.getElementById(\"size\");\n    sizeEl.value = Number(sizeEl.value) + 1;\n    sizeEl.onchange();\n}\n\nfunction decreaseFontSize() {\n    var sizeEl = document.getElementById(\"size\");\n    sizeEl.value = Number(sizeEl.value) - 1;\n    sizeEl.onchange();\n}\n\n$(document).ready(function() {\n\tvar cookieValueSpacing = document.cookie.replace(/(?:(?:^|.*;\\s*)spacing\\s*=\\s*([^;]*).*$)|^.*$/, \"$1\");\n\tvar cookieValueSize = document.cookie.replace(/(?:(?:^|.*;\\s*)size\\s*=\\s*([^;]*).*$)|^.*$/, \"$1\");\n\tvar cookieValueAntialiasing = document.cookie.replace(/(?:(?:^|.*;\\s*)antialiasing\\s*=\\s*([^;]*).*$)|^.*$/, \"$1\");\n\tvar cookieValueTheme = document.cookie.replace(/(?:(?:^|.*;\\s*)theme\\s*=\\s*([^;]*).*$)|^.*$/, \"$1\");\n\tvar cookieValueLanguage = document.cookie.replace(/(?:(?:^|.*;\\s*)language\\s*=\\s*([^;]*).*$)|^.*$/, \"$1\");\n\n\tif (cookieValueSpacing !== \"\") {\n\t\t$(\"#spacing\").val(cookieValueSpacing);\n\t}\n\tif (cookieValueSize !== \"\") {\n\t\t$(\"#size\").val(cookieValueSize);\n\t}\n\tif (cookieValueAntialiasing === \"smooth\") {\n\t\t$(\"#aliasing\").prop(\"checked\", true);\n\t} else if (cookieValueAntialiasing === \"no-smooth\") {\n\t\t$(\"#aliasing\").prop(\"checked\", false);\n\t}\n\tif (cookieValueTheme !== \"\") {\n\t\t$(\"#select-theme\").val(cookieValueTheme);\n\t}\n\tif (cookieValueLanguage !== \"\") {\n\t\t$(\"#select-language\").val(cookieValueLanguage);\n\t}\n\n\tselectTheme();\n\tsetSize();\n\tsetSpacing();\n\tsetAntialiasing();\n\tselectLanguage();\n\trenderSelectList();\n\n\t$(\"#theme-next\").click(function() {\n\t\t$(\"#select-theme :selected\").next().prop(\"selected\", true);\n\t\tselectTheme();\n\t});\n\n\t$(\"#theme-previous\").click(function() {\n\t\t$(\"#select-theme :selected\").prev().prop(\"selected\", true);\n\t\tselectTheme();\n\t});\n\n\t$(\"body\").on(\"keydown\", function(event) {\n\t\tif (\n\t\t\tevent.target === document.querySelector(\".select-list\")\n\t\t\t&& ! event.ctrlKey\n\t\t\t&& ! event.altKey\n\t\t\t&& ! event.metaKey\n\t\t\t&& ! event.shiftKey\n\t\t) {\n\t\t\tif (event.key === \"ArrowUp\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tpreviousFont();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === \"ArrowDown\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tnextFont();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\n\t\tif (event.ctrlKey || event.metaKey) {\n\t\t\tif (event.key === \"-\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tdecreaseFontSize();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === \"=\") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tincreaseFontSize();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\t});\n});\n"}]
SublimeLinter: #1 linter.py:73        eslint output:
[{'errorCount': 5,
  'filePath': '/Users/koenlageveen/Sites/programmingfonts/index.js',
  'fixableErrorCount': 1,
  'fixableWarningCount': 6,
  'messages': [{'column': 3,
                'endColumn': 71,
                'endLine': 87,
                'line': 87,
                'message': "All 'var' declarations must be at the top of the function scope.",
                'nodeType': 'VariableDeclaration',
                'ruleId': 'vars-on-top',
                'severity': 1},
               {'column': 3,
                'endColumn': 16,
                'endLine': 93,
                'line': 93,
                'message': 'Unexpected console statement.',
                'messageId': 'unexpected',
                'nodeType': 'MemberExpression',
                'ruleId': 'no-console',
                'severity': 2},
               {'column': 4,
                'fix': {'range': [3239, 3249], 'text': '{return -1;}'},
                'line': 98,
                'message': "Expected { after 'if' condition.",
                'messageId': 'missingCurlyAfterCondition',
                'nodeType': 'IfStatement',
                'ruleId': 'curly',
                'severity': 1},
               {'column': 4,
                'fix': {'range': [3306, 3315], 'text': '{return 1;}'},
                'line': 99,
                'message': "Expected { after 'if' condition.",
                'messageId': 'missingCurlyAfterCondition',
                'nodeType': 'IfStatement',
                'ruleId': 'curly',
                'severity': 1},
               {'column': 4,
                'fix': {'range': [3368, 3378], 'text': '{return -1;}'},
                'line': 100,
                'message': "Expected { after 'if' condition.",
                'messageId': 'missingCurlyAfterCondition',
                'nodeType': 'IfStatement',
                'ruleId': 'curly',
                'severity': 1},
               {'column': 4,
                'fix': {'range': [3431, 3440], 'text': '{return 1;}'},
                'line': 101,
                'message': "Expected { after 'if' condition.",
                'messageId': 'missingCurlyAfterCondition',
                'nodeType': 'IfStatement',
                'ruleId': 'curly',
                'severity': 1},
               {'column': 10,
                'endColumn': 24,
                'endLine': 130,
                'line': 130,
                'message': "'toggleFavorite' is defined but never used.",
                'nodeType': 'Identifier',
                'ruleId': 'no-unused-vars',
                'severity': 2},
               {'column': 3,
                'endColumn': 71,
                'endLine': 132,
                'line': 132,
                'message': "All 'var' declarations must be at the top of the function scope.",
                'nodeType': 'VariableDeclaration',
                'ruleId': 'vars-on-top',
                'severity': 1},
               {'column': 6,
                'fix': {'range': [4505, 4505], 'text': ';'},
                'line': 136,
                'message': 'Missing semicolon.',
                'nodeType': 'ExpressionStatement',
                'ruleId': 'semi',
                'severity': 2},
               {'column': 67,
                'endColumn': 70,
                'endLine': 140,
                'line': 140,
                'message': "'Set' is not defined.",
                'nodeType': 'Identifier',
                'ruleId': 'no-undef',
                'severity': 2},
               {'column': 3,
                'endColumn': 16,
                'endLine': 142,
                'line': 142,
                'message': 'Unexpected console statement.',
                'messageId': 'unexpected',
                'nodeType': 'MemberExpression',
                'ruleId': 'no-console',
                'severity': 2},
               {'column': 5,
                'endColumn': 12,
                'endLine': 248,
                'fix': {'range': [7022, 7761],
                        'text': 'function(event) {\n\t\tif (\n\t\t\tevent.target === document.querySelector(".select-list")\n\t\t\t&& ! event.ctrlKey\n\t\t\t&& ! event.altKey\n\t\t\t&& ! event.metaKey\n\t\t\t&& ! event.shiftKey\n\t\t) {\n\t\t\tif (event.key === "ArrowUp") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tpreviousFont();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === "ArrowDown") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tnextFont();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\n\t\tif (event.ctrlKey || event.metaKey) {\n\t\t\tif (event.key === "-") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tdecreaseFontSize();\n\t\t\t\t\n\t\t\t} else if (event.key === "=") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tincreaseFontSize();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\t}'},
                'line': 248,
                'message': 'Unnecessary return statement.',
                'nodeType': 'ReturnStatement',
                'ruleId': 'no-useless-return',
                'severity': 1},
               {'column': 5,
                'endColumn': 12,
                'endLine': 253,
                'fix': {'range': [7022, 7761],
                        'text': 'function(event) {\n\t\tif (\n\t\t\tevent.target === document.querySelector(".select-list")\n\t\t\t&& ! event.ctrlKey\n\t\t\t&& ! event.altKey\n\t\t\t&& ! event.metaKey\n\t\t\t&& ! event.shiftKey\n\t\t) {\n\t\t\tif (event.key === "ArrowUp") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tpreviousFont();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === "ArrowDown") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tnextFont();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\n\t\tif (event.ctrlKey || event.metaKey) {\n\t\t\tif (event.key === "-") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tdecreaseFontSize();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === "=") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tincreaseFontSize();\n\t\t\t\t\n\t\t\t}\n\t\t}\n\t}'},
                'line': 253,
                'message': 'Unnecessary return statement.',
                'nodeType': 'ReturnStatement',
                'ruleId': 'no-useless-return',
                'severity': 1}],
  'source': '/* global CodeMirror window document $ */\n/* eslint-disable no-implicit-globals */\n\n// CodeMirror\nvar editor = CodeMirror.fromTextArea(document.getElementById("code"), {\n\tlineNumbers: true,\n\tstyleActiveLine: true,\n\tmatchBrackets: true,\n\ttheme: "pastel-on-dark",\n\tlineWrapping: true\n});\n\n// CodeMirror theme selector\nvar input = document.getElementById("select-theme");\n\nfunction selectTheme() {\n\tvar theme = "monokai";\n\n\tif (input.selectedIndex > -1) {\n\t\ttheme = input.options[input.selectedIndex].innerHTML;\n\t}\n\teditor.setOption("theme", theme);\n\tdocument.cookie = "theme=" + theme + ";max-age=172800";\n}\n\n// ProgrammingFonts font selector\nfunction selectFont() {\n\tvar font = window.location.hash.substring(1);\n\n\tif (font === "") {\n\t\tfont = "input";\n\t}\n\n\tif (font === "input") {\n\t\t$("pre").css({ fontFamily: "Input Mono, monospace" });\n\t\t$("textarea").css({ fontFamily: "Input Mono, monospace" });\n\t\t$(".CodeMirror").css({ fontFamily: "Input Mono, monospace" });\n\t} else {\n\t\t$("pre").css({ fontFamily: font + ", monospace" });\n\t\t$("textarea").css({ fontFamily: font + ", monospace" });\n\t\t$(".CodeMirror").css({ fontFamily: font + ", monospace" });\n\t}\n\t$("#select-font a[data-value]").removeClass("active");\n\t$("#select-font a[data-value=\'" + font + "\']").addClass("active");\n\n\tdocument.cookie = "font=" + font + ";max-age=172800";\n}\n\nwindow.onhashchange = selectFont;\n\nfunction setSize() {\n\tvar size = $("#size").val();\n\n\t$(".CodeMirror").css({ fontSize: size + "px" });\n\tdocument.cookie = "size=" + size + ";max-age=172800";\n}\nfunction setSpacing() {\n\tvar spacing = $("#spacing").val();\n\n\t$(".CodeMirror").css({ lineHeight: spacing });\n\tdocument.cookie = "spacing=" + spacing + ";max-age=172800";\n}\nfunction setAntialiasing() {\n\tif ($("#aliasing").is(":checked")) {\n\t\t$(".CodeMirror").removeClass("no-smooth");\n\t\tdocument.cookie = "antialiasing=smooth;max-age=172800";\n\t} else {\n\t\t$(".CodeMirror").addClass("no-smooth");\n\t\tdocument.cookie = "antialiasing=no-smooth;max-age=172800";\n\t}\n}\nfunction selectLanguage() {\n\tvar lang = $("#select-language").val();\n\n\teditor.setOption("mode", lang.toLowerCase());\n\tdocument.cookie = "language=" + lang + ";max-age=172800";\n}\n\nfunction renderSelectList() {\n\tvar icon = \'<svg class="octicon" viewBox="0 0 12 14" version="1.1" width="12" height="14" aria-hidden="true"><path fill-rule="evenodd" d="M11 10h1v3c0 .55-.45 1-1 1H1c-.55 0-1-.45-1-1V3c0-.55.45-1 1-1h3v1H1v10h10v-3zM6 2l2.25 2.25L5 7.5 6.5 9l3.25-3.25L12 8V2H6z"></path></svg>\';\n\tvar pinIcon = \'<svg class="octicon octicon-pin" viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M10 1.2V2l.5 1L6 6H2.2c-.44 0-.67.53-.34.86L5 10l-4 5 5-4 3.14 3.14a.5.5 0 0 0 .86-.34V10l3-4.5 1 .5h.8c.44 0 .67-.53.34-.86L10.86.86a.5.5 0 0 0-.86.34z"></path></svg>\';\n\tvar favoritesMap = {};\n\n\t$("#select-font").empty();\n\n\ttry {\n\t\tvar favorites = JSON.parse(localStorage.getItem(\'favorites\')) || [];\n\t\tfavoritesMap = favorites.reduce(function(acc, alias) {\n\t\t\tacc[alias] = true;\n\t\t\treturn acc;\n\t\t}, {});\n\t} catch (err) {\n\t\tconsole.error(\'could not render favorites\', err);\n\t}\n\n\t$.getJSON("fonts.json", function(data) {\n\t\tdata.sort(function(a, b) {\n\t\t\tif (favoritesMap[a.alias] && !favoritesMap[b.alias]) return -1;\n\t\t\tif (!favoritesMap[a.alias] && favoritesMap[b.alias]) return 1;\n\t\t\tif (a.name.toLowerCase() < b.name.toLowerCase()) return -1;\n\t\t\tif (a.name.toLowerCase() > b.name.toLowerCase()) return 1;\n\t\t\treturn 0;\n\t\t});\n\t\t$.each(data, function(k, v) {\n\t\t\tvar liga_info = "";\n\t\t\tvar render_info = "";\n\n\t\t\tif (v.ligatures) {\n\t\t\t\tliga_info = ", ligatures";\n\t\t\t}\n\t\t\tif (v.rendering === "bitmap") {\n\t\t\t\trender_info = ", bitmap";\n\t\t\t}\n\t\t\t$("#select-font").append(\n\t\t\t\t"<div class=\'entry\'>" +\n\t\t\t\t"<a href=\'#" + v.alias + "\' data-value=\\"" + v.alias + "\\">" +\n\t\t\t\t\t"<span class=\'name\'>" + v.name + "</span>" +\n\t\t\t\t\t"<span class=\'details\'>" + v.author + " (" + v.year + ") — " + v.style + render_info + liga_info + "</span>" +\n\t\t\t\t"</a>" +\n\t\t\t\t"<a class=\\"favoritelink" + (favoritesMap[v.alias] ? \' pinned\' : \'\') + "\\" onclick=\\"toggleFavorite(\'" + v.alias + "\')\\">" +\n\t\t\t\t\tpinIcon +\n\t\t\t\t"</a>" +\n\t\t\t\t"<a class=\'website\' href=\'" + v.website + "\' rel=external> <span>Info & Download</span>" + icon + "</a></div>"\n\t\t\t);\n\t\t});\n\t\tselectFont();\n\t});\n}\n\nfunction toggleFavorite(alias) {\n\ttry {\n\t\tvar favorites = JSON.parse(localStorage.getItem(\'favorites\')) || [];\n\t\tif (favorites.indexOf(alias) > -1) {\n\t\t\tfavorites = favorites.filter(function(v) {\n\t\t\t\treturn v !== alias;\n\t\t\t})\n\t\t} else {\n\t\t\tfavorites.push(alias);\n\t\t}\n\t\tlocalStorage.setItem(\'favorites\', JSON.stringify(Array.from(new Set(favorites))));\n\t} catch (err) {\n\t\tconsole.error(\'could not save favorite\', err);\n\t}\n\trenderSelectList();\n\treturn false;\n}\n\nfunction nextFont() {\n\tvar activeLink = document.querySelector(".entry .active");\n\tvar activeEntry = activeLink.parentNode;\n\tvar next = activeEntry.nextSibling;\n\tif (next && next.matches(".entry")) {\n\t\tnext.querySelector("a").click();\n\t\tnext.scrollIntoView();\n\t}\n}\n\nfunction previousFont() {\n\tvar activeLink = document.querySelector(".entry .active");\n\tvar activeEntry = activeLink.parentNode;\n\tvar next = activeEntry.previousSibling;\n\tif (next && next.matches(".entry")) {\n\t\tnext.querySelector("a").click();\n\t\tnext.scrollIntoView();\n\t}\n}\n\nfunction increaseFontSize() {\n    var sizeEl = document.getElementById("size");\n    sizeEl.value = Number(sizeEl.value) + 1;\n    sizeEl.onchange();\n}\n\nfunction decreaseFontSize() {\n    var sizeEl = document.getElementById("size");\n    sizeEl.value = Number(sizeEl.value) - 1;\n    sizeEl.onchange();\n}\n\n$(document).ready(function() {\n\tvar cookieValueSpacing = document.cookie.replace(/(?:(?:^|.*;\\s*)spacing\\s*=\\s*([^;]*).*$)|^.*$/, "$1");\n\tvar cookieValueSize = document.cookie.replace(/(?:(?:^|.*;\\s*)size\\s*=\\s*([^;]*).*$)|^.*$/, "$1");\n\tvar cookieValueAntialiasing = document.cookie.replace(/(?:(?:^|.*;\\s*)antialiasing\\s*=\\s*([^;]*).*$)|^.*$/, "$1");\n\tvar cookieValueTheme = document.cookie.replace(/(?:(?:^|.*;\\s*)theme\\s*=\\s*([^;]*).*$)|^.*$/, "$1");\n\tvar cookieValueLanguage = document.cookie.replace(/(?:(?:^|.*;\\s*)language\\s*=\\s*([^;]*).*$)|^.*$/, "$1");\n\n\tif (cookieValueSpacing !== "") {\n\t\t$("#spacing").val(cookieValueSpacing);\n\t}\n\tif (cookieValueSize !== "") {\n\t\t$("#size").val(cookieValueSize);\n\t}\n\tif (cookieValueAntialiasing === "smooth") {\n\t\t$("#aliasing").prop("checked", true);\n\t} else if (cookieValueAntialiasing === "no-smooth") {\n\t\t$("#aliasing").prop("checked", false);\n\t}\n\tif (cookieValueTheme !== "") {\n\t\t$("#select-theme").val(cookieValueTheme);\n\t}\n\tif (cookieValueLanguage !== "") {\n\t\t$("#select-language").val(cookieValueLanguage);\n\t}\n\n\tselectTheme();\n\tsetSize();\n\tsetSpacing();\n\tsetAntialiasing();\n\tselectLanguage();\n\trenderSelectList();\n\n\t$("#theme-next").click(function() {\n\t\t$("#select-theme :selected").next().prop("selected", true);\n\t\tselectTheme();\n\t});\n\n\t$("#theme-previous").click(function() {\n\t\t$("#select-theme :selected").prev().prop("selected", true);\n\t\tselectTheme();\n\t});\n\n\t$("body").on("keydown", function(event) {\n\t\tif (\n\t\t\tevent.target === document.querySelector(".select-list")\n\t\t\t&& ! event.ctrlKey\n\t\t\t&& ! event.altKey\n\t\t\t&& ! event.metaKey\n\t\t\t&& ! event.shiftKey\n\t\t) {\n\t\t\tif (event.key === "ArrowUp") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tpreviousFont();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === "ArrowDown") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tnextFont();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\n\t\tif (event.ctrlKey || event.metaKey) {\n\t\t\tif (event.key === "-") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tdecreaseFontSize();\n\t\t\t\treturn;\n\t\t\t} else if (event.key === "=") {\n\t\t\t\tevent.preventDefault();\n\t\t\t\tevent.stopPropagation();\n\t\t\t\tincreaseFontSize();\n\t\t\t\treturn;\n\t\t\t}\n\t\t}\n\t});\n});\n',
  'warningCount': 8}]
SublimeLinter: #1 __init__.py:1273    ERROR: Unhandled exception:
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/backend.py", line 111, in execute_lint_task
    errors = linter.lint(code, view_has_changed) or []
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 932, in lint
    return self.filter_errors(self.parse_output(output, virtual_view))
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 957, in filter_errors
    for error in errors
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 956, in <listcomp>
    error
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1003, in parse_output_via_regex
    yield self.process_match(m, virtual_view)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/SublimeLinter/lint/linter.py", line 1081, in process_match
    if filename:
UnboundLocalError: local variable 'filename' referenced before assignment
SublimeLinter: sublime_linter.py:502  Linting 'index.js' took 0.49s
```

braver:
Ok, looks good now. I don't see anything wrong from a cursory test right now. I'll run it with a couple of plugins the coming days, see if anything turns up.

kaste:
Okay, `LintMatch` is in #1539

kaste:
In `VirtualView.from_file` I tell it to open the file with `encoding='utf8'` now. Is that a Windows thing? It crashes for me otherwise (depends on the file obviously).

braver:
I don't know anything about utf8 or paths on Windows. I tend to pretend Windows doesn't exist.

I ran on this branch today with eslint (on stdin and embedded files) and stylelint (on save). Both seem to work just fine, although I didn't look at the output panel too closely. Perhaps should do that tomorrow. Anyway, looks pretty good already.

@kaste want me to run with #1539 and https://github.com/SublimeLinter/SublimeLinter-eslint/pull/268? Or do you want to merge this PR first, release that, and then move forward?

braver:
... Or #1540 and just close both this and #1539?

mheinzler:
The new `LintMatch` looks great.

I'm not having any problems with or without your encoding change. Are you really seeing a crash with no error message? That would seem very strange (even for Windows 😄). But even so, specifying `utf8` is definitely better than leaving the choice to the system.

About your flake8 regex: Why not simply keep the non-greedy regex: `(?P<filename>.+?):`? This worked for me and should produce the same matches. Also Windows paths are valid with forward slashes and at least with my build tools I see them a lot, so if you don't like it non-greedy you should use something like `[\\/]` to catch these too.

I guess treating `stdin` as the main file would be a good idea. But I currently have no linter that produces such output.

braver:
I'm basically OK with merging this and pushing it out. I doubt I run into anything more and we probably really need to expand the test base beyond us three. There might be a couple of people out there that are still on the pre-release channel, wanna roll out to those first? We can announce what we're doing and get some more testers that way perhaps. In my experience that approach doesn't really amount to much unless there is a lot of immediate value for end users (which this PR by itself doesn't provide). So, I guess the question is how confident you guys are about this.. wanna take the plunge?

kaste:
@mheinzler Yes, I made the regex too complicated. I thought, shit I must match a filepath on Windows, but in the context of the typical linter output the lazy `+?` should be good enough. Hej, actually even the dumb `(?P<filename>.+):` can work.

Of course it didn't crash, it raised an Exception we don't catch saying that `cp1252` (default encoding) can't read blahblah.

mheinzler:
Everything looks good to me. I guess the smaller improvements can be added later.

Feel free to squash/merge in whatever order you think is best.

mheinzler:
I rebased everything on master and integrated your changes from #1540 (I have no access there).

When I tested some python linters I ran into problems with pyflakes and pydocstyle, because both of them already use the filename group. pyflakes prints "<stdin>" as a filename and pydocstyle uses temp files which are no longer valid when the errors are processed.

This is fixed now by treating all of these errors as coming from the main file.

kaste:
Interesting. Now we actually own both linters and they don't do anything with the filename so we could also just don't capture it. 

The tempfile thing will break mypy which works with tempfiles but then outputs real filenames. (You can tell it do so via a simple mapping.)

We generally shouldn't forget that thanks to extensible LintMatch plugins can change the captured filename in `split_match` if we can't handle it automatically. For the implementation, it looks like I create a new context object here and put the tempfile in there. I actually want to change that and put it in the original context because it is useful information for plugins. Also the context will then be available via `self.context`. (The context is currently available under `self.settings.context`. Why the hell didn't I make it `self.context.settings`?)

I have found two issues:

- `get_working_dir` actually returns `Optional[str]` and then it blows of course if it returns `None`. So we either do a `realpath` here or maybe better in `get_working_dir`. (Actually I don't know from top of my head if `realpath(None)` works on linux, it works on Windows though.)

- I think we shouldn't apply the errors if the secondary views are actually dirty. This is a confusing user experience. Sometimes I actually fixed these errors already without saving, sometimes I just edited these files and then all drawn regions are off of course. Line, col, and regions are just not applicable for dirty views.

mheinzler:
> The tempfile thing will break mypy which works with tempfiles but then outputs real filenames. (You can tell it do so via a simple mapping.)

I thought long about such a situation and then decided to not implement it 😄. Will fix it to check for both possibilities.

> We generally shouldn't forget that thanks to extensible LintMatch plugins can change the captured filename in `split_match` if we can't handle it automatically.

I would still keep the changes here because I think they are useful to many linters and make them simpler. If a plugin needs to filter exotic filenames it can do it like you say.

> I actually want to change that and put it in the original context because it is useful information for plugins.

👍. Then we can also drop the `self.temp_filename` variable again.

> `get_working_dir` actually returns `Optional[str]` and then it blows of course if it returns `None`.

`realpath(None)` fails for me on Linux. Is ignoring the cwd OK in such a case? Relative paths will fail of course, but as this is most likely because of a configuration error it shouldn't really matter.

> I think we shouldn't apply the errors if the secondary views are actually dirty. This is a confusing user experience.

True, but personally, I would like to see that there still is a problem (even if it is just because of an unsaved file). Maybe error out in such a case or display a visible warning?

kaste:
But `get_working_dir() or realpath('.')` is 👌 I guess.

> ... I would like to see that there still is a problem (even if it is just because of an unsaved file). Maybe error out in such a case or display a visible warning?

We would not report that file as clean! If the file had errors before they stay.

kaste:
Ah, don't start on the panel, we have PR's open and you get into merge hell, @mheinzler.

kaste:
LGTM! I want to merge. @braver? (Off-topic: the panel here is scary compared to the new PR's)

braver:
Yeah, looks good to me. Doesn't seem to break anything I use.

kaste:
Okay, let's have fun.

kaste:
🙌

PR #1532, opened 2019-01-04
Update documentation

dnicolson:
This pull request improves the documentation for ignoring ESLint in HTML files, and also uses double quotes so it can be copied as valid JSON.


braver:
Cool, thanks!

PR #1538, opened 2019-01-28
Prevent scrolling the x-axis in the panel view

### LINKED ISSUE #1537: Panel View horizontal scrolling is not stable
If the messages become very long, so that the panel view shows scroll bars for both directions, we scroll in all directions depending on where the cursor is. 

- For testing, just make the window small enough.
- Now place the cursor on an error. The line will be selected and Sublime scrolls showing the end of the selection. 

We want to restrict scrolling to the y-axis here.

kaste:
Fixes #1537

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -436,10 +436,20 @@ def update_selection(panel, region=None):
 
 class _sublime_linter_update_selection(sublime_plugin.TextCommand):
     def run(self, edit, a, b):
+        x1, y1 = self.view.viewport_position()
+
         region = sublime.Region(a, b)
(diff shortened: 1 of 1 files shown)


braver:
Can't we simply set the `x` to `0`?

kaste:
Yeah, that's possible. I actually had the panel scrolled to the right to read the messages. If we hardcode to `0` we would then scroll to the left whenever the user clicks in the view. I think if we do it this way, we just 'respect' the x-axis value the user actually had.

braver:
Ah, ok. 👍 then.

PR #1539, opened 2019-01-28
Extensible `LintMatch`

### LINKED ISSUE #1475: Add a "NOTE" severity level

kaste:
`LintMatch` is now a basically a dict. It is roughly compatible with the previous namedtuple, tuple. 

Added `filename`, `error_type`, and `code` as common capture groups in the regex. That means if the user explicitly captures `code` or `error_type` we will just take that value and don't compute it using `m.error or m.warning` anymore. 

`filename` is obviously for #1528.

Unlike a dict, _all present_ keys can be accessed using dot chaining. E.g.

```
	error = LintMatch(foo='bar)
	error.foo  # 'bar'
	error.quux  # AttributeError
```

All commonly used names like `line`, `col` etc. can be accessed using the dot even if there are not present. 

```
	error = LintMatch()
	error.line  # None
```

This is basically user convenience since a lot of plugin authors probably are not used to python that much. 

'Proper' python in usually

```
	error.get('line', None)
	# OR:
	try:
		line = error['line']
	except KeyError:
		...
	else:
		# Do something with line
```

We also do NOT normalize the `match.groupdict()`. The groudict has a value for every named group in the regex, this value includes `None` and the empty string. So, we could say *we* as framework authors have three falsy values to deal with: KeyError, None, `''`. I think this is again for the user: she can be sure that every name from the regex finds it place in the dict. 

Fixes #1475

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1096,7 +1096,7 @@ class Linter(metaclass=LinterMeta):
         return error
 
     def process_match(self, m, vv):
-        error_type = self.get_error_type(m.error, m.warning)
+        error_type = m.get('error_type', None) or self.get_error_type(m.error, m.warning)
 
         col = m.col
         line = m.line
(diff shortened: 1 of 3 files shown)


kaste:
Note that we can pass `error_type=note` just fine.

image

kaste:
🎉

PR #1541, opened 2019-01-29
Calculate column widths in the error panel

kaste:
Current state is that e.g. the linter name column is 12 chars wide etc.

With this patch we compute the max length we need per column and use these values.


kaste:
Before:

image

After:

image


Before:

image


After:

image

braver:
This is really smart and nice! Awesome. ⛵️ it!

PR #1542, opened 2019-01-29
Meta-mark the whole line containing the filename

kaste:
Unfortunately we can only mark up to the `$` and not to the width of the view. So this is the best I can come up with right now.

### DIFF
--- a/panel/panel.sublime-syntax
+++ b/panel/panel.sublime-syntax
@@ -9,9 +9,11 @@ contexts:
     - include: linter-body
 
   linter-body:
-    - match: '^(.+):$'
-      captures:
-        1: entity.name.filename.sublime_linter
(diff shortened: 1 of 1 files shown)


braver:
Screenshot?

kaste:
Here comes the ugly screenshot

image

Look, I would like (probably, maybe it looks shit) mark the whole line like a `<div>` block, but I actually don't know how to make such `<div>`s within Sublime. So if I mark the 'whole' first line, Sublime actually only marks until the `\n` char. So what you see as grey box is all **I** get, and it's better than nothing.

braver:
Oh, right. Yeah, layout doesn't work like that in Sublime. But it's a good idea to mark it with a scope like that, let the color schemes figure out how it should look. 🛥 it!

Issue #1415, opened 2018-05-18
output panel only for current view


braver:
The output panel lists all result for all files in the current window. It would be nice if you could have a panel with just the errors for the current file. 

I imagine this would be setting:

```
"show all errors per": "window" | "view"
```

ouweiya:
+1

ouweiya:
Is there any way to temporarily work around?
When there are ten files, the error message will become very large?

kaste:
@ouweiya I'm not aware of any work around here. I think this really needs to be implemented, aka the hard way.

ouweiya:
@kaste I think there is a compromise way, just clear the output records.
When I'm used to using `"lint_mode": "manual"`, it remembers the history of previous errors and can only be cleared by closing the file.
![c](https://user-images.githubusercontent.com/28357731/46008999-da2f0a00-c0f0-11e8-9855-0afa89b07cec.gif)

kaste:
Exactly when should should we forget the old results? (This would also erase all gutter marks and code highlights btw.)

haolly:
Or just provide some command to clear the result

PR #1544, opened 2019-01-30
Highlight nearby error row in panel

### LINKED ISSUE #1548: Re-evaluate calling `fill_panel` on the main thread
#1545 introduces that `fill_panel` from the panel view gets called on the main thread. This is merely to reduce unwanted x-axis scrolling, and visual jank.

- Do we really need `sel()` restoring after `view.replace`?
- Can we at least just `sel().clear()` before `replace`?
- Ensure that the conditional `if` after `set_timeout(fill_panel)` can be called immediately and not after `fill_panel`.

### LINKED ISSUE #1549: Improve orientation in error panel
For testing you need to have more errors than fit on screen, also ideally coming from multiple files.

- If you navigate your buffer, it looks like we're showing the error line always in the middle of the panel, which results in visual clutter bc the panel always scrolls back and forth.

- Almost never, the current filename line is at the top of the panel. 

Instead, we should try to scroll so that the current filename gets on the first shown line of the panel. Navigating within a file should only scroll if necessary. We should try to avoid scrolling too much. Errors from the next file should not appear if possible (t.i. if we still have enough errors to show from the current active file).

### LINKED ISSUE #1554: Remove the line marker in the error panel if a file gets clean
You can get something like this

image

if you had errors in the active view, and then clean them up. 

Looks like we exit early if a view/file has no errors but we must cleanup the panel of course.

kaste:
Current state: In the error panel we draw a tiny mark (as cursor) if we're somewhere in the code, and if we're on an error, we highlight the whole line. 

This is not good enough IMO. If you have a lot of information in the panel (e.g. you use mypy and really have to read that stuff) it is too hard to find the error you're actually staring at bc clearly most of the time your cursor is only 'nearby' an error and not directly on it. 

I propose that we highlight the nearest error. We don't highlight anything (t.i. we still have the tiny marker) if we're just somewhere in the code. The snap distance is 5 lines. (See the `SNAP` constant in the code.)

Fixes #1554 
Fixes #1549 
Fixes #1548

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -450,11 +450,6 @@ def update_panel_selection(active_view, cursor, **kwargs):
 
     mark_visible_viewport(panel, active_view, all_errors)
 
-    if not all_errors:
-        draw_position_marker(panel, None)
-        mark_lines(panel, None)
-        return
-
     row, _ = active_view.rowcol(cursor)
     errors_with_position = (
         (
@@ -504,8 +499,12 @@ def update_panel_selection(active_view, cursor, **kwargs):
                 if error['region'].begin() > cursor
             )
         except StopIteration:
-            last_error = all_errors[-1]
-            panel_line = last_error['panel_line'] + 1
+            try:
+                last_error = all_errors[-1]
+            except IndexError:
+                panel_line = None
+            else:
+                panel_line = last_error['panel_line'] + 1
         else:
             panel_line = next_error['panel_line']
 
--- a/panel_view.py
+++ b/panel_view.py
@@ -430,6 +430,8 @@ def update_panel_selection(active_view, cursor, **kwargs):
         return
 
     all_errors = sorted(persist.errors[bid], key=lambda e: e['panel_line'])
+    mark_visible_viewport(panel, active_view, all_errors)
+
     errors_under_cursor = [
         error
         for error in all_errors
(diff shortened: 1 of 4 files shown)


braver:
I get the point, the requirement to be *on* an error for it to highlight is perhaps to strict. But if there is an error on line 7 to me it feels really weird that it highlights when I'm 5 lines away. 

highlight

With SNAP at 1,0 it makes sense to me, highlight whatever's on the same line (but isn't that what we do already? I hardly ever use the panel). 2,0 also works for me. Wider than that and the highlighting just stops making sense to me.

panel_view.py
```diff
@@ -382,51 +379,132 @@ def update_panel_selection(active_view, current_pos, **kwargs):
...
+                 sublime.DRAW_NO_OUTLINE | sublime.DRAW_EMPTY_AS_OVERWRITE)
+        panel.add_regions(KEY, regions, scope=scope, flags=flags)
+
+        scope = 'comment'
+        scope = 'region.bluish'
+        flags = sublime.DRAW_NO_OUTLINE
+        head_line = panel.text_point(head['panel_line'], 0)
+        end_line = panel.text_point(end['panel_line'] + 1, 0)
+        regions = [sublime.Region(r.a, r.a + 1) for r in panel.lines(sublime.Region(head_line, end_line))]
+        panel.add_regions(KEY2, regions, scope=scope, flags=flags)
```
kaste: Comment one of the `add_regions` lines to hide that marker. The first one adds small cursors to the left. The second one highlights the whole area with short FILL regions. Colors (scopes) can be changed as well.

kaste:
E.g.

image

image

kaste:
- Implemented own scrolling behavior
- (Try to pin the current filename at the top of the panel.)
- Sort the panel sections by filename (Why did we have a random order here for so long?)

Easier to put this on top of this PR here. SNAP value is at 3 atm after a week. Experimental viewport marker is mostly off. I don't want user facing options right now. Maybe we ship experimental code under a flag so that we can further test stuff without going to merging conflicts hell. Setting SNAP to 1 is probably not very controversial. So probably not necessary to unmerge/merge that code over and over again only because we don't know what a good default here is. 

(Just like in web dev, we would ship code to production usually invisible for users but we already know roughly that it is the right direction just bc we don't want merging hell bc it would slow us down and we still want to ship whenever we want.)

braver:
I'm going to try this out for a while. I don't use the panel often, or not intensively. Either the code has very little errors if any and I quickly squash them, or it has so many errors that I don't even try to grok what's in the panel and rely on the tooltips and statusbar for info. But, I'll give it a whirl.

I already like the highlighting and could get used to this SNAP value. 1 or 2 might be a better default, but I get why you like it like this. 

I didn't understand what's going on here though. 

screenshot 2019-02-12 at 22 38 01

It looks kinda weird and blobby if you look closely. But I also didn't intuitively understand what it means. Now I read your info and I get that it's the viewport and I understand why that's useful. Perhaps we should do the markers, but not the backgrounds. Or perhaps there is a border-radius on the areas that we can disable so that it looks right?

Scrolling behaviour looks solid. Nice work man.

kaste:
But look at my pics above where the background just looks nice. I have a opacity 10% here and it looks 👌 

We had this before; super hard to find scopes that just work. Aren't you a designer, can't we ship/showcase template schemes a user can copy-paste and then edit for a better experience. I don't want to patch user schemes, but I would appreciate a template with variables that make sense. Here *I* have the same color for the marker and the background just the opacity is different. The background is barely visible but if you turn it off you're clueless because it doesn't fill a space between the markers.

Anyway. With this scrolling thing. We could use this to solve #1415. If I sort the errors so that the focused view always comes last, the scroller will pin the last file at the top of the viewport essentially hiding all other errors. (All other errors will still be there above the current filename, and a user can easily scroll to them manually.) This will look more calm and focused. (The PR already is calm and focused compared to the master version.)

braver:
> I have a opacity 10% here and it looks

Yeah, no, it doesn't. You still get the rounded blobbiness, you probably just don't see it. I do though. Areas are rounded according to the color scheme, I'm not sure what we can do here. Might be worth a try to insert a minihtml phantom?

> With this scrolling thing. We could use this to solve #1415.

I already like the behavior a lot better than before. If we can get a good default behavior we don't need a setting. With what you're proposing though, I wonder what happens if you use the panel to switch to a particular error in a different file? Wouldn't the panel completely re-arrange itself then?

kaste:
> Yeah, no, it doesn't.

🤣 I *can* see them but barely noticeable at least when coding. Phantoms are really are stupid part of Sublime and they don't work here. I also think Sublime could do better with the `region.colorish` scopes. They usually have a nice foreground color set but their background value never looks good. (I tried regions to highlight the current filename but they look shitty. Mostly too greyish, not vibrant at all. Way too low contrast with the standard foreground/text color....)

> Wouldn't the panel completely re-arrange itself then?

As I understand #1415 it proposes that the panel only shows the errors from the current file which basically means the panel redraws completely on_activated. No? My interpretation of that issue is that the panel is too confusing and basically unfocused. This PR already helps. An idea, don't know how it looks, is to leave all errors in the panel but to visually hide them. If the solution would produce visual yank on_activated t.i. you can perceive the re-arrangement we cannot do it of course.

kaste:
I don't intend to ship the background viewport stuff bc it is def ugly by default. Unfortunately if we set a `foo.not_here` scope a user could then define Sublime uses greyish/whitish for that. Do we have a transparent/backgroundish? (`region.backgroundish.panel_viewport.sublime_linter`)

braver:
I have been super concentrated on other work the last few days, so I've only just now started to really try out this branch and the changes to the panel. 

The scrolling and switching between files seems really solid and much better than master. It's basically instant when I switch files and not disorienting at all.

It does make me wonder why we have all the errors in the panel in the first place. Shouldn't the default panel just be for the view that has focus, and perhaps introduce a new panel for all open files? The only times I open the panel are when I have a lot of errors (e.g. JS files where previous authors were sloppy with indentation, `;` and `===`) and I want to get a feel for what's up. So, for what I need just one file in the panel is actually better. That said, I don't use the panel more that once a month.

I find that the viewport markers don't update with file changes in a way that I understand, it looks like on of the markers goes missing?
viewport


>Do we have a transparent/backgroundish

Not that I know of.

kaste:
The viewport markers are only a fast draft implementation to get a feeling how they would look like.

braver:
Ok, I think they look good, they’re obviously secondary to the current location marker, and once you understand what they mean they’re really useful.

kaste:
Updated the viewport marker code.

kaste:
Hooray

PR #1545, opened 2019-01-30
Pin x-axis when drawing the error panel

kaste:
- If you have a lot of info in the panel, and async updates (bc e.g. flake8 and mypy and annotations linter all really have different runtime characteristics) the error panel basically flickers and jitters. 👎 
- If the widths of the rows change, you easily again loose your x-axis and scroll to the right

For the x-axis, we do pretty much the same as before but here (since the running text command actually replaces the content) we need to defer the pinning to the next task/tick. So we use `set_timeout` and a separate command now. (More exact: We cannot *measure* the new viewport in the same TextCommand as we used `replace`. We need to wait until the function returns!)

For the general flicker/jitter: `on_lint_result` runs on the worker thread but the actual render functions run on the main thread as TextCommands. So I basically lift `fill_panel` execution as a whole to the main thread using `set_timeout`.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -40,7 +40,10 @@ def on_lint_result(buffer_id, **kwargs):
     for window in sublime.windows():
         if buffer_id in buffer_ids_per_window(window):
             if panel_is_active(window):
-                fill_panel(window)
+                # `on_lint_result` runs on the worker thread, we lift
+                # `fill_panel` to the main thread to eliminate redraw jank.
+                # We just *want* this fast and smooth.
+                sublime.set_timeout(lambda: fill_panel(window))
 
             if maybe_toggle_panel_automatically:
                 toggle_panel_if_errors(window, buffer_id)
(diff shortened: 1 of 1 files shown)


kaste:
This one is important! Bc we already shipped #1538 we currently regress here. 

#1538 'respects' the x-axis in `update_selection`, so when Sublime now scrolls to the right bc of `fill_panel` we never scroll back to the end except the user actively scrolls. (Before #1538 every movement scrolled implicitly to the left so we never really noticed. It was just glitchy/janky.)

panel_view.py
```diff
@@ -40,7 +40,10 @@ def on_lint_result(buffer_id, **kwargs):
     for window in sublime.windows():
         if buffer_id in buffer_ids_per_window(window):
             if panel_is_active(window):
-                fill_panel(window)
+                # `on_lint_result` runs on the worker thread, we lift
+                # `fill_panel` to the main thread to eliminate redraw jank.
+                # We just *want* this fast and smooth.
+                sublime.set_timeout(lambda: fill_panel(window))
```
kaste: I'll look into this one later again. This PR should be regarded as a hotfix.

braver:
Ok, let’s roll this one out then 👍🏻

PR #1547, opened 2019-02-01
Start importing from `typing` (and make flake happy)

kaste:


kaste:
Yeah, again semver for the win

kaste:
That's "idiomatic mypy" 😁 I set a magic variable to False so that python33 doesn't execute the block. Only when running MYPY this var is always True (regardless what you set). (There is similar stuff when you use webpack and use dev/prod code switches.)

I would have to do this 

```
MYPY = False
if MYPY:
   ...
```

thing in every file I use types/typing. 

I'm not sure I would want this. I actually added some type information in so called 'type comments' for the human reader only. But now the new flake8 actually reads this stuff and I'm undecided if I should remove my type hints or make them work for mypy.

braver:
Awesomesauce. 

Over the years I have become less and less happy with type annotations. Or perhaps mostly with TypeScript, which tries to specify a language that actively resists such things. In this particular case, knock yourself out 😉

PR #1556, opened 2019-02-13
Showcase: Focus current file in panel

kaste:
As discussed in #1544. Solves #1415. Obviously on top of #1544.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -90,7 +90,7 @@ class UpdateState(sublime_plugin.EventListener):
         })
         ensure_panel(window)
         if panel_is_active(window):
-            update_panel_selection(**State)
+            fill_panel(window)
             start_viewport_poller()
         else:
             stop_viewport_poller()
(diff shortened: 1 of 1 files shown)


kaste:
Well, of course I don't add any setting here. You just switch branches 🙄

kaste:
So take a look at this one @braver. This is still an implementation idea, not even 20 lines of code(!). 

Here we want that the user has clear focus on the current file, and not the other errors. See #1415. Due to the new scroller/viewport stuff in master this is unbelievable fast, yankfree. I mean you ctrl-tab the views, and the panel just updates basically instantly, no scrolling, no animations, no flicker. It is 🤹‍♀️ LOL

What I actually like here, and what is maybe a bit surprising when you only read #1415, is that it only *visually* focuses the current file, t.i. it actually does *NOT* hide the other files. E.g. it does not pretend I have a clean repo, and I can scroll up to see those if I ever want to without flipping flags. It is really just a visual trick, e.g. the panel looks empty/clean for clean files which is 👍 

(Disclaimer: I never had this branch running bc we just merged the other PR's I had the wrap-message PR checked out to see if it has bugs or is otherwise annoying.)

braver:
Coolio! I definitely got some use out of the panel @ master today  and it was really solid. Will check this out next. 🎉

braver:
Woo, this is rly nice m8.

kaste:
Follow up in #1681

PR #1560, opened 2019-02-22
Wrap message in panel

kaste:
Obviously, on top of #1544 

For user ergonomics, if the linter spits out a lot of information, and you're maybe using portrait mode (have narrow Sublime windows side-by-side) the panel gets small but x-axis scrolling is not a good option. Remember the panel never gets focus so you have to use your mouse just to read that message. 👎 

So we have to think about wrapping the message.

Looks like this

image

A bit of number fiddling to get all markers right 😅  

Note that I mark errors near the cursor like so

image

t.i. I only mark the first line of each error. Otherwise it just looks ugly. It becomes like a blue or grey massive block. With this approach I still see that I mark two errors, and can more easily find the line, col etc meta data. 

Overall, although a dense table without wrapping looks a bit better, the scrolling ux *is* really terrible. Most users probably use wide setups, and never encounter this.

### DIFF
--- a/lint/persist.py
+++ b/lint/persist.py
@@ -7,6 +7,26 @@ from .util import printf
 from .settings import Settings
 
 
+if False:
+    from typing import DefaultDict, Dict, List, Type, Optional
+    from mypy_extensions import TypedDict
+    import sublime
+    import subprocess
+    from .linter import Linter
+
+    LintError = TypedDict('LintError', {
+        'line': int,
+        'start': int,
+        'end': int,
+        'region': sublime.Region,
+        'linter': str,
+        'error_type': str,
+        'code': Optional[str],
+        'msg': str,
+        'panel_line': int
+    })
+
+
 api_ready = False
 kill_switch = True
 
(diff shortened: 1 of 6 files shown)


braver:
Nifty! That's a lot nicer than enabling the wrap option for the view.

kaste:
Had to fix a bug in the viewport updater.

kaste:
(Now the viewport marker is brutally fancy, you can just scroll the main view with your mouse, no clicking, and the viewport scrolls with you 🙄 )

braver:
Hmmm, the wrapping should update when the viewport resizes. Also, right now it breaks the highlighting.

screenshot 2019-02-26 at 12 23 42

kaste:
Do you mean the syntax highlighting?

Well yes, I have set a base color for `meta.linter.body` and then it works. Maybe someone comes along and writes a 'good' syntax def. That's a waste of time for me bc I can almost only do random changes and see what these do. I'm just basically clueless how this syntax system works.

I don't listen for viewport resizes. Nobody does that 😁, and if so you can 🤞 just close and reopen the panel. It would be easy enough to implement though.

kaste:
(Aside: For stylelint you should capture the rules (e.g. 'string-quotes') as 'code'. As you can see, code is always 'error', and it's ugly.)

braver:
Well that tells you how often I use the panel 🙄 thank for the tip!

I’ll have a look at the color scheme stuff.

kaste:
Fixed a bug and added support for multiline messages to the syntax.

kaste:
Wow, I totally crashed the eslint issue/discussion at #270

braver:
Pretty nice sir!

Sorry, but I'm still adjusting to the new job and I don't have the mind space left over to interact here in a timely manner.

I love this though, it works really well. I haven't really tried it out in the field because the panel just isn't in my workflow. But from just having a go at it, it works really well. I should try tomorrow with some files with huge amounts of errors, see what that does.

kaste:
I also think it really works well, and if you compare this PR with master, e.g. I had to to #1565 based on master, it is a significant step. I think you never want to go back if you used this for two weeks.

braver:
Let’s just merge it. If we want to fine tune later we can always do it whenever.

kaste:
Hooray!

PR #1565, opened 2019-03-11
Fix `disable_if_not_dependency`

### LINKED ISSUE #1559: linter(erred) shown even though 'disable_if_not_dependency' is True
I'm working on a SublimeLinter plugin: https://github.com/KidkArolis/SublimeLinter-contrib-healthier.

I'm turning on the following default in the next version:

```
defaults = {
  ...
  'disable_if_not_dependency': True
}
```

I think that makes sense for most npm based linters, especially these more niche ones. In any case, right now, in the status bar I'm still getting `healthier(erred)` in every project, even if the project doesn't depend on healthier.

It's because, this line calls notify_failure when no bin is found even if   'disable_if_not_dependency': True is configured.

https://github.com/SublimeLinter/SublimeLinter/blob/1152b89b386e17f3da6d9520392e2370d4bb4b9e/lint/linter.py#L955-L958

I can open a PR that takes into account `disable_if_not_dependency` before calling notify_failure in this case? In the meantime, I'm exploring if I can use a custom cmd to workaround this issue.

kaste:
Fixes #1559

The setting `disable_if_not_dependency` was disfunctional. If a user set it to
`true` the status bar still indicated '(erred)'.

Now, there aren't any good solutions here so we dig a new hole. Generally we
raise during linter execution to rewind the stack. We introduce a new error
`PermanentError` as a compagnon to `TransientError`, and define new side-effects
for the status bar.

- If we receive a `TransientError` we keep the old errors, and the old status
  bar information. To achieve this we re-raise the/an error in the backend
  which transform into a `None` which in turn signals to *NOT* call the
  continuation `sink()`.

- If we receive a `PermanentError` we erase old errors, and remove the linter
  ('unassign') from the status bar. If we don't do the latter we would get an
  '(ok)'. To achieve this we return `[]` in the backend and call the `sink()`
  normally. This will update the error repo and with it e.g. the panel. To not
  get a '(ok)' we introduce a new event `sublime_linter_unassigned` which we
  send to the status bar.

We clearly see that 'active_linters_view' and the backend almost wants to be
coupled, and we don't want that so at least we only send events. Note for the
future: The messages still always flow from backend to view.

Ideally we could just use `can_lint_view` for that feature but it - as a class
method - doesn't have access to the information we need to make the decision.


braver:
Isn't this an opportunity to really work out how this should really be handled? I mean, not like someone else is going to do it for us and we didn't introduce the bug last week right? Apparently not that many are really that bothered by it. 
It just feels wrong to deal with this via exception handling, it should be dealt with initially to figure out if the linter "can_lint_view".
I'm ok with the patch though, that's also ok. But like you say, digging a deeper hole is no fun.

kaste:
Well, as I first saw the issue I thought we're doomed and just have to remove that feature altogether. Compared to that this PR is very 👍 

Just from the name it looks like 'can_lint_view' is a match. But there aren't any just good solutions here. For 'can_lint_view' we actually have the dont-ask-your-mom policy. It still runs single-threaded, and you actually have to decide the question 'yes I can lint this view' only using the arguments `(view, settings)`. 

On the other side, we already throw to abort an ongoing lint process. And if you see `TransientError` then a `PermanentError` looks like a missing piece. Why not have both? 

What I actually don't like is the interaction between the status bar view and the backend. The status bar listens for assigned, failed, unassigned, and I think lint_result events. And I don't know why, I'm a bit surprised here. For example, for the panel and the highlighting (both are complicated views) it would be just enough to return `[]` to clear the errors and so to have a 'clean' view. But for the status bar, this is never the case. It always asks for special handling. As we did the multi files feature, the only view that mucked iirc was the status bar. 

So the only +1 for 'can_lint_view' besides its name is if it returns false we don't 'assign' the linter and then don't have to 'unassign' it. The backend doesn't care, it's merely a decision for the status bar. 

It is still better to have the technical debt *in* the code base though, bc then we can just go through the code and enumerate all the possible side-effects we actually need to get the features we want and try a refactoring based on facts.

kaste:
Contrary to #1572 I still think this is the right approach. Even if we rewrite more parts, `can_lint_view` must be fast, dont-ask-your-mom contract. And we need such a function. 

Say you have ten linter plugins installed, we always need a fast way to reduce the plugin candidates to the usually one or maybe two linters applicable for the current view. It is a hot-path function and I think a fan-out to multiple threads is not an option. The coordination waste is probably higher than what we have right now.

lint/base_linter/node_linter.py
```diff
@@ -51,7 +51,17 @@ def context_sensitive_executable_path(self, cmd):
...
+            logger.info(
+                "Skipping '{}' since it is not installed locally.\n"
+                "You can change this behavior by setting 'disable_if_not_dependency' to 'false'."
+                .format(self.name)
+            )
+            # This looks like a hack (and maybe it is) but at this point in
+            # the execution flow we can only stop and rollback the stack.
+            # If we could use 'can_lint_view' for the check the below would not
+            # be necessary.
+            self.notify_unassign()
```
kaste: Fishy comment alert

kaste:
A git research revealed that `can_lint_view`, `notify_failure` et.al. are refactorings esp. made for the new status bar view #1283. So this is not the typical coupling, it is the first implementation to make the status bar work. (I don't remember how it looked before but the PR enables the '(erred)', '(ok)' stuff.) 

It also explicitly changes the notion of 'deactivated' to 'failed' and that's actually then the origin of the this bug. (A running lint task can only fail and not 'deactivate' the linter.)

braver:
I think before it would flash a message on the statusbar that a linter had died, and simply no persistent info in the statusbar. Remember, we didn't have info on which linter reported what, it was just the total number of errors/warnings (we still have that in the screenshot in the README... I should update that). So if a linter didn't run, you basically didn't hear about it.

So that's perhaps also why we didn't carefully consider the difference between deactivated or failed, there are more subtle distinctions we added later.

The statusbar is different from the other views because it matters why there are no results. You should be able to give it `[]` to clear the numbers, but it also needs to know which linters are enabled on the view and what their status is (disabled, erred). So I'm not really surprised there is tighter coupling with the backend here (not a value judgement, perhaps we shouldn't have what we have now). 

That said, I am in general in favor of fixing now and refactoring later. Perhaps we should just merge and release this PR, and then if you want we can discuss what to do with the status bar states and interface?

kaste:
Ah well, it all makes more sense with the git history. Let alone we added this for v4.3, I thought this was in the initial 4 release. For a subrelease it is of course more natural to have some oddities in the implementation. 

That's really easy fixable if we had the time to think and enumerate about all possible outcomes for a lint task.

Issue #1564, opened 2019-03-04
Ensure 'Linter' instances are used once even if a view lints multiple partials of the buffer


kaste:
People expect objects like the 'self' they see to be mutable. This is not safe. 

Last year I took care that we never mutate the 'self' state to be safe for multi-threading. But that's not realisitic bc users don't expect that. They just want their mutability hell, and use 'self' as storage as they need. 

We currently only reuse linter instances if the linter selects multiple partials of a buffer. And so this operation mode is not safe and can produce random errors. After mheinzler PR (which *I* merged) we have code in master which mutates 'self' so master is essentially not thread safe.

Somehow we must 'clone' the linters in the backend so that 'self', 'context' and 'settings' are independent and don't leak information to other running instances.

PR #1566, opened 2019-03-13
Prepare release message for 4.10.0

kaste:
Just start with compiling what changed.


braver:
Good idea! LGTM 👍🏻🚀

kaste:
We have a milestone 4.10.0 now bc there is a bug in master (#1564) (Of course, it's around an edge case, and probably nobody would ever notice 🤷‍♂️)

braver:
👍

kaste:
We must add 

- fixed #1565  
- added #1576

PR #1568, opened 2019-03-14
Export LintMatch

### LINKED ISSUE #1563: Expose 'LintMatch' as a toplevel export
We encourage users to use and return 'LintMatch' instances instead of 7-tuples so we need to proper export this class.

Consider though renaming it to 'LintError' first.

kaste:
Fixes #1563

### DIFF
--- a/lint/__init__.py
+++ b/lint/__init__.py
@@ -12,7 +12,7 @@ from . import (
 from .const import WARNING, ERROR
 from .util import STREAM_STDOUT, STREAM_STDERR, STREAM_BOTH
 
-from .linter import Linter
+from .linter import Linter, LintMatch
 from .base_linter.python_linter import PythonLinter
 from .base_linter.ruby_linter import RubyLinter
 from .base_linter.node_linter import NodeLinter


kaste:
🙄 I did not rename it

kaste:
Ah well, `LintMatch` or `LintError` which can also be a warning or note. `LintError` is like we have the severity in the name. `LintMatch` is like it's only useful for regex matchers but actually it is extremely useful for JSON reporters. 🆘

kaste:
In Atom it's called `Message`, and what we call 'message' they call 'description'. In VSCode it's a list of `Diagnostic` (items?) LOL.

braver:
Cache invalidation, off by one and naming things...

braver:
🚤 it because 🤷🏼‍♂️❓

kaste:
Obviously, we go with `LintFact` :grin:

braver:
Obviously! 

LintThing? LintReport?

kaste:
`yield Fact`

braver:
I like that actually

braver:
It’s too nice really

PR #1571, opened 2019-03-20
Rename arg `syntax` -> `settings`

kaste:
We actually change the `__init__` signature in #1435 and forgot to update
our subclassed linters. 🤷‍♂️


braver:
👍🏻👍🏻

PR #1572, opened 2019-03-20
Ensure independent, mutable linter instances in backend

### LINKED ISSUE #1564: Ensure 'Linter' instances are used once even if a view lints multiple partials of the buffer

kaste:
Fixes #1564

People are used to the OOP rule that if they see a `self` they can mutate
at their will. This works okay if the linter selects the whole buffer because
then the linter is used only once. If the linter selects multiple regions
instead we always reused the linter which can lead to subtle bugs.

We therefore clone the linter if the linter selects multiple regions.

We ensure that cloned linters inherit all modifications made to the settings
from the 'parent'. E.g. a linter could modify the settings in `can_lint_view`,
and these changes must be visible to the cloned linters.

All changes after the cloning are local.

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -54,8 +54,17 @@ def run_tasks(tasks, next):
 
 
 def get_lint_tasks(linters, view, view_has_changed):
+    total_tasks = 0
     for (linter, regions) in get_lint_regions(linters, view):
-        yield linter, _make_tasks(linter, regions, view, view_has_changed)
+        tasks = _make_tasks(linter, regions, view, view_has_changed)
+        total_tasks += len(tasks)
+        yield linter, tasks
+
+    if total_tasks > 4:
+        logger.warning(
+            "'{}' puts in total {}(!) tasks on the queue."
+            .format(short_canonical_filename(view), total_tasks)
+        )
 
 
 def _make_tasks(linter_, regions, view, view_has_changed):
@@ -73,6 +82,11 @@ def _make_tasks(linter_, regions, view, view_has_changed):
         executor = partial(modify_thread_name, task_name, task)
         tasks.append(executor)
 
+    if len(tasks) > 3:
+        logger.warning(
+            "'{}' puts {} {} tasks on the queue."
+            .format(short_canonical_filename(view), len(tasks), linter_.name)
+        )
     return tasks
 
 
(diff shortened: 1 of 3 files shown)


kaste:
Okay, this is not sufficient so we cannot pull. We also have `should_lint` which could modify `self` and which would not be protected. 

The code to clone settings is still okay so I don't close this immediately. (Why do we have two phases here, `can_lint_view` and `should_lint`? That's again code for the status bar!) 

1. I will see why `should_lint` is a method (users can override) and not a plain function.
2. Maybe we can move the call to `should_lint`?

Honestly it feels a bit like end-of-lifecycle for this code. (Interaction between `sublime_linter.py` and `backend` and esp. the status bar view.)

kaste:
Okay, I looked into it. This fix solves a real, exploitable edge case bug and trades that by (possibly) breaking users which implemented `should_lint` and also mutated 'self' in it.

- but probably nobody did that anyway?
- I deprecated `should_lint` here immediately. We can get the functionality in a plain function, or a classmethod (which has no 'self' to mutate), we will see.
- An overall refactor is not an option here bc there is just no time for it. As I see it you have to do some refactoring steps, to get to the point where you can touch and modify 'should_lint'. It is clearly not the first step here. That's life. 🤷‍♂️ 'cloning' linters is of course not what we want. It is also not necessary if we refactor but ⌚️ 🤷‍♂️ 

I think the meat of this PR is now deprecation warnings and other logs.

kaste:
Actually, regarding the refactoring, it's not only ⏲ but we already have 100 commits since the last release, it just feels so risky to put another 30 on top of that.

braver:
> but probably nobody did that anyway

I did a search across all linters, nobody overrides `should_lint`.


> That's again code for the status bar!

Which again doesn't surprise me, for the status bar it matters if a linter is enabled for a view, but maybe shouldn't do anything yet (because it's run manually for instance). It does seem to complicate matters a lot though.


So, how do I test something here, what is the expected different behavior and can I observe that, or is it enough that we have tests and now some extra loggin?

kaste:
These are all edge cases. I actually hacked/broke a linter to exploit the bug and to confirm that this PR fixes the issue.

But you actually only need a tempfile linter, and a file which contains multiple lintable regions. It's just that I didn't find any tempfile linters on my disk. Without this PR such a linter will err and probably also show wrong highlights (race condition).

braver:
Well, I can't really get anything like that set up to run quickly. The only tempfile linters I'm familiar with (pylint, tslint) don't really do embedded code at all it seems. They make a completely unusable mess of things. And I don't run into use cases with multiple lintable regions otherwise. I'll take your word for it though.

Wanna merge this and complete 4.10?

PR #1573, opened 2019-03-20
 Make `LintMatch.match` optional

### LINKED ISSUE #1562: Make 'match' in 'LintMatch' optional
The standard contract here should be plugins return 'None' to throw/skip the current error, or a 'LintMatch()'. 'LintMatch.match' holds the original regex 'Match' object but 'SublimeLinter' itself doesn't use it. 

There is also a funny duplication

```
def split_match(self, match: Match) -> LintMatch:
	error = super().split_match(match)
	# Now this holds
	assert error.match == match
```

This probably never made much sense. 

However, in the end 'LintMatch.match' must be truthy otherwise we discard the error later. Esp. if users do JSON processing they should be free to pass around a detailed error object or none if they don't use it in their *own* code.

kaste:
Fixes #1562

We do not use `LintMatch.match` internally, and for JSON reporters it is
annoying to pass a truthy value here just to get the error passed.

We implement here easy as **breaking change** because it is unlikely users
explicitly set the first item of our 7-tuple to `None` but have valid values
for the other holes.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1035,7 +1035,7 @@ class Linter(metaclass=LinterMeta):
                 self.name, textwrap.indent(output.strip(), '  ')))
 
         for m in self.find_errors(output):
-            if not m or not m[0]:
+            if not m:
                 continue
 
             if not isinstance(m, LintMatch):  # ensure right type
--- a/tests/test_regex_parsing.py
+++ b/tests/test_regex_parsing.py
@@ -1045,24 +1045,6 @@ class TestSplitMatchContract(_BaseTestCase):
 
         self.assertResult([], result)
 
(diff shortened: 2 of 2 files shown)


kaste:
Hm, the overall diff is unreasonable. If you check the diff of the first commit it becomes clear that this is a breaking change bc we remove a test. The "implementation" is 4 chars approx 😄

braver:
If we don't use it, why is it part of that tuple in the first place?

Anyway, 🚢 it I guess.

kaste:
Just a historical mistake, wasn't used in SL3 as well, and since changing a tuple structure is always a breaking change you never could fix that.

PR #1574, opened 2019-03-20
Turn off codecov comments

kaste:
😶


kaste:
So insanely annoying that they still do the double comments after weeks/months. Fix your fucking bugs.

Issue #1323, opened 2018-04-17
"toggle highlights" off by default


krishnaghatti:
Is there a way to keep the highlights toggle to off by default?
I have to go to the command pallet and choose the "toggle highlights" for each file opened.

<img width="1014" alt="screen shot 2018-04-18 at 1 42 09 am" src="https://user-images.githubusercontent.com/10704122/38894113-f99ca41c-42a9-11e8-8c6f-02b156e75385.png">

braver:
Do you want to have them “off” permanently, or just by default and them switch them on when needed?

krishnaghatti:
I want them to be off by default, but should be able to enable when required.

middlesister:
+1

It would be great if this could be a project/user specific setting.

kaste:
I also would like to see/test that. (I wrote the current toggler here, and it just came naturally from the code and data model we had there. I don't think a global toggler is completely naive, trivial code. But then, the code changed since then anyways. 🤷‍♀️ )

kaste:
Okay, after looking at it it seems relatively trivial to add.

I cannot do a project wide setting bc we don't have such a layer (yet?). So it would be global switch. We also do not persist settings, so it would be a default after starting Sublime. Like `highlights.on_after_start: true|false`.

I also don't know if we need both, a toggler for views and a global toggler for the whole app. 🤷‍♂️

kaste:
Something very similar has been implemented as a plugin https://github.com/kaste/SublimeLinter-addon-filter

Look esp. for the on/off switch at the bottom. Install via Package Control; search for SublimeLinter-addon-filter

krishnaghatti:
@kaste Thank you for the update. I was using the plugin for a few days now. Just to clarify, the "off" part of the toggle will switch the linter off completely right?

kaste:
The linters actually run in the background but don't produce any visual outcome when off. This is similar to the toggle highlights command and has the advantage of being very fast or immediate.

If you want to actually toggle linters, there is another plugin for that as well. Search for addon-toggler or something like that in Package Control.

AndrewMast:
I created a pull request to allow this (#1576).

PR #1576, opened 2019-03-24
Highlights can now start as hidden

AndrewMast:
Fixes SublimeLinter/SublimeLinter#1323

Changes:
- Adds a "highlights.start_hidden" setting (boolean)
- When lint results come in, if the "highlights.start_hidden" is true, the views are added to the "quiet views state"

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -41,7 +41,7 @@
     "highlights.time_to_idle": 1.5,
 
     // Set to true to start the highlights hidden. Use the command
-    // "sublime_linter_toggle_highlights" to toggle the highlights
+    // "SublimeLinter: Toggle Highlights" to toggle the highlights
     "highlights.start_hidden": false,
 
     // Send a "terminate" signal to old lint processes, if their result would
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -40,6 +40,10 @@
     // you save the buffer.
     "highlights.time_to_idle": 1.5,
 
+    // Set to true to start the highlights hidden. Use the command
+    // "sublime_linter_toggle_highlights" to toggle the highlights
+    "highlights.start_hidden": false,
+
     // Send a "terminate" signal to old lint processes, if their result would
     // be thrown away. If false we fire-and-forget processes instead.
     "kill_old_processes": false,
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -54,7 +54,8 @@ State = {
     'active_view': None,
     'current_sel': tuple(),
     'idle_views': set(),
-    'quiet_views': set()
+    'quiet_views': set(),
+    'views': set()
 }
 
 
@@ -105,8 +106,11 @@ def on_lint_result(buffer_id, linter_name, **kwargs):
     for view in views:
         vid = view.id()
 
-        if persist.settings.get('highlights.start_hidden') and vid not in State['quiet_views']:
+        if persist.settings.get('highlights.start_hidden') and vid not in State['quiet_views'] and vid not in State['views']:
             State['quiet_views'].add(vid)
+
+        if vid not in State['views']:
+            State['views'].add(vid)
 
         draw(
             view,
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -103,14 +103,19 @@ def on_lint_result(buffer_id, linter_name, **kwargs):
         view, linter_name, errors_for_the_gutter)
 
     for view in views:
+        vid = view.id()
+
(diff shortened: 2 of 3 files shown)


kaste:
Hi @AndrewMast, thanks for contributing! 

🤔 The current code you provide will put the views into quiet state every time new results fly in. (E.g. you look at a view with highlights, then continue hacking, the highlights will disappear (if you're linting in background mode) until you invoke the highlights toggle command again.)

I think iirc the issue #1323 is just about the initial state of a view. Right now, each view starts 'loud', and then you toggle 'quiet..loud..quiet..loud', and the question was if we can make it so that each view starts 'quiet'.

AndrewMast:
@kaste I just woke up realizing that this bug would happen... Lemme go try to fix it.

AndrewMast:
@kaste There, that should fix it.

kaste:
👌  That's a great, simple solution! Now, 'views' is ever growing so we must keep that in mind now or for later.

AndrewMast:
Yeah, I was hesitant to create 'views', but I couldn't think of any other way of seeing if it is a new lint view or just a re-linting of an old view.

AndrewMast:
Should I add anything about the 'views' set to highlight_view.py's `plugin_loaded `or `plugin_unloaded` functions?

kaste:
Hm, I think

```
class GarbageController(sublime_plugin.EventListener):
    def on_pre_close(self, view):
        vid = view.buffer_id()
        State['idle_views'].discard(vid)
        State['quiet_views'].discard(vid)
        State['views'].discard(vid)
```

does the trick?

AndrewMast:
Looks good. Want me to add it in?

kaste:
But when we have a controller, maybe we can also just do

```
    def on_new_async(self, view):
        if persist.settings.get('highlights.start_hidden'):
            State['quiet_views'].add(vid)
```

Then we don't need the 'views' state. But we're also in API land so I don't know if Sublime calls us here exactly how we expect :grin:

Yes sure it's `view.id()`.

Actually, the whole approach here is so super easy. I always thought solving the issue would take way more time.

AndrewMast:
Okay, added the Garbage controller. I can't think of anything else to add regarding the hiding of views.

kaste:
The lazy @braver should look if he likes the name of the option, and if we actually implemented the right feature.

braver:
Hey, I'm not lazy! Not really. And if I were, that's not a bad thing. 😇

I'll check this out this afternoon so we can merge and then hopefully move towards a release today.

SublimeLinter.sublime-settings
```diff
@@ -40,6 +40,10 @@
     // you save the buffer.
     "highlights.time_to_idle": 1.5,
 
+    // Set to true to start the highlights hidden. Use the command
+    // "sublime_linter_toggle_highlights" to toggle the highlights
```
braver: `sublime_linter_toggle_highlights` is indeed the name of the command, but we should assume users will call the command via the command palette, and therefor refer to it via the name we gave it there.

braver:
This is excellent. 🏅 I always thought it made sense to have this, so really glad you decided to contribute this.

highlight_view.py
```diff
@@ -103,17 +104,37 @@ def on_lint_result(buffer_id, linter_name, **kwargs):
...
             gutter_regions,
             protected_regions,
-            idle=(view.id() in State['idle_views']),
-            quiet=(view.id() in State['quiet_views'])
+            idle=(vid in State['idle_views']),
+            quiet=(vid in State['quiet_views'])
         )
 
 
+class GarbageController(sublime_plugin.EventListener):
```
braver: I'm not sure I'd call it GarbageController, garbage is a loaded term.. To really complain about it I should come up with a different term. `ViewListCleanupController`?

SublimeLinter.sublime-settings
AndrewMast: So "SublimeLinter: Toggle Highlights"?

highlight_view.py
kaste: When @braver complains about such names it's usually a joke. 🤔

SublimeLinter.sublime-settings
AndrewMast: I went ahead and changed it to that. Luckily the line is the same length after the edits 😉

AndrewMast:
@braver Both of the changes have been made. Anything else I should do?

SublimeLinter.sublime-settings
braver: I guess without the SublimeLinter prefix would also work but 👍🏻

braver:
Nope, LGTM, thanks!

Issue #1447, opened 2018-06-08
Nice error message when package.json cannot be loaded


kaste:
As seen in #554 

If the NodeLinter cannot decode `package.json` we just throw at the user. Instead we should show a sane, actionable error message to the user.

kaste:
Same for `composer.json`. See https://github.com/SublimeLinter/SublimeLinter-phpcs/issues/44

PR #1577, opened 2019-03-26
Move common manifest-related features to a standalone mixin

AndrewMast:
This would enable us to create fancy errors for the failing of reading the manifest files (#1447). I plan on making those errors in this PR, but I just want to make sure I am going in the right direction.

Mind taking a look at this, @kaste or @braver?

(Edit: Also, kaste, you should join Sublime's discord server so we potentially don't have to spam people with notifications)

### DIFF
--- a/lint/base_linter/composer_linter.py
+++ b/lint/base_linter/composer_linter.py
@@ -1,24 +1,22 @@
 """This module exports the ComposerLinter subclass of Linter."""
 
-import codecs
-import json
-import hashlib
 import logging
-import os
-import shutil
 
-from .. import linter, util
+from ..linter import Linter
+
+from .manifest_linter_mixin import ManifestLinterMixin
 
 
 logger = logging.getLogger(__name__)
 
 
-class ComposerLinter(linter.Linter):
+class ComposerLinter(ManifestLinterMixin, Linter):
     """
     This Linter subclass provides composer-specific functionality.
 
     Linters installed with composer should inherit from this class.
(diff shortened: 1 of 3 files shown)


kaste:
Hi again!

These base classes must be rewritten. For NodeLinter, we have some code here https://github.com/SublimeLinter/SublimeLinter-eslint/issues/270#issuecomment-470481041 and https://github.com/SublimeLinter/SublimeLinter-eslint/issues/270#issuecomment-471006154

We wouldn't go the mixin way, because plain functions give us better reusability and composability.

AndrewMast:
So scrap the mixin idea? How would you propose to do it? All of the functions used in the NodeLinter and ComposerLinter were exactly the same except for the 'disable_if_not_dependency' in the NodeLinter...

AndrewMast:
Should I make it just ManifestLinter and have Composer and Node extend it? It does the same purpose but I originally didn't do it because I didn't want people trying to extend the ManifestLinter....

kaste:
I'm not prepared to answer this question right now. I'll come back tomorrow. (We usually also get bug reports tomorrow because we shipped 100+ commits just now.) 

I think there are probably more interesting things for you than copy/pasting my code from the comments I linked to above.

AndrewMast:
I’ve decided to close this.

kaste:
You can take a look at the new proposed NodeLinter in #1579. Sorry that I interrupted you here but I really had this worked out before in the eslint thread I linked above.

PR #1579, opened 2019-03-28
Rewrite `NodeLinter`

### LINKED ISSUE #1447: Nice error message when package.json cannot be loaded

kaste:
Fixes #1447 

The current `NodeLinter` has a couple of problems:

- It does 'package.json' lookup work at `__init__` time which slows down linter election on our main worker thread
- It implements 'caching' on the instance which is useless bc we basically never reuse an instance
- It basically obfuscates the actual lookup routine. (It is completely non-reasonable, you can sit there for hours deciphering.)
- It 'solves' an edge case where a linter lints its own code while hacking on it. Unfortunately, the implementation doesn't work. 
- It raised if a package.json file couldn't be read. #1447 (Often it is not even necessary to really read that file but here we always 💥 'ed.)

So a rewrite is necessary:

- We lazy lookup in `context_sensitive_executable_path`.
- The happy path is a locally installed linter, in which case we don't traverse or read any 'package.json' file at all.
- The json cache is global so that multiple views, and multiple linters participate here.
- Notable: checks if a project is not installed yet, and does not use the global linter if so.
- Tests 🔥


braver:
I’ve been using this for a few days now. No hiccups, looks good!

kaste:
@braver You forgot to release this. Technically, it's a 4.11.0 if that even makes sense for apps.

braver:
👌🏻

PR #1590, opened 2019-04-08
Normalize temp_file path to fix mismatch error with some linters. Fixes#1589

JoyceBabu:


JoyceBabu:
I have fixed it and force pushed.

kaste:
The change is 👌, but I'm hesitant. I would now prefer to solve it in the phpmd plugin first. It is basically a phpmd bug bc they do the unusual `realpath` https://github.com/pdepend/pdepend/blob/master/src/main/php/PDepend/Engine.php#L235 and if we take this PR here, phpmd would still have the same bug if we exchange temp file handling with 'real' file handling using `${file}` or `$file_path`. 

Now, the whole filename feature has just shipped so I would like to wait here if we get more bugs/feedback, and of course if multiple linter plugins need some 'realpath' handling we can reconsider this one here bc as I said it's 👌 I'm also extra sorry bc you already had another PR iirc hanging. 

If we were to fix it in the plugin, we just don't capture the 'filename' as a named regex group.

I would like to hear @braver here.

JoyceBabu:
> if we take this PR here, phpmd would still have the same bug if we exchange temp file handling with 'real' file handling using ${file} or $file_path

It does work correctly with `${file}`. What is failing is the check whether the error was reported file on the temporary file. When `${file}` is used, there is no filename comparison, hence it works without any error.
https://github.com/SublimeLinter/SublimeLinter/blob/master/lint/linter.py#L1247

I think passing the absolute real path, is the safe thing to do. Since `/private/var/tmp` is the canonical path, there is no reason for a plugin to return `/var/tmp` or even a relative path in the output.

> I'm also extra sorry bc you already had another PR iirc hanging.

No worries. Ultimately, what is good for the project is good for me too :)

braver:
To me the change also looks 👌 I don't know all the context, but normalizing to absolute real paths is never a bad idea. As you say:

> Since /private/var/tmp is the canonical path, there is no reason for a plugin to return /var/tmp or even a relative path in the output.

Your change looks like "why wasn't it like that before" to me.

I can get behind @kaste's idea of letting the current version of SL simmer a few more days. So if you can do a quick work around in the phpmd plugin that might be nice and quick, and we can merge this PR in a few weeks. Personally I'm totally ok with just merging and shipping, but I have to add that usually I'm not the one who has to do the fixing. Also, it's not like this broke yesterday, and you've got the fix you need now, so everyone else can perhaps wait a little while.

kaste:
Well, `realpath` is actually almost always wrong. What the 'real' path is, is usually an implementation detail, other say these links are 'transparent'. 

If your working on '/a/foo/pa.py', you sure want reports for that exact filename. More importantly, in UI's like Sublime the reported file '/b/baz/pa.py' must match the open views otherwise you open the same file under different names multiple times. 

To put it different, there is no inverse function for 'realpath'. 

I still doubt `$file` would work here bc of later checks. Unless Sublime keeps a reverse/inverse map from realpath to userpath. E.g. Sublime's `window.find_open_file(filename)` must/should return the original view we started with.

braver:
Good points. I guess a real path may be ok in a terminal context, or internally, but when feeding back to the user space you definitely need the original path as the application knows it.
Sublime doesn’t keep track of the real paths, so we can’t  couple the feedback to a view.

JoyceBabu:
But we are referring to a temporary file that was created just for the purpose of linting the unsaved buffer. Sublime is oblivious of this file, only the linter plugin knows about it.

We need the file content to process the match. But the file no longer exists by the time `process_match` runs. So it is important that we detect the filename as the temporary file.

Using `${file}` works because the file is still accessible by the returned path, absolute path or not.

kaste:
@JoyceBabu For the temp file it sure works, but it doesn't gain anything bc we could also just not report and/or capture the filename in phpmd. (We only get a problem if phpmd starts to report errors from multiple files.)

For `$file`, pretty sure we do the `window.find_open_file(filename)` check later in `sublime_linter.py`. 

But overall, the filename feature is pretty young so we first wait which issues arise, and then see how we improve SL or individual linter plugins.

kaste:
We didn't get any further reports. We also completely removed all `real_path` calls in the code base. So, I'm closing this one.

Issue #1583, opened 2019-03-28
Remove deprecated features (first candidates)


kaste:
- Remove the old 'cls.syntax' handling.
- Remove special '@python' handling in `cmd`
- ~~If possible remove deprecations in `finalize_cmd` bc we have plugins hacking around here to prevent auto appending.~~
- ~~Log as error in `on_stderr`~~

braver:
Wanna do this as a 4.x, or move to 5?

kaste:
4.x

I think bigger plugins like SL or GitSavvy are apps, and so a v5 must have big UI/UX impact. Technically, we shipped a lot of gradual breaking changes. The latest 4.10 is technically a 5.0. Other than that, I don't care, the semantic versioning here is fake bc users have the latest version installed automatically anyway. 

As a last step before removing, we could also show an error message in-ya-face. All three things above are trivial to update in the plugins, like 5 minutes work.

braver:
Oh, totally agree, just to consider if we're breaking an API. But we can break API's no-one uses because we have access to all users of the API. And those we can't write to probably won't update until we really break them because maintainers are lazy (I should know ;)).

So, I'm all for this, just push ahead and drop the old stuff. 👍

kaste:
Also: in `on_stderr` log as error, we announced this change long ago.

kaste:
We actually only shipped two of the bullets in 4.12.0 🤷‍♀

PR #1591, opened 2019-04-13
Various small improvements in linter.py

### LINKED ISSUE #1497: Reminder: Defining `regex` should be optional for a linter bc maybe it uses json
Need to conform this first, so this is a reminder, ... but I think classes without `regex` member do not register but only log a warning.

### LINKED ISSUE #1578: Disable and warn if `defaults.selector` is not set
In `LinterMeta` warn if 'selector' is not set because it is mandatory. Also add a test for this if possible. 

Note that 'defaults' in `attrs` for that reason is also mandatory.

EDIT: Actually also disable the plugin.

### LINKED ISSUE #1585: Ensure `parse_output_via_regex` does not yield `None`'s
After a recent mod, `process_match` can return `None` to skip errors. These must be filtered asap to not leak into and break user land code.

kaste:
- Remove deprecated `cls.selectors` and `cls.syntax` handling. Rel #1583 
- Remove '@python' special handling

- FEAT: Automatically set 'cls.multiline=True' if the pattern sets the flag. 

Fixes #1497 
Fixes #1578 
Fixes #1585

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -10,7 +10,7 @@ import multiprocessing
 import os
 import threading
 
-from . import style, util, linter as linter_module
+from . import style, linter as linter_module
 
 
 if False:
@@ -25,7 +25,6 @@ if False:
 
 logger = logging.getLogger(__name__)
 
-WILDCARD_SYNTAX = '*'
 MAX_CONCURRENT_TASKS = multiprocessing.cpu_count() or 1
 orchestrator = ThreadPoolExecutor(max_workers=MAX_CONCURRENT_TASKS)
 executor = ThreadPoolExecutor(max_workers=MAX_CONCURRENT_TASKS)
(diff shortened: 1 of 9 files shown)


braver:
@kaste can you rebase this on master so I can run it for a couple of days?

kaste:
Rebased.

braver:
Looks good to me I guess

kaste:
Since this is def breaking I'll merge a bit later. For 4.12.0 we should have a stable `should_lint` again, and I need #1595 for that.

braver:
👌

kaste:
I merged this bc #1613 is on top of it, and it's too long already.

braver:
Sure. It sat too long anyway. I’ll update my installs so I can test run for a while before we release what’s in master (if we even want to release that before merging more stuff).

kaste:
No release, I've added this to a milestone bc I knew the refactorings will get rather long. But hey we never refactored before.

braver:
Oh, sweet, I missed the milestone. 👌

PR #1593, opened 2019-04-13
Efficient use of `ThreadPoolExecutor`

kaste:
The prior implementation did not reuse the `executor` which is a waste. To
actually limit the number of threads, we additionally had to use a
`BoundedSemaphore`.

We replace this system with two global executors. We need *two* executors
because we still have a nested structure here. The orchestrator does the
fan-out, fan-in dance for each linter in case the linter has multiple regions
to lint. The executor just runs the long lint tasks.

### DIFF
--- a/lint/backend.py
+++ b/lint/backend.py
@@ -1,7 +1,6 @@
 import sublime
 
 from concurrent.futures import ThreadPoolExecutor, wait, FIRST_EXCEPTION
-from contextlib import contextmanager
 from itertools import chain, count
 from functools import partial
 import hashlib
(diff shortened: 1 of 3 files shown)


braver:
Looks good... do you want me to test for anything in particular?

kaste:
I think this is straight forward. I tested with a bloated setup, spawning ~30 tasks per view. With the multiple threads stuff waiting for each other, you have the risk to deadlock, so I go the easy route for now and use two executors. 

Overall, since I refactor my own code here, I actually didn't know how the Executors and Futures work in python during the first implementation. E.g. I thought if I submit a task, it doesn't actually run before I await it. And I thought, executors can be used exactly once, you fill in the unit of work via submit, then await the tasks, and then it's ready for garbage.  But that's all shit: If you submit a task, the task runs immediately, and one executor can run for the whole app life. 🤷‍♀️

braver:
Ok, nifty. I’d say merge as soon as you feel comfortable with the stability of master.

PR #1597, opened 2019-04-24
Fix Sublime's `expand_variables` for UNC paths

kaste:
Fixes https://github.com/SublimeLinter/SublimeLinter-eslint/issues/276

Explainer and origin: https://github.com/SublimeTextIssues/Core/issues/1878

Sublime's `expand_variables` handles `\\` (two slashes) and reduces them to one. We fix by adding more slashes before substituting.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -349,6 +349,9 @@ def substitute_variables(variables, value):
     # `${varname}` syntax and supports placeholders (`${varname:placeholder}`).
 
     if isinstance(value, str):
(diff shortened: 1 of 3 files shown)


braver:
Wow, why is "Textmate compatibility" still a thing? 🚢 it, it's not like I can test this right?

kaste:
The current PR's are all hard to test, I also have to mock my env, run fake linter plugins. It really pays off that we have tests now. We had basically zero issues with our latest releases.

PR #1600, opened 2019-05-02
Escape '$' chars in arg values while computing the `cmd`

### LINKED ISSUE #1599: Filenames with $ signs mangled when using ESLint
Somewhat recently, SublimeLinter started failing in an interesting way when scanning my files with paths that contain `$` signs using ESLint (not sure if other linters are affected too).  For example, when scanning the file `C:\Code\workspace\reviewable\client\src\model\reviews\files\reviews.$reviewKey.files.$fileKey.js`, which has some lint errors, SublimeLinter fails with the following:

```
[{'errorCount': 4,
  'filePath': 'C:\\Code\\workspace\\reviewable\\client\\src\\model\\reviews\\files\\reviews..files..js',
  'fixableErrorCount': 1,
  'fixableWarningCount': 0,
...
SublimeLinter: #323 linter.py:1177    WARNING: eslint reported errors coming from 'C:\Code\workspace\reviewable\client\src\model\reviews\files\reviews..files..js'. However, reading that file raised:
  [Errno 2] No such file or directory: 'C:\\Code\\workspace\\reviewable\\client\\src\\model\\reviews\\files\\reviews..files..js'.
```

Running ESLint from the command line with `--format json` outputs the correct filename, so I'm guessing the problem is somewhere on SublimeLinter's side.  Looks like it's accidentally interpolating file paths and treating `$foo` as variable names.

Versions used:
SublimeLinter v4.11.1
SublimeLinter-eslint v4.1.6
Sublime Text v3.2.1
ESLint v5.16.0
Windows 10

kaste:
Double slashes (`\\`) and of course `$` signs are special characters for
Sublime's expand-variables-system. Here we escape '$' signs before putting
them on the `cmd` since we expand variables on the whole `cmd` again
(currently implemented in `finalize_cmd`).

We don't patch `substitute_variables` but on the fly while creating the `cmd`.
Otherwise, e.g. `executable` had such a special char, all user code operating
on settings would have to deal with these escape sequences. E.g. a simple

```
os.path.exists(self.settings['executable'])
```

would not be safe anymore. Thus, after all, the var substitution machinery
would'nt be transparent anymore.

Fixes #1599

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -865,13 +865,19 @@ class Linter(metaclass=LinterMeta):
 
             joiner = arg_info['joiner']
             for value in values:
+                # We call `substitute_variables` in `finalize_cmd` on the whole
+                # command. Since all settings are already transparently
+                # 'expanded' we need to make sure to escape remaining '$' chars
+                # in the arg value here which otherwise denote variables within
+                # Sublime.
+                final_value = str(value).replace('$', r'\$')
                 if prefix == '@':
-                    args.append(str(value))
+                    args.append(final_value)
                 elif joiner == '=':
-                    args.append('{}={}'.format(arg, value))
+                    args.append('{}={}'.format(arg, final_value))
                 else:  # joiner == ':' or ''
                     args.append(arg)
-                    args.append(str(value))
+                    args.append(final_value)
 
         return args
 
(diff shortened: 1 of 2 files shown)


braver:
Makes a lot of sense 👍

kaste:
Then I just merge and release?

braver:
I usually like to just a cursory check that you didn’t break anything I care about 😉 Super busy day today though. Anyway, you can merge this and I’ll double check and release tomorrow.

PR #1609, opened 2019-06-11
Hard-wrap tooltip contents

### LINKED ISSUE #1601: max height of tooltip is really small
I remember historically sublimelinter never did this, I think this was introduced in a major version ages ago

Essentially part of my workflow is to screenshot the linting error for code review purposes
but the max height cuts of the message and gives a very small window for showing lint errors, is it possible to configure the max width and height via settings? or unlocked as default

Screenshot 2019-05-09 at 10 26 23

kaste:
Fixes #1601
Work-around https://github.com/SublimeTextIssues/Core/issues/2854

Sublime actually wraps long lines automatically but then calculates the height
of the tooltip wrong. We work-around by hard-wrapping all content.

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -3,6 +3,7 @@ import html
 from itertools import chain
 from functools import partial
 import re
+import textwrap
 
 import sublime
 import sublime_plugin
@@ -660,7 +661,7 @@ TOOLTIP_STYLES = '''
 TOOLTIP_TEMPLATE = '''
     <body id="sublimelinter-tooltip">
         <style>{stylesheet}</style>
-        <div>{message}</div>
+        <div>{content}</div>
     </body>
 '''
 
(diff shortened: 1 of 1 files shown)


braver:
May need to subtract a bit from the width:
Screen Shot 2019-06-11 at 17 07 07


I'm also not 100% sure I agree with the dynamic approach here. The tooltip doesn't have to stay within the window, so its max-width is not related to the width of the window. Also, your cursor doesn't have to be at x = zero, so the allowable width is different.

Another thing is that the calculated width based on the em-width of the view is not bullet proof. It is possible for the tooltip to have a different font for instance. 

So I'd rather do the low tech hard wrap at a common standard line length like 80 chars or 120 chars. We do a basic pre-processing and let Sublime do the hard stuff.

kaste:
I'm not happy with this one either. Sublime has a dynamic approach here which is hard to recreate.

braver:
How about this? I kinda like it this way. It prevents the messages getting overly long, which is something I run into a lot actually. I don't often get the bug in #1601, but the wrapping is nice anyway. 80 is kind of a magic number, but in this context it's not a weird number to pick IMO.

Screen Shot 2019-06-21 at 21 11 19
 
Seems pretty solid so far.

kaste:
Did you test this? On Windows at least, the tooltip is now maybe ~300px wide without the `max_width` setting. 

Generally, 80 chars is probably okay for readability although the tooltip looks rather small.

braver:
I think I tested it, may not have been super thorough... I’ll check again

kaste:
image

braver:
Oh yeah, you're right. Weird. Fixed now though. Let's merge it like this?

kaste:
Merged manually

braver:
👍🏻

PR #1610, opened 2019-06-11
Avoid resolving via `realpath`

### LINKED ISSUE #1606: Do not use `realpath('.')` as the fallback working dir
We use the short, handy `realpath('.')` as a fallback working dir

https://github.com/SublimeLinter/SublimeLinter/blob/11607e156389601bb67de3d0a348623490bfac76/lint/linter.py#L1247

Why not just use `os.getcwd()`?

kaste:
Fixes #1606

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -917,8 +917,7 @@ class Linter(metaclass=LinterMeta):
         if not cls.matches_selector(view, settings):
             return False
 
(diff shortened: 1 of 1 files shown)


braver:
Looks like this works, let's ⛵️ it.

kaste:
I wonder if we should push this one to the users as another patch release.

PR #1620, opened 2019-06-21
Compute line:col visual width using 1-based counting

### LINKED ISSUE #1618: Line/column width is not computed correctly in panel
Small visual glitch:

image

Line or columns widths are not computed correctly. 🤷‍♂

kaste:
Fixes #1618

The error store is based on (0, 0) counting. But only in the panel, we present
the errors using natural (1, 1) line/column counting.

Here the width was computed wrong, e.g. for `start=9` we computed a `1`, but
the user is actually seeing a `10` which is 2 chars.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -394,8 +394,8 @@ def fill_panel(window, then=draw_on_main_thread):
                 max,
                 zip(*[
                     (
-                        len(str(error['line'])),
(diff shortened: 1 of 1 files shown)


braver:
Ah, I can repro the bug, and after this it's gone. 👍 ⛵️ it!

PR #1622, opened 2019-06-25
Expect the value of `bin` in a 'package.json' to be `str`

### LINKED ISSUE #1621: SublimeLinter sometimes parses node's package.json file incorrectly
If I use SublimeLinter with a node-based linter, and I have a package.json property named bin that is a string, then I receive a TypeError.

Replication steps and code: https://github.com/ArtskydJ/sublimelinter-eslist-bin-issue

### tldr

I get this error
```
...
File "C:\Users\Joseph\AppData\Roaming\Sublime Text 3\Installed Packages\SublimeLinter.sublime-package\lint/base_linter/node_linter.py", line 126, in find_local_executable
    script = os.path.normpath(os.path.join(path, manifest['bin'][npm_name]))
TypeError: string indices must be integers
```

if I have a package.json file like this

```json
{ "bin": "whatever.js" }
```

but this package.json file works just fine

```json
{ "bin": { "cli": "whatever.js" } }

kaste:
Fixes #1621

We assumed that the value of `bin` is a mapping from `str -> str` but it can
be just a plain str as well. See #1621.

For now, it is probably enough to just catch and eat the thrown `TypeError`.

### DIFF
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -124,7 +124,7 @@ class NodeLinter(linter.Linter):
                 # but must run as a normal script. E.g. `/usr/bin/env node eslint.js`
                 try:
                     script = os.path.normpath(os.path.join(path, manifest['bin'][npm_name]))
-                except KeyError:
+                except (KeyError, TypeError):
                     pass
                 else:
                     if not os.path.exists(os.path.join(path, 'node_modules', '.bin')):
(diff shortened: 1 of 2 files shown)


kaste:
We could release this as a hotfix, if we branch off of the last release and cherry-pick. master is in a dirty state.

braver:
Yeah that’s a good idea, I’ll do some git magic later today to roll out the fix. 👍🏻

Issue #1582, opened 2019-03-28
In the panel show 'affected' files nearby active file


kaste:
Usual sort order for the files in the panel is `a, b, c, d` (surprise!). 

In #1556 we focus the active view, and for that, say `b` is active, we re-sort to `a, c, d, b`. 

Since we have now multi-file linters, we should group 'related/affected' files. This information is stored in `affected_filenames_per_bid`. Say linting b produced results for files `b, x, a`, we should re-sort to `c, d, b, a, x`. 

Rationale: Typical, traditional linters report only per-file, and with #1556 we get a pretty clean picture. Type-checkers (ususally) can show results from dependent files and a user will likely see these results nearby. I think this could be a good default, automatic behavior.

kaste:
@AndrewMast Or this one. That's a graphical thing on top of #1556

AndrewMast:
I'll take two looks then

PR #1624, opened 2019-06-27
Show errors for closed files

### LINKED ISSUE #1607: Show linter output for closed files
Hi guys,

I know it has been some time but I finally wanted to finish the mess I started :smile: and show the output of a multi-file linter also for closed files.

I was looking through the code to see how this could be done but I'm not really sure what the best way is. All views seem to depend on a bid, so my first thought was to use a special bid value to signal that a file is closed and requires special handling. Or something like a bid/filename pair as the key to store errors. These could then be updated later when the files are opened.

Any ideas?

mheinzler:
Closes #1607


mheinzler:
@kaste This works pretty well already with some minor bugs like the active linters in the status bar.

So far, these kinds of errors are cleaned when the original file is linted again and reports no more errors. I guess there should also be a command to do this manually.

I will also start working on observing renames, etc.

kaste:
This looks relatively straight-forward. 👍 

Do you already know why the tests fail?

sublime_linter.py
```diff
@@ -355,58 +368,61 @@ def group_by_filename_and_update(window, bid, view_has_changed, linter, errors):
         grouped[filename]  # For the side-effect of creating a new empty `list`
 
     did_update_main_view = False
+    window_id = main_view.window().id()
     for filename, errors in grouped.items():
+        for error in errors:
+            error['window'] = window_id
```
kaste: I don't like this one at all. The `window_id` does not belong to an error. Also the same file can open in multiple windows, and we can maybe support that in the future.

panel_view.py
```diff
@@ -282,59 +283,45 @@ def draw_on_main_thread(*args, **kwargs):
...
             errors,
             key=lambda e: (e["line"], e["start"], e["end"], e["linter"])
         )
-        for bid, errors in (
-            (bid, errors_by_bid.get(bid))
-            for bid in buffer_ids_per_window(window)
-        )
-        if errors
+        for filename, errors in errors_by_file.items()
+        if errors and errors[0]['window'] == window.id()
```
kaste: This is probably the only place where you use `window` but if you leave the structure of the function and swap `buffer_ids_per_window` with `filenames_per_window` this should work out ok.

sublime_linter.py
```diff
@@ -153,6 +153,18 @@ def on_activated_async(self, view):
...
+            # again, the errors could be cleared in such a case. Instead we
+            # broadcast an event to give everyone a chance to update the new
+            # view.
+            if util.get_filename(view) in persist.file_errors:
+                events.broadcast('loaded_file_with_errors', {
+                    'filename': util.get_filename(view)
+                })
+                return
+
             hit(view, "on_load")
```
kaste: This has edge-cases. Why is relinting such a file a bad idea? Is it generally a bad idea, or just for some linters?

sublime_linter.py
```diff
@@ -329,12 +340,14 @@ def kill_active_popen_calls(bid):
...
 
-def group_by_filename_and_update(window, bid, view_has_changed, linter, errors):
-    # type: (sublime.Window, Bid, ViewChangedFn, LinterName, List[LintError]) -> None
+def group_by_filename_and_update(main_view, view_has_changed, linter, errors):
+    # type: (sublime.View, ViewChangedFn, LinterName, List[LintError]) -> None
     """Group lint errors by filename and update them."""
     if view_has_changed():  # abort early
         return
 
+    bid = main_view.buffer_id()
```
kaste: Generally, the getters `id()`, `buffer_id()`, `view.window()` are not safe, that's why we extract these and pass them down/around. (Here you removed `bid` from the function signature but you still need it and get the information from the view object. Same for `window` below, you removed it from the signature and then get a live version in a for loop.)

sublime_linter.py
```diff
@@ -355,58 +368,61 @@ def group_by_filename_and_update(window, bid, view_has_changed, linter, errors):
...
+            update_file_errors(filename, linter, errors)
 
     # For the main view we MUST *always* report an outcome. This is not for
     # cleanup but functions as a signal that we're done. Merely for the status
     # bar view.
     if not did_update_main_view:
-        update_buffer_errors(bid, linter, [])
+        update_file_errors(util.get_filename(main_view), linter, [])
 
     affected_filenames_per_bid[bid][linter] = current_filenames
```
kaste: Wonder if this should be `affected_filenames_per_filename` instead.

mheinzler:
Thanks for the fix. It is strange because I worked on this for a few days now without ever encountering this error. :confused:

kaste:
You mostly edited top folder files. These reload on save. 😁

sublime_linter.py
```diff
@@ -355,58 +368,61 @@ def group_by_filename_and_update(window, bid, view_has_changed, linter, errors):
...
 
 
-def update_buffer_errors(bid, linter, errors):
-    # type: (Bid, LinterName, List[LintError]) -> None
+def update_file_errors(filename, linter, errors):
+    # type: (LinterName, List[LintError]) -> None
```
kaste: Type error: function takes 3 args

sublime_linter.py
kaste: A `view` is a real, live OOP thing here, all getters are actually side-effects, everything mutates. (It is somewhat save if we would run on the so-called main thread which blocks/freezes the UI, but we run basically all our code multi-threaded so the UI/the world changes underneath us.)

I think we should go for `affected_filenames_per_filename`.

sublime_linter.py
kaste: Yes okay, but the linter will relint maybe in background, on next save etc. and will eventually produce confusing results. So to solve the c++ thing we would probably a different approach. (I actually thought c++ linters would skip linting *just* header files.)

Anyway, I don't like the conditional here. If the user really just changed the syntax, we *want* to relint. It's probably okay to have explicit state here. If we don't find a `view` in `group_by_filename_and_update` for some filename, we store that information in e.g. `awaiting_view = Set[FileName]`. We check that here, flip the switch, draw and return. 

Ideally though, I hoped that this would be local to `highlights_view.py` in that in receives updates 'on_lint_result' just like before and remembers for which filename it couldn't yet draw the highlights. And I think that probably the c++ linter should reject linting on load in that particular case.

panel_view.py
kaste: Yes of course. Then I would say the filenames the panel should consider per window is the set of `{util.get_filename(v) for v in window.views()}` + all the sets from `affected_filenames_per_filename` for these. This would describe clearly: show all errors from the open files and from their dependencies.

sublime_linter.py
```diff
@@ -205,6 +219,20 @@ def on_pre_close(self, view):
...
 
 
+def has_filename_changed(view):
+    bid = view.buffer_id()
+    current_filename = util.get_filename(view)
+
+    try:
+        old_filename = buffer_filenames[bid]
+    except KeyError:
+        return True
```
kaste: Probably `return False` because just opening a file should not count as a rename. (We would rename from `None -> str`.)

sublime_linter.py
```diff
@@ -129,6 +129,7 @@ def other_visible_views():
 
 global_lock = threading.RLock()
 guard_check_linters_for_view = defaultdict(threading.Lock)  # type: DefaultDict[Bid, threading.Lock]
+buffer_filenames = {}  # type: Dict[Bid, FileName]
```
kaste: `buffer_filenames` is new state which needs cleanup in `on_pre_close`

sublime_linter.py
```diff
@@ -152,6 +153,19 @@ def on_activated_async(self, view):
...
+        if has_filename_changed(view):
+            # update the error store and the views
+            new_filename = util.get_filename(view)
+            if old_filename in persist.file_errors:
+                persist.file_errors[new_filename] = persist.file_errors.pop(old_filename)
+
+            events.broadcast('renamed_file', {
+                'new_filename': new_filename,
+                'old_filename': old_filename
+            })
```
kaste: Since we update our store, we should ideally just use `update_buffer_errors` but 'updated_error_positions' violates this already. So this is only a reminder that we might need to change something here in the future.

active_linters_view.py
```diff
@@ -65,26 +65,27 @@ def count_problems(errors):
 
 
 class sublime_linter_assigned(sublime_plugin.WindowCommand):
```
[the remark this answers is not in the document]
kaste: Yeah, `active_linters_view.py` is end-of-lifecycle as it stands but I hope we can come up with another hack for this PR.

kaste:
Ok, managed to test drive a second time. I have two windows open which report different errors on the same file basically through a different context. This obviously confuses the current system. Before we used `bid` which are unique per window. (Yes, if you edit the same file in different windows, Sublime will not work on the same buffer and thus not automatically update the other view.) 

- So, I get wrong errors on that shared file, the last result wins and overwrites the other one. 
- Closing that file in one window can remove the errors from the other window.

Technically, it looks like a buffer id is a tuple `(window.id(), view.filename())`. One way to model that is to make the error store `Dict[Wid, Dict[Filename, [List[LintError]]]]`.

- Drawing the highlights is too late. If I open the panel and then double-click on a result, it will open that file but not activate it. Here it looks like we must listen for real 'load' events.

- Goto-Anything `ctrl+p` will open files in transient mode for preview. If you just scroll through the file list it will remove the errors from the store as soon as it closes that preview view (either bc you select a different file, or doesn't open anything at all aka `ESC`). 

- Not relinting on activate is not an option if you have multiple linters running on the same files. (E.g. a style checker and a type checker) Even if the file was already open it will not lint and thus show errors from other linters if it had *some* errors in the store.

- If you have the panel open, errors from closed files do not show before we modify the buffer. (Likely the errors are in the store but the panel doesn't update.)

That's it for now.

mheinzler:
> Before we used `bid` which are unique per window. (Yes, if you edit the same file in different windows, Sublime will not work on the same buffer and thus not automatically update the other view.)

This is true when you open the file twice, but you can actually clone a view and then both have the same ID even across windows.

> Technically, it looks like a buffer id is a tuple `(window.id(), view.filename())`. One way to model that is to make the error store `Dict[Wid, Dict[Filename, [List[LintError]]]]`.

If we want to fully support cloned views (which are rare but still), something like this might be better (just to make it clear, bad design):
`Dict[Filename, Dict[(actual bid of the view or None, bid of the main view), List[LintError]]]`

This way we would know exactly which errors belong to which window and different buffers can have their own independent errors. The panel can show the errors for the visible views and those associated with them. This is similar to what you wrote above, but with an additional separation per buffer.

Maybe we could also drop `affected_filenames_per_filename` then.

>     * Drawing the highlights is too late. If I open the panel and then double-click on a result, it will open that file but not activate it. Here it looks like we must listen for real 'load' events.

I did this because of the comment _on_load_async fires too early for preview buffers_, but I think it shouldn't matter when we are just adding highlights. Will change it.

>     * Goto-Anything `ctrl+p` will open files in transient mode for preview. If you just scroll through the file list it will remove the errors from the store as soon as it closes that preview view (either bc you select a different file, or doesn't open anything at all aka `ESC`).

Any idea how to detect previews? I quickly tested them but there seems to be no difference to a normal view. We could clean up only when the view has been activated at least once?

>     * Not relinting on activate is not an option if you have multiple linters running on the same files. (E.g. a style checker and a type checker) Even if the file was already open it will not lint and thus show errors from other linters if it had _some_ errors in the store.

OK. I will drop this and move it into the linter.

kaste:
You *can* clone a view and move it to somewhere but what I did is just UI testing. If I double-click the panel in a different window, Sublime creates a new buffer, likewise if I just open via `ctrl+p`. 

So, let's say that the first point, looking at the same file through different windows/contexts, is a configuration problem I should solve as a user. Let's just forget it for now. 

- We keep the current design without the hardcoded `window_id` as I described before with the simple rule: a panel shows all errors from open files in that window plus their dependencies. 

- We just don't throw away errors on close. Removing them automatically feels odd if the linter actually reports for the whole project or larger parts of it. Instead we introduce a new command to 'forget results'. (Even if we don't keep such a command it would be nice-to-have for the UI testing.) Maybe a dropdown: `forget results > [List of filenames + 'all']`

- The panel usage is also a bit confusing bc its focus on the active view. #1582 (and #1556) should address this. But for the UX of this PR it is important.

@mheinzler, which linters are you actually using?

mheinzler:
This PR is for the Clang Tidy C++ linter, but I also use flake8, mypy, eslint, yamllint, ...

kaste:
Great commits!

sublime_linter.py
```diff
@@ -357,56 +377,73 @@ def group_by_filename_and_update(window, bid, view_has_changed, linter, errors):
     did_update_main_view = False
     for filename, errors in grouped.items():
         if not filename:  # backwards compatibility
-            update_buffer_errors(bid, linter, errors)
+            update_file_errors(main_filename, linter, errors)
         else:
-            # search for an open view for this file to get a bid
-            view = window.find_open_file(filename)
+            # search for an open view for this file
+            view = main_view.window().find_open_file(filename)
```
kaste: Don't forget `window()` is one of those live getters and can return None and in fact while testing I already had a None exception here.

kaste: If you look at the signature we had `Wid, Bid, ...` and it looks like we have `Wid, Filename, ...` right now.

mheinzler:
> We just don't throw away errors on close.
How do you want to do this? For all linters or just the ones that can report multiple files?

kaste:
For all linters, I thought let's see how this would feel. Would be definitely a new UX, but users suggested that before so 🤷 maybe that's good.

kaste:
Wonder if that is a good idea/UX. It is very usual to git checkout or (partially) revert files (etc.) these days. And then the errors still stay. Does that mean I would have to run the forget command *manually* all day long; for something that *just works* right now? (We can detect file reloads for *open* files but not for closed ones btw.) 

Iirc, technically a preview is a view where `view in view.window().views()` does not hold. So the view belongs to a window, but the window does not enumerate the view on some/all of its getters.  

But that's an UX thing, you basically have to see it to feel it. As I tested I disliked that the errors went away on close. I thought 'forget' is something you do once in a while but if I had to do this on every checkout ... 🤔

kaste:
Should I do some rebasing here already, or do it yourself. I see 3 do-revert pairs and a couple of fixups. Maybe we can get to a first commit only containing the rename `persist.errors -> file_errors` refactoring (although Github doesn't let me mark specific changesets as reviewed I think).  I can def do the rebase onto master in case it's tricky bc I know the upstream changes good enough 😄

mheinzler:
> But that's an UX thing, you basically have to see it to feel it. As I tested I disliked that the errors went away on close. I thought 'forget' is something you do once in a while but if I had to do this on every checkout ... thinking

Yes. Sometimes I just take a look at some files outside of a project and if I had to clear the panel every time there is an error, it would be very annoying. The current UX feels very good for these 'simple' cases.

I would really like it if we link the errors of dependencies to the linted file, so that when that file is linted again, the other errors will be updated and also removed if the file is closed. When linting a whole project we need something else of course, but that would feel very close to how the files interact with each other.

But I will go ahead and make the errors persistent and also implement the forget command. And then we can see where to go from there.

> Should I do some rebasing here already, or do it yourself. 

I have cleaned some of the commits. Rebasing on master seems like a bit of fun :grin:.

> Maybe we can get to a first commit only containing the rename `persist.errors -> file_errors` refactoring

If I understand correctly, it should be enough to squash the first few commits.

kaste:
Oh my god, the rebase took an hour. I had to fixup my rebased commits. 😳

Will push later.

kaste:
I have the old HEAD tagged, here is the new one.

kaste:
I like the idea of using `affected_files` for cleaning the error store. Say we have file `a, b`, a reports `a -> {b}`, b doesn't report anything. If we close `b`, the errors stay iff `a` is still open otherwise they go. If we close `a`, the errors go if `b` is closed, otherwise `a` goes and `b` stays. 

Is it like that?

For a relation `file -> {deps...}`, we take the set `{file} + closed({deps}) - {all deps from all other files}`?

mheinzler:
> I like the idea of using `affected_files` for cleaning the error store. Say we have file `a, b`, a reports `a -> {b}`, b doesn't report anything. If we close `b`, the errors stay iff `a` is still open otherwise they go. If we close `a`, the errors go if `b` is closed, otherwise `a` goes and `b` stays.

That's a good idea. This will still have some issues if you also have a file `c` that reports for `b`, because the errors from `a` will stay even if only `c` is open. But it seems to cover the most common use case and is simple enough.

I was already testing the forget command, so here it is for reference, but with the above I don't think we need that anymore (for this PR).

kaste:
Okay if we also have a `c` we have `a -> {b}, b -> {}, c -> {b}`. All open.

```
- if we first close a: {a} + closed({b}) - {b} => {a} + {} - {b} = forget {a}

- if we first close b: {b} - closed({})  - {b} - {b}                    = {}
  then a             : {a} + closed({b}) - {b} - {}  => {a} + {b} - {b} = {a}
  then c             : {c} + closed({b}) - {}        => {c} + {b} - {}  = {c, b}
```

mheinzler:
Yes, but what I meant was that when `a` is closed, the errors it reported for `b` (which may be unique to the `a -> {b}` relationship and different from what `c` reported) will still be visible, even though they make no more sense after `a` is gone.

But again, this may be a very special case that doesn't need to be addressed right now. I tried to reproduce it, but then realized it can't even happen at the moment, as the errors for `b` will overwrite each other because they come from the same linter. This may be another thing that can be fixed later if it turns out to be a problem.

mheinzler:
This feels pretty good.

Do you know how to best fix mypy in sublime_linter.py (`Optional[str]` and `Dict` vs. `defaultdict`)?

kaste:
Which mypy errors exactly? The one in `grouped[error.get('filename')].append(error)`. What do you mean with `Dict vs defaultdict`?

kaste:
Whenever I see a bug I just write it down here, okay?

If you open a dependency listed in the panel, it highlights imm if I double-click 👍 but it doesn't draw if I use `ctrl+p` to open the file. on_load fires but window.views() will probably not report it bc it is still a preview.

kaste:
If you run say `mypy` on a view with some errors, then disable the linter, it will print `mypy(ok)` in the status bar. (Ideally the status bar will show just `mypy` on assigned t.i. before the first run.)

kaste:
`update_on_filename_change` has a type error which indicates a smell I think. `has_filename_changed` is not that useful here bc we need to get `old_filename` twice. Maybe `detect_rename(view) -> Optional[Tuple[Oldstr, Newstr]]` then the output is either `None` for no rename or the arg signature of `update_on_filename_change`.

kaste:
I think the sets in `affected_filenames` should *not* store the main_filename. We rather have the mapping `main_file -> set(dependencies)`. This makes discarding errors a bit easier.

mheinzler:
I see the following errors and was wondering how to best tell mypy that the `Optional[str]` are actually `str` in these cases, and also how to assign the `Dict` to the `defaultdict`:

```
lint\linter.py:
 1315:25  error  mypy  Incompatible types (expression has type "Optional[str]", TypedDict item "filename" has type "str")

sublime_linter.py:
  164:13  error  mypy  Argument 1 to "update_on_filename_change" has incompatible type "Optional[str]"; expected "str"
  286:13  error  mypy  Incompatible types in assignment (expression has type "Dict[Any, List[LintError]]", variable has type "defaultdict[str, List[LintError]]")
  409:9   error  mypy  Invalid index type "Optional[str]" for "defaultdict[str, List[LintError]]"; expected type "str"
```

kaste:
I pushed 3 fixups. errors on line 164 and 409 (now 410) are code smells and must be fixed differently. About 164 see ☝️ https://github.com/SublimeLinter/SublimeLinter/pull/1624#issuecomment-511986851

About 409, we should avoid `None`s here. We allow no filename for compatibility but that's only as long as we're in the `Linter` code. In backend finalize we control everything and should set the filename.

sublime_linter.py
```diff
@@ -278,9 +278,10 @@ def input(self, *args):
...
 
     def run(self, filename):
+        # type: (str) -> None
         """Clear errors for one or all filenames."""
         if filename == "<All>":
-            errors_to_clear = persist.file_errors
+            errors_to_clear = persist.file_errors  # type: Mapping[str, List[LintError]]
             persist.file_errors = defaultdict(list)
         else:
             errors_to_clear = {filename: persist.file_errors.pop(filename)}
```
kaste: A defaultdict is not a dict so we basically downcast to something, here: `Mapping`. I think this is an mypy error. We only use `.items()` here so a sophisticated type checker should infer e.g. they state `Dict` but they only use `Mapping`. 

Just like when you do `int -> int -> int` with the implementation `a + b` the type system actually knows that the type can be `Summable[T] -> Summable[T]`. 

So, a `type ignore` would maybe be preferrable?

mheinzler: I have no experience with mypy, but as `Mapping` is fixing the error it is probably better than a ignore which I guess could swallow other errors.

mheinzler:
The panel will now update when previewing closed files. But this does not work when switching to already open files. I can't find any event that is fired in such a case.

kaste:
The bug I reported was that the highlights do not draw imm if I open the file via Ctrl+p.

You say the panel does not update (or scroll?) if you switch to another file. Which method do you use for switching here? Double click in the panel, click on a tab, Ctrl+tab, click in the sidebar...

mheinzler:
Sorry, I thought you were talking about the highlights the panel draws to show at which error the cursor currently is.

The panel does not immediately update while you switch between open files using Ctrl+p, but it does of course when you actually activate the file with Enter. I will drop the latest commit again, seeing that it only works for closed files and isn't very useful.

mheinzler:
These should fix all the bugs you found so far.

kaste:
I had some short Sublime freezes in the last days indicating the main thread is blocked. I switch to master for the next days.

mheinzler:
Continuing the conversation from: https://github.com/mheinzler/SublimeLinter-contrib-clang-tidy/commit/7a2d27864c6780cfda3aba6f7892508afe3325e4

I added the `canonical_filename`. Do you already have an idea how to make `persist` an object (mainly how to name it and in which module)?

But maybe it's easier that I switch to listening to `on_lint_result` events in the linter. This would probably put less restrictions on the development of the core.

kaste:
I want to avoid making `persist` an object for now. 

I switched back to this branch for the last week. 891205b introduces basic caching. (Using this branch I can easily get the linter reporting hundreds of errors, with the caching it is easily ten times faster on medium and large results.) Hope I didn't make an error here.

I will have a look at the panel view. It def not expect to get results via `LINT_RESULT` that fast (multiple times per tick) as we do right now due to splitting a result set into multiple one per filename. (It is also the only part of the code where we explicitly switch to the main thread bc writing to a buffer, measuring its size and scroll position (and changing it) belongs to the main thread. So I think if we're doing something stupid here we *could* freeze Sublime bc that's what the main thread does essentially. 

The status bar is the only thing left not really okay. Do you have a special reason to flip assigned with unassigned I don't see. The main implicit 'unassigner' is actually 'sublime_linter._assign_linters_to_view'. Here we 'assign' linters every lint sweep and by doing so implicitly 'unassign' all others. E.g. all linters now disabled but active in the previous run get unassigned automatically. This also counts for the exclusion filter built into SublimeLinter. 

I still have to think about this.

kaste:
I came up with d5ec71d for the status bar which would be very simple. Works for me but maybe lacking good test scenarios right now.

mheinzler:
The caching looks good. I haven't yet seen so many errors at once but the main thread could be a problem if that happens. I guess for you it was not the number of errors but the many updates they produced? Maybe we could debounce the panel updates to limit the number of updates per second?

> Do you have a special reason to flip assigned with unassigned I don't see.

Because that implicitly assigned all linters also to affected files and updated their status bar correctly (at first). Your fix looks much better though.

kaste:
🎶 How many times I have to write a throttler? 🎶

This is generally good to go though.

kaste:
Next week: Urlaub. Plenty of time for you to find other bugs. @braver We're almost done here, basically need to clean up the commits, check it! You actually don't need multi-file reporters. Just make sure we don't regress for all the other linters.

kaste:
Okay, ready for merge. This already took a bit too long bc summer. (No pun here, just reviewing the same changeset multiple times usually isn't that effective.) We merge and fix everything in separate PR's. We should not ship immediately. Just let it sink a week or so.

- I simplified some set algebra stuff
- I removed the `Clear Errors` command. It was intended for testing here, and I don't want to introduce such a command to the end user. Ideally, clean up must be automatic, and if we don't ship such a command it is much more likely people will complain if we've got it wrong.

The future: 

- We need to look a bit on the performance. Multi-file reporters can report hundreds of errors at once, and there are a lot of opportunities now to make some non-pre-mature optimizations. 
- `context['canonical_filename']` should be used if possible
- rename `util.get_filename -> util.canoncial_filename`

braver:
🎉

mheinzler:
:+1:

PR #1627, opened 2019-07-09
Users can copy linter messages from show_popup tooltip

kylebebak:
This is a minor but nice usability win for users. The tooltip created by `show_popup` now includes a link/button labeled **Copy** that lets users copy the linter output to the clipboard.

Output is formatted and has no HTML tags. Here's an example for a line with both `mypy` and `flake8` errors.

Screen Shot 2019-07-08 at 7 01 03 PM 2

Actual text copied to clipboard:

Screen Shot 2019-07-08 at 7 01 30 PM

Status message printed on copy to give user feedback:

Screen Shot 2019-07-08 at 7 01 24 PM

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -657,7 +657,7 @@ TOOLTIP_STYLES = '''
         font-weight: bold;
     }
     .copy {
-        margin-top: 8px;
+        margin-top: 0.5em;
     }
 '''
 
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -692,12 +692,27 @@ def open_tooltip(view, point, line_report=False):
     if not errors:
         return
 
+    def on_navigate(href: str) -> None:
+        if href == "copy":
+            sublime.set_clipboard(join_msgs_raw(errors))
+            sublime.active_window().status_message("SublimeLinter: info copied to clipboard")
+            view.hide_popup()
+
     tooltip_message = join_msgs(errors, show_count=line_report, width=80)
     view.show_popup(
-        TOOLTIP_TEMPLATE.format(stylesheet=TOOLTIP_STYLES, content=tooltip_message),
+        TOOLTIP_TEMPLATE.format(stylesheet=TOOLTIP_STYLES, content=tooltip_message) + "<a href=\"copy\">Copy</a>",
         flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
         location=point,
-        max_width=1000
+        max_width=1000,
+        on_navigate=on_navigate
+    )
+
+
+def join_msgs_raw(errors):
+    # Take an `errors` iterable and reduce it to a string without HTML tags.
+    sorted_errors = sorted(errors, key=lambda r: (r["linter"], r["error_type"]))
+    return "\n\n".join("{}: {}\n{}{}".format(
+        e["linter"], e["error_type"], e["code"] + " - " if e["code"] else "", e["msg"]) for e in sorted_errors
     )
 
 
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -656,12 +656,16 @@ TOOLTIP_STYLES = '''
         color: var(--yellowish);
         font-weight: bold;
     }
+    .copy {
+        margin-top: 8px;
+    }
 '''
 
 TOOLTIP_TEMPLATE = '''
     <body id="sublimelinter-tooltip">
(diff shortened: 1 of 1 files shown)


kaste:
Hi! That's a good and simple implementation. What's the use case you have here? (*I* usually open the error panel and then just copy paste the text.)

highlight_view.py
```diff
@@ -692,12 +692,27 @@ def open_tooltip(view, point, line_report=False):
...
+    def on_navigate(href: str) -> None:
+        if href == "copy":
+            sublime.set_clipboard(join_msgs_raw(errors))
+            sublime.active_window().status_message("SublimeLinter: info copied to clipboard")
+            view.hide_popup()
+
     tooltip_message = join_msgs(errors, show_count=line_report, width=80)
     view.show_popup(
-        TOOLTIP_TEMPLATE.format(stylesheet=TOOLTIP_STYLES, content=tooltip_message),
+        TOOLTIP_TEMPLATE.format(stylesheet=TOOLTIP_STYLES, content=tooltip_message) + "<a href=\"copy\">Copy</a>",
```
braver: Please add the link to the `TOOLTIP_TEMPLATE` in stead of just slapping it onto the end. Now the `<a>` sits outside the `<body>`, which is probably the cause of it being misaligned.

braver:
I like the idea if we can do it without too much visual clutter. If you're hovering for more info and want to do something with that info, it's a bit of a hassle to have to find the same error in the panel and copy it from there.

highlight_view.py
kylebebak: @braver 
Just did this, thanks for the suggestion!

The copy link now looks like this. It's properly aligned horizontally with the rest of the content, and I gave it a bit of `margin-top` so it's not so vertically close.

Screen Shot 2019-07-09 at 10 01 12 AM

kylebebak:
> Hi! That's a good and simple implementation. What's the use case you have here? (_I_ usually open the error panel and then just copy paste the text.)

I'm glad you think the implementation is alright =)

My use case is much the same as @braver 's -- having the error info copyable from the tooltip is more convenient than having to find it in the panel. I also think it's a natural user expectation that text from any tooltip should be copyable, and the info in SublimeLinter's tooltip is often particularly "worth copying" and looking up after.

highlight_view.py
```diff
@@ -692,12 +696,27 @@ def open_tooltip(view, point, line_report=False):
     if not errors:
         return
 
+    def on_navigate(href: str) -> None:
+        if href == "copy":
+            sublime.set_clipboard(join_msgs_raw(errors))
+            sublime.active_window().status_message("SublimeLinter: info copied to clipboard")
```
braver: In stead of taking the active window here, you can get the window the view belongs to. I don’t have the API docs at hand right now, bet you can find that easily. I think that’s a bit cleaner since the pop up belongs to a view and you act on the view on the next line.

braver:
Excellent!

highlight_view.py
kylebebak: Good suggestion, I just made that change!

highlight_view.py
```diff
@@ -656,12 +656,16 @@ def run(self):
         color: var(--yellowish);
         font-weight: bold;
     }
+    .copy {
+        margin-top: 8px;
+    }
```
braver: Sorry for not putting all the feedback in one go, but I’m also kinda just thinking aloud here and reacting from my phone...

So, iirc we can use em’s here instead of pixels and that would cause everything to scale nicely with font size. However, that’s not necessarily the best option. Perhaps we can steal what the scope name popup does that’s built into ST (i.e. it’s in the Default package so we can look at it, it’s not really documented or anything).

Anyway, if what I just wrote makes any sense to yourself out 😄 otherwise I’ll have a look at it myself later.

kylebebak: Also a good call, I replaced `8px` with `0.5em` and it looks good regardless of how big or small you make ST's font.

Hopefully this is ready to merge now!

braver: Woohoo 🎉 thanks man! Great work 👍🏻

braver:
@kaste I’m happy here, you too?

braver:
💃

kaste:
That really looks and works nice. Great 💐

kylebebak:
Wooohoooo! Thanks guys

PR #1630, opened 2019-07-22
prep 4.12

braver:
@kaste Please check the message and if there is anything else we should mention. I don't want to list all the deprecations per se, just point everyone to the relevant PR.

### DIFF
--- a/messages/4.12.0.txt
+++ b/messages/4.12.0.txt
@@ -1,0 +1,14 @@
+SublimeLinter 4.12
+==================
+
+- Improved errors tooltip:
+  Now wraps long lines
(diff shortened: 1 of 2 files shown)


messages/4.12.0.txt
```diff
@@ -0,0 +1,14 @@
+SublimeLinter 4.12
+==================
+
+- Improved errors tooltip:
+  Now wraps long lines
+  Easily copy the error messages to you clipboar
```
kaste: A d for the clipboard, and some flowers bc this was a contribution.

braver: Oh, sure, of course!

PR #1638, opened 2019-07-27
Clarify two examples in linter_settings.rst.

fsaad:
The examples in the "styles" section now show that the
linter-specific settings need to be nested under the "linters" key,
similar to the example of yamllint in the "selector" section.

This example is now consistent with the example for "selector".


braver:
Excellent, thanks for the contribution!

PR #1645, opened 2019-08-11
Start 'xvfb' service on Travis' xenial

kaste:
Let's start the fix travis game here


kaste:
Howdy, this worked on the first try.

PR #1649, opened 2019-08-31
Iterate `NodeLinter` executable lookup

### LINKED ISSUE #1626: Wrong working directory on nested local eslint installs (breaks eslint@6)
In my mono-repository setup I have the following structure:

```
bar
  - .eslintrc
  - package.json < installs eslint, eslint-plugin-mocha
  - foo.js < file to be linted
package.json < parent package.json, in mono-repository setup
```

Linting via SublimeLinter-eslint uses the project root as the working directory despite the fact that `eslint` is located in the sub-directory. Because of that eslint fails to locate plug-ins, installed in the sub-directory, next to the eslint installation:

```
SublimeLinter: #11 linter.py:1478     Running ...

  {{DIR}}  (working dir)
  $ cat bar/foo.js | {{DIR}}/bar/node_modules/.bin/eslint --format json --stdin --stdin-filename {{DIR}}/bar/foo.js

SublimeLinter: #11 linter.py:49       ERROR: 

Oops! Something went wrong! :(

ESLint: 6.0.1.

ESLint couldn't find the plugin "eslint-plugin-mocha".

(The package "eslint-plugin-mocha" was not found when loaded as a Node module from the directory "{{DIR}}".)
```

__Expected Behavior__

The directory into which `eslint` is installed (locally) is used as a `cwd` when running the command.

__Reproduce__

I've created [a example setup](https://github.com/nikku/sublime-eslint-test/) that shows the issue. To reproduce, clone and install it

```
git clone git@github.com:nikku/sublime-eslint-test.git
(cd sublime-eslint-test/bar && npm install
```

Then open `sublime-eslint-test` in Sublime Text, navigate to `sublime-eslint-test/bar/foo.js` in Sublime Text.

### LINKED ISSUE #1644: Support Yarn Plug'n'Play
This is more of a feature request. When working on a project that uses Yarn's [Plug'n'Play (PnP)](https://yarnpkg.com/lang/en/docs/pnp/) feature to manage dependency resolution SublimeLinter complains that it cannot find eslint ..

> SublimeLinter: WARNING: Skipping 'eslint' for now which is listed as a devDependency in C:\Path\To\package.json but not installed.  Forgot to 'npm install'? 

I suppose the current implementation [depends on the presence of a local `node_modules` folder](https://github.com/SublimeLinter/SublimeLinter/blob/9ed6b21468e9b5f4d0b03f52dedf78258e553a7e/lint/base_linter/node_linter.py#L68-L70)?

HusamIbrahim:
Fixes #1626 
Fixes #1644 

TODO:
- ~Address #1626~
- ~Tests~

### DIFF
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -7,7 +7,7 @@ import logging
 import os
 import shutil
 
-from .. import linter
+from .. import linter, util
 
 
 if False:
@@ -102,12 +102,12 @@ class NodeLinter(linter.Linter):
 
     def find_local_executable(self, start_dir, npm_name):
         # type: (str, str) -> Union[None, str, List[str]]
-        for path in paths_upwards(start_dir):
+        for path in paths_upwards_until_home(start_dir):
             executable = shutil.which(npm_name, path=os.path.join(path, 'node_modules', '.bin'))
             if executable:
+                self.context['project_root'] = path
                 return executable
 
-        for path in paths_upwards_until_home(start_dir):
             manifest_file = os.path.join(path, 'package.json')
             if os.path.exists(manifest_file):
                 try:
@@ -137,6 +137,7 @@ class NodeLinter(linter.Linter):
 
                     node_binary = self.which('node')
                     if node_binary:
+                        self.context['project_root'] = path
                         return [node_binary, script]
 
                     logger.warning(
@@ -146,21 +147,69 @@ class NodeLinter(linter.Linter):
                     self.notify_failure()
                     raise linter.PermanentError()
 
-                # A 'package.json' not yet installed?
-                if not os.path.exists(os.path.join(path, 'node_modules', '.bin')):
-                    is_dep = bool(manifest.get('dependencies', {}).get(npm_name))
-                    is_dev_dep = bool(manifest.get('devDependencies', {}).get(npm_name))
-                    if is_dep or is_dev_dep:
-                        logger.warning(
-                            "Skipping '{}' for now which is listed as a {} "
-                            "in {} but not installed.  Forgot to 'npm install'?"
-                            .format(
-                                npm_name,
-                                'dependency' if is_dep else 'devDependency',
-                                manifest_file
+                is_dep = bool(manifest.get('dependencies', {}).get(npm_name))
+                is_dev_dep = bool(manifest.get('devDependencies', {}).get(npm_name))
+                if is_dep or is_dev_dep:
+                    self.context['project_root'] = path
+
+                    # Perhaps this is a Yarn PnP project?
+                    yarn_lock_exists = os.path.exists(os.path.join(path, 'yarn.lock'))
+                    package_lock_exists = os.path.exists(os.path.join(path, 'package-lock.json'))
+                    pnp_js_exists = os.path.exists(os.path.join(path, '.pnp.js'))
+                    pnp_in_manifest = bool(manifest.get('installConfig', {}).get('pnp'))
+                    is_pnp_project = pnp_js_exists or pnp_in_manifest
+
+                    if yarn_lock_exists and is_pnp_project:
+                        yarn_binary = shutil.which('yarn')
+                        if not yarn_binary:
+                            logger.warning(
+                                "This seems like a Yarn PnP project. However, finding "
+                                "a Yarn executable failed. Make sure to install Yarn first."
                             )
+                            self.notify_failure()
+                            raise linter.PermanentError()
+                        elif pnp_js_exists:
+                            return [yarn_binary, 'run', '--silent', npm_name]
+
+                    next_path = os.path.dirname(path)
+
+                    for path in paths_upwards(next_path):
+                        executable = shutil.which(npm_name, path=os.path.join(path, 'node_modules', '.bin'))
+                        if executable:
+                            return executable
+
+                    logger.warning(
+                        "Skipping '{}' for now which is listed as a {} "
+                        "in {} but not installed.  Forgot to '{} install'?"
+                        .format(
+                            npm_name,
+                            'dependency' if is_dep else 'devDependency',
+                            manifest_file,
+                            'yarn' if yarn_lock_exists and not package_lock_exists else 'npm'
                         )
-                        self.notify_failure()
-                        raise linter.PermanentError()
+                    )
+                    self.notify_failure()
+                    raise linter.PermanentError()
 
         return None
+
+    def run(self, cmd, code):
+        # type: (Union[List[str], None], str) -> Union[util.popen_output, str]
+        result = super().run(cmd, code)
+
+        try:
+            npm_name = cmd[3]
+            if ('error Command "' + npm_name + '" not found') in str(result):
+                logger.warning(
+                    "We did execute 'yarn run --silent {0}' but "
+                    "'{0}' cannot be found.  Forgot to 'yarn install'?"
+                    .format(
+                        npm_name
+                    )
+                )
+                self.notify_failure()
+                raise linter.PermanentError()
+        except IndexError:
+            pass
+
+        return result
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -171,7 +171,9 @@ class NodeLinter(linter.Linter):
 
                     next_path = os.path.dirname(path)
                     for path in paths_upwards(next_path):
-                        executable = shutil.which(npm_name, path=os.path.join(path, 'node_modules', '.bin'))
+                        executable = shutil.which(
+                            npm_name, path=os.path.join(path, 'node_modules', '.bin')
+                        )
                         if executable:
                             return executable
 
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -192,22 +192,22 @@ class NodeLinter(linter.Linter):
         return None
 
     def run(self, cmd, code):
-        # type: (Union[List[str], None], str) -> Union[util.popen_output, str]
+        # type: (Optional[List[str]], str) -> Union[util.popen_output, str]
         result = super().run(cmd, code)
 
-        try:
+        if cmd and cmd[1:3] == ['run', '--silent'] and len(cmd) >= 4:
             npm_name = cmd[3]
-            if ('error Command "' + npm_name + '" not found') in str(result):
+            if 'error Command "{}" not found'.format(npm_name) in (
+                result.stderr or ''
+                if isinstance(result, util.popen_output)
+                else result
+            ):
                 logger.warning(
                     "We did execute 'yarn run --silent {0}' but "
                     "'{0}' cannot be found.  Forgot to 'yarn install'?"
-                    .format(
-                        npm_name
-                    )
+                    .format(npm_name)
                 )
                 self.notify_failure()
                 raise linter.PermanentError()
-        except IndexError:
-            pass
 
         return result
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -168,8 +168,8 @@ class NodeLinter(linter.Linter):
                             )
                             self.notify_failure()
                             raise linter.PermanentError()
-                        elif pnp_js_exists:
-                            return [yarn_binary, 'run', '--silent', npm_name]
+
+                        return [yarn_binary, 'run', '--silent', npm_name]
 
                     next_path = os.path.dirname(path)
 
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -154,30 +154,28 @@ class NodeLinter(linter.Linter):
 
                     # Perhaps this is a Yarn PnP project?
                     yarn_lock_exists = os.path.exists(os.path.join(path, 'yarn.lock'))
-                    package_lock_exists = os.path.exists(os.path.join(path, 'package-lock.json'))
-                    pnp_js_exists = os.path.exists(os.path.join(path, '.pnp.js'))
-                    pnp_in_manifest = bool(manifest.get('installConfig', {}).get('pnp'))
-                    is_pnp_project = pnp_js_exists or pnp_in_manifest
-
-                    if yarn_lock_exists and is_pnp_project:
+                    if yarn_lock_exists and (
+                        os.path.exists(os.path.join(path, '.pnp.js'))
+                        or bool(manifest.get('installConfig', {}).get('pnp'))
+                    ):
                         yarn_binary = shutil.which('yarn')
-                        if not yarn_binary:
-                            logger.warning(
-                                "This seems like a Yarn PnP project. However, finding "
-                                "a Yarn executable failed. Make sure to install Yarn first."
-                            )
-                            self.notify_failure()
-                            raise linter.PermanentError()
+                        if yarn_binary:
+                            return [yarn_binary, 'run', '--silent', npm_name]
 
-                        return [yarn_binary, 'run', '--silent', npm_name]
+                        logger.warning(
+                            "This seems like a Yarn PnP project. However, finding "
+                            "a Yarn executable failed. Make sure to install Yarn first."
+                        )
+                        self.notify_failure()
+                        raise linter.PermanentError()
 
                     next_path = os.path.dirname(path)
-
                     for path in paths_upwards(next_path):
                         executable = shutil.which(npm_name, path=os.path.join(path, 'node_modules', '.bin'))
                         if executable:
                             return executable
 
+                    package_lock_exists = os.path.exists(os.path.join(path, 'package-lock.json'))
                     logger.warning(
                         "Skipping '{}' for now which is listed as a {} "
                         "in {} but not installed.  Forgot to '{} install'?"
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -1,7 +1,7 @@
 """This module exports the NodeLinter subclass of Linter."""
 
 from functools import lru_cache
-from itertools import takewhile
+from itertools import chain, takewhile
 import json
 import logging
 import os
@@ -34,9 +34,9 @@ def paths_upwards(path):
         path = next_path
 
 
-def paths_upwards_until_home(path):
-    # type: (str) -> Iterator[str]
-    return takewhile(lambda p: p != HOME, paths_upwards(path))
+def paths_upwards_until_home(path, home=HOME):
+    # type: (str, str) -> Iterator[str]
+    return chain(takewhile(lambda p: p != home, paths_upwards(path)), [home])
 
 
 def read_json_file(path):
@@ -102,7 +102,8 @@ class NodeLinter(linter.Linter):
 
     def find_local_executable(self, start_dir, npm_name):
         # type: (str, str) -> Union[None, str, List[str]]
-        for path in paths_upwards_until_home(start_dir):
+        paths = paths_upwards_until_home(start_dir)
+        for path in paths:
             executable = shutil.which(npm_name, path=os.path.join(path, 'node_modules', '.bin'))
             if executable:
                 self.context['project_root'] = path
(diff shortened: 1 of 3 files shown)


HusamIbrahim:
@kaste I'd appreciate your input on this. For now I tested manually with npm and yarn v1 and v2 (w/ and w/o pnp). Seems to work fine for simple test cases.

kaste:
👍 You see how easy it can be. 

Notes: 

1. Looking for `yarn.lock` is too wide. There are a lot of packages/project shipping such a file although end-user can install or hack on the project using `npm`. (We don't want to prefer `yarn` here bc ... market share.)

2. What's the reason for the `yarn bin` call?

3. `shutil.which('yarn')` can return `None` for failure.

Ad 2: Is this to test if the project has been installed, like we do for `npm install` has been run a couple of lines below?  If so, isn't the purpose of plug'n'play it just works after `git clone`. In the 'berry' package there is also a `.yarn/cache` checked in. Should we look for that instead?

HusamIbrahim:
> 👍 You see how easy it can be.

lol the hardest part so far has been having a development setup in place on windows. I eventually settled on `winpdb` for debugging. I wonder what you guys use (windows specific/cross platform setup)? Also currently if I introduce changes to the code I'm creating a package file and copying it over to `Installed Packages` to be able to test it. Is there a way to make this process less cumbersome? I guess I might be spoiled by live reload/hmr in the javascript ecosystem? The thing is there's so little information around about Sublime plugin development.

HusamIbrahim:
>  Looking for `yarn.lock` is too wide. There are a lot of packages/project shipping such a file although end-user can install or hack on the project using `npm`. (We don't want to prefer `yarn` here bc ... market share.)

I did this as per **arcanis**'s [recommendation](https://github.com/SublimeLinter/SublimeLinter/issues/1644#issuecomment-524825828). Unfortunately other markers such as `.yarn/cache`  or `.pnp.js`  would either exclude yarn v1 or conflict with other package managers that might implement PnP in the future (or both).

> What's the reason for the yarn bin call?

Yes that was to test if a project was installed. I know for example that a user can manually edit a `package.json` file to add a dependency before running `npm install` or `yarn install`. So the presence of a `yarn.lock` or `.yarn/cache` does not guarantee that a dependency has already been installed.

> shutil.which('yarn') can return None for failure.

I kind of assumed that if `yarn bin` ran successfully then `shutil.which('yarn')` should do too, shouldn't it?

kaste:
About the setup, it's actually easy and less cumbersome. Just clone SublimeLinter into your packages folder. (This one `Preferences > Browse Packages`.) From there on, the git version takes precedence over the installed package. (Maybe restart Sublime.) If you done with hacking and a PR, just delete the git directory, maybe restart, and the installed package will be used again. 

On hot reloading: Every top level file gets hot-reloaded on save. (These contain the views and event handlers.) The imported modules/files (subfolders) you manually reload using the shipped command `SublimeLinter: Reload SublimeLinter ...`. Usually you set this as a build command so you can just `ctlr+b`.

The testing: Just install the package UnitTesting from PackageControl. After that `Test current package`. Use `Test current file` on any test file to program it which file counts as 'the current file'.

kaste:
> > shutil.which('yarn') can return None for failure.

> I kind of assumed that if yarn bin ran successfully then shutil.which('yarn') should do too, shouldn't it?

Yeah, sure. Unlikely it throws at runtime but it's a TypeError here. We return a `List[str]` not a `List[Optional[str]]` iirc. (It is actually far more likely than you think because `which` is a fast python implementation whereby you run the `yarn bin` command with a full shell.)

HusamIbrahim:
> About the setup, it's actually easy and less cumbersome. Just clone SublimeLinter into your packages folder. (This one `Preferences > Browse Packages`.) From there on, the git version takes precedence over the installed package. (Maybe restart Sublime.) If you done with hacking and a PR, just delete the git directory, maybe restart, and the installed package will be used again.
> 
> On hot reloading: Every top level file gets hot-reloaded on save. (These contain the views and event handlers.) The imported modules/files (subfolders) you manually reload using the shipped command `SublimeLinter: Reload SublimeLinter ...`. Usually you set this as a build command so you can just `ctlr+b`.
> 
> The testing: Just install the package UnitTesting from PackageControl. After that `Test current package`. Use `Test current file` on any test file to program it which file counts as 'the current file'.

Yeah the testing part I kind of figured out from the import statements in the test files. But the rest of this is pure gold man 👍. I spent an entire day trying to get something set up to work with. Should've asked first and saved myself all the trouble.

HusamIbrahim:
> Yeah, sure. Unlikely it throws at runtime but it's a TypeError here. We return a `List[str]` not a `List[Optional[str]]` iirc. (It is actually far more likely than you think because `which` is a fast python implementation whereby you run the `yarn bin` command with a full shell.)

I see. I'll modify the PR to handle that case then. Other than that I guess I should put this on hold until the dust settles on which marker we should use, no?

HusamIbrahim:
> Looking for yarn.lock is too wide. There are a lot of packages/project shipping such a file although end-user can install or hack on the project using npm. (We don't want to prefer yarn here bc ... market share.)

Sorry I didn't get this at first. Do you think I should give yarn a lower priority then (after checking for the script edge case)?

kaste:
Well, which marker we use. I don't know yet. But at the point where you look for the marker `yarn.lock` we just know that this dir contains a `package.json`. This is too early. 

Following the code we then try to read that `package.json`, then handle the uninteresting edge case of 'using eslint while hacking on it', only then we check if e.g. 'eslint' is actually declared as a dependency or devDependency.

Iirc we want this: Find the package.json which declares a dependency on the `cmd_name`. Store this directory as `self.context['project_root']` so we later run the command from this working dir. Search starting from this dir (and including this dir) for ordinary 'binaries' in 'node_moudles/bin/' upwards. This counts for npm installs, regular old-school yarn installs, lerna too I think. If found, be happy and return it. (We have a binary and a working dir, so everything is ok. We generally should prefer binaries even if a 'yarn run' would also do this job bc just running the executable should be faster and more efficient than invoking a wrapper script. 🤷‍♂)

If not, try a marker side-by-side from the package.json, e.g. '.yarn/cache', '.pnp.js' and 'installConfig: pnp: true' in package.json bc they truly mark intend. If found return 'yarn run ...' if 'which yarn' is successful. Otherwise tell the user 'yarn must be installed e.g. 'npm install -g yarn'.  

If not check for 'yarn.lock' and tell the user to run `npm install` or `yarn` depending on the presence of 'yarn.lock'. 

This would not check `yarn bin` because it is probably too expensive to run on each invocation. We therefore check after `run` if stderr contains 'error Command "eslint" not found.'. If so and the actual 'cmd' started with 'yarn run --silent' we tell the user she should run 'yarn' to install all dependencies. 

That's how I currently see this feature. Might be the right thing, might too much, not enough 🤷‍♀ (The description includes #1626 though.)

HusamIbrahim:
This is quite detailed and provides a nice framework/set of rules to work with. Might make testing a bit more difficult but it should cover all bases. Let me see what I can come up with.

HusamIbrahim:
@kaste Made some updates. What do you think?

kaste:
This is generally the right direction but we basically need to turn it upside down. The old node resolution algo was to first just look for `node_module/bin/cmd_name` going upwards. Super naive but worked for years. Only if this did not succeed we actually searched for 'package.json' files, most of the time for a better error message. 

We want to turn this upside down. First look for `package.json` files going upwards, and *then* look for executables in `node_modules/bin` upwards, or other ways e.g. yarn pnp to run the linter. (This is now necessary because different package manager strategies put the executable not nearby the package.json anymore.)

HusamIbrahim:
But won't that come at the expense of efficiency for the majority of users? My guess is that most users still use `npm` or normal `yarn` (no PnP), and for them parsing `package.json` is going to introduce unnecessary overhead when you could just run an executable from `node_modules` 😕

HusamIbrahim:
Re: #1626. While browsing the code I stumbled upon the `working_dir` setting. Can't we use that if it wasn't set by the user instead of introducing `self.context['working_dir']`?

kaste:
> But won't that come at the expense of efficiency for the majority of users? 

Yes, but I don't see another solution for #1626 and we actually cache reading `package.json`.

> While browsing the code I stumbled upon the working_dir setting.

Yes, we have options here. But in general we *receive* info from the user and the plugins using `self.settings` and we *provide* info to the user and the plugins via `self.context`.

HusamIbrahim:
> Yes, but I don't see another solution for #1626 and we actually cache reading `package.json`.

Ah okay I didn't notice that. I'll shuffle things around then to look for `package.json` first. 

> Yes, we have options here. But in general we receive info from the user and the plugins using self.settings and we provide info to the user and the plugins via self.context.

I see. So should I stick to that naming convention then - `self.context['working_dir']` instead of `self.context['linter_working_dir']`?

kaste:
I suggest using `project_root` or something similar.

HusamIbrahim:
@kaste More updates. Tests seem to be holding but I don't know if I'm going in the right direction here.

lint/base_linter/node_linter.py
```diff
@@ -102,12 +102,12 @@ def get_start_dir(self):
 
     def find_local_executable(self, start_dir, npm_name):
         # type: (str, str) -> Union[None, str, List[str]]
-        for path in paths_upwards(start_dir):
+        for path in paths_upwards_until_home(start_dir):
             executable = shutil.which(npm_name, path=os.path.join(path, 'node_modules', '.bin'))
             if executable:
+                self.context['project_root'] = path
                 return executable
```
kaste: That's a nice happy path shortcut for traditional installs 👌

lint/base_linter/node_linter.py
```diff
@@ -146,11 +147,45 @@ def find_local_executable(self, start_dir, npm_name):
                     self.notify_failure()
                     raise linter.PermanentError()
 
-                # A 'package.json' not yet installed?
+                # No 'node_modules' folder ?
                 if not os.path.exists(os.path.join(path, 'node_modules', '.bin')):
```
kaste: We shouldn't test this here bc we're looking for package.json files which declare a dependency in this loop.

lint/base_linter/node_linter.py
```diff
@@ -146,11 +147,45 @@ def find_local_executable(self, start_dir, npm_name):
...
                     is_dep = bool(manifest.get('dependencies', {}).get(npm_name))
                     is_dev_dep = bool(manifest.get('devDependencies', {}).get(npm_name))
                     if is_dep or is_dev_dep:
```
kaste: I think we can immediately have the side effect of setting project_root here bc we have found the right package.json.

lint/base_linter/node_linter.py
```diff
@@ -146,11 +147,45 @@ def find_local_executable(self, start_dir, npm_name):
...
+                                    .format(
+                                        npm_name,
+                                        'dependency' if is_dep else 'devDependency',
+                                        manifest_file
+                                    )
+                                )
+                                self.notify_failure()
+                                raise linter.PermanentError()
+
+                        # Definitely not a Yarn PnP project
```
kaste: Then here comes the inner loop searching upwards for binaries. If it fails we tell the user to run npm or yarn install.

kaste: This must be exhaustive bc if we found a package.json we def break the outer loop!

HusamIbrahim:
Thanks for reviewing @kaste. I've addressed your comments in the latest commit.

kaste:
Boy, I've written good tests. They don't fail although you've rewritten the whole function/feature. 🧐

Can you make it green, only one (probably style error) from flake8.

HusamIbrahim:
> Boy, I've written good tests. They don't fail although you've rewritten the whole function/feature. 🧐

Lol no denying that 😁

> Can you make it green, only one (probably style error) from flake8.

👍

lint/base_linter/node_linter.py
```diff
@@ -146,21 +147,67 @@ def find_local_executable(self, start_dir, npm_name):
...
+
+                    # Perhaps this is a Yarn PnP project?
+                    yarn_path = shutil.which('yarn')
+                    yarn_lock_exists = os.path.exists(os.path.join(path, 'yarn.lock'))
+                    package_lock_exists = os.path.exists(os.path.join(path, 'package-lock.json'))
+                    pnp_js_exists = os.path.exists(os.path.join(path, '.pnp.js'))
+                    pnp_in_manifest = bool(manifest.get('installConfig', {}).get('pnp'))
+
+                    if yarn_path is None:
+                        if yarn_lock_exists and (pnp_in_manifest or pnp_js_exists):
```
kaste: Why checking for yarn.lock? Isn't a yarn.lock mandatory when you use yarn? Even if if it doesn't exist it would be still a pnp project before the first install?

lint/base_linter/node_linter.py
```diff
@@ -146,21 +147,67 @@ def find_local_executable(self, start_dir, npm_name):
...
-                                npm_name,
-                                'dependency' if is_dep else 'devDependency',
-                                manifest_file
+                is_dep = bool(manifest.get('dependencies', {}).get(npm_name))
+                is_dev_dep = bool(manifest.get('devDependencies', {}).get(npm_name))
+                if is_dep or is_dev_dep:
+                    self.context['project_root'] = path
+
+                    # Perhaps this is a Yarn PnP project?
+                    yarn_path = shutil.which('yarn')
```
kaste: which yarn is way more expensive than the other checks, so we should think about flipping the conditionals. I.e. on Windows you often have lots of paths in PATH, which does an fs.stat for every path whereby checking if pnp.js exists is just one fs.stat call.

lint/base_linter/node_linter.py
```diff
@@ -146,21 +147,67 @@ def find_local_executable(self, start_dir, npm_name):
...
+
+                    if yarn_path is None:
+                        if yarn_lock_exists and (pnp_in_manifest or pnp_js_exists):
+                            logger.warning(
+                                "This seems like a Yarn PnP project. However, finding "
+                                "a Yarn executable failed. Make sure to install Yarn first."
                             )
+                            self.notify_failure()
+                            raise linter.PermanentError()
+                    elif yarn_lock_exists and pnp_js_exists:
```
kaste: You only check for pnp.js. Probably better to have a var is_yarn_pnp_project which combines all possible checks using or.

lint/base_linter/node_linter.py
```diff
@@ -146,21 +147,67 @@ def find_local_executable(self, start_dir, npm_name):
...
+                            return executable
+
+                    logger.warning(
+                        "Skipping '{}' for now which is listed as a {} "
+                        "in {} but not installed.  Forgot to '{} install'?"
+                        .format(
+                            npm_name,
+                            'dependency' if is_dep else 'devDependency',
+                            manifest_file,
+                            'yarn' if yarn_path and yarn_lock_exists and not package_lock_exists else 'npm'
```
kaste: Is having a yarn.lock and a package.json lock a real thing? What does that actually mean: I don't care use whatever you want? Or is that a config error on projects maintainer side?

lint/base_linter/node_linter.py
```diff
@@ -146,21 +147,67 @@ def find_local_executable(self, start_dir, npm_name):
...
                         )
-                        self.notify_failure()
-                        raise linter.PermanentError()
+                    )
+                    self.notify_failure()
+                    raise linter.PermanentError()
 
         return None
+
+    def run(self, cmd, code):
```
kaste: Reminder: Needs type annotations. Copy from super class probably.

lint/base_linter/node_linter.py
```diff
@@ -146,21 +147,67 @@ def find_local_executable(self, start_dir, npm_name):
...
+        result = super().run(cmd, code)
+
+        try:
+            if ('error Command "' + cmd[3] + '" not found') in str(result):
+                logger.warning(
+                    "Skipping '{}' for now which is listed as a dependency/devDependency "
```
kaste: Maybe: "We did run yarn run --silent cmd but the command cmd cannot be found.  Forgot to run ...." Then we don't have to think about a probable edge case that project_root is not set. I.e. if the user set executable manually.

HusamIbrahim:
> I need to look at the tests as well. The old tests did not fail so we have to check if they're really end-to-end tests. Also need to check if adding tests for the new functionality is feasible. 

I thought a few new tests were needed. Cases:

- Yarn project but no yarn executable.
- Properly installed yarn PnP project.
- When running a cmd starting with `yarn run --silent` results in `error Command "cmd" not found` (Don't know how easy this would be to mock?)
-  #1626

And maybe update `test_uninstalled_local_dependency` to include yarn projects. It all depends how much coverage you want for this feature and how much you see the implementation changing in the future. What do you think?

lint/base_linter/node_linter.py
HusamIbrahim: According to **arcanis** [other package managers](https://github.com/SublimeLinter/SublimeLinter/issues/1644#issuecomment-524825828) such as `pnpm` are considering implementing pnp.

lint/base_linter/node_linter.py
HusamIbrahim: I noticed there's a `self.which` method. Is that more optimized or does it just delegate to `shutil.which`?

lint/base_linter/node_linter.py
HusamIbrahim: Maybe better to use `is_pnp_project` in case you guys want to support other package managers which might implement `pnp` in the future? Combines `pnp_js_exists` and `pnp_in_manifest`. Maybe we could add `.yarn/cache` but that's yarn 2 specific?

lint/base_linter/node_linter.py
HusamIbrahim: It *is* strongly discouraged to mix package managers like that, but still a user might checkout a repo that has yarn.lock and install the project using npm. I know yarn provides a warning for the mirror case but since npm is more widely used 🤷‍♀

lint/base_linter/node_linter.py
HusamIbrahim: Will do 👍

lint/base_linter/node_linter.py
kaste: 👍

kaste:
Good work here. The tests are actually nice, I feared they would become terrible but they're just okay.

lint/base_linter/node_linter.py
kaste: Yeah, typically

	if is_yarn_pnp_project:
		yarn_path = ...
		if not yarn_path:
			raise ...
		return [yarn, run, ...]

no?

lint/base_linter/node_linter.py
kaste: Then we don't have `yarn_path` defined here and we just don't test for it, just * yarn.lock* but not *packagejson.lock* should be ok?

kaste:
For the `run` test, I usually mock `when(linter)._communicate(...).thenReturn(OUTPUT)` and the then start the test with `linter.lint('foo', lambda: False)`.

HusamIbrahim:
> For the `run` test, I usually mock `when(linter)._communicate(...).thenReturn(OUTPUT)` and the then start the test with `linter.lint('foo', lambda: False)`.

Thanks for the tip. I'll give it a try.

kaste:
That's an awesome PR you did here 👏

HusamIbrahim:
> That's an awesome PR you did here 👏

All thanks to your help 🍻

kaste:
🎉

HusamIbrahim:
Yaaay 🎉🥳

ildella:
Hi. I see there's been a long discussion and I read it all. 
I have a brand new project with yarn v2 w/ pnp, and there's no way the Linter works in Sublime. 
It just work when, without pnp, the /bin/eslint file is still in its usual place. 

The issue is that v2 pnp saves deps in .yarn/cache as zip, and `yarn bin eslint" returns exactly a path into a zip and overriding the `path` settins like

```
    "paths": {
        "linux": [".yarn/cache/eslint-npm-5.16.0-91d3d3fc21.zip/node_modules/eslint/bin/eslint.js"],
    },
```

Does not change anything. 

From CLI, `yarn lint` properly call the script executing `eslint .` but there is no way to have it working in Sublime. Any tip?

kaste:
Setting paths should have no effect here, actually all should be automatic. Enable debug mode and look at the Sublime console what it prints. Open a new issue and post the log in case it's not helping. The important parts are if it finds your package.json file together with e.g. a yarn.lock and a pnp.js etc.

PR #1653, opened 2019-09-19
Add GitHub actions for flake8 and sphinx

kaste:
Let's add flake8 and sphinx as GitHub actions and remove them from travis.


kaste:
I have a mypy workflow stashed. But mypy is still too red to add it now.

PR #1655, opened 2019-09-20
Remove docstring which had a lint error

kaste:


kaste:
Such docstring don't have any value imo.

braver:
regardless of the lint error, that docstring didn't even say anything useful. 👍

kaste:
The docstrings are a strange thing here in Sublime python land.

PR #1657, opened 2019-09-20
Deprecate all the things we forgot

### LINKED ISSUE #1633: Deprecate defining `config_file`
Just like all the other things we had e.g. `npm_name` we do not have a built in config look atm. So defining `config_file` has no effect and should be removed.

### LINKED ISSUE #1641: Remove `Linter.tab_width` setting
`tab_width` (the default is `1` which does nothing) is unused throughout all the registered linters so it can go away, probably without deprecation, just the warning that this thing doesn't do anything anymore.

🙌

### LINKED ISSUE #1646: Deprecate '*' usage in `cmd`
We forgot this one 🤷‍♀. Instead of '*' we use the explicit '$args' these days, we forgot to print a warning here.

### LINKED ISSUE #1647: Proper deprecate in `Linter.finalize_cmd`
We use `logger.info` here which is too low. We should use our helper fn `deprecation_warning`.

### LINKED ISSUE #1651: Deprecate old-school project settings
We still support the old nested project settings, and forgot to deprecate here.  

I don't remember any issue with the new, flat view settings so it is time to deprecated the old thing, to get a chance for removal later.

kaste:
Fixes #1633 
Fixes #1651 
Fixes #1647 
Fixes #1646 
Fixes #1641

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -449,7 +449,7 @@ class LinterMeta(type):
             'comment_re', 'shebang_match',
             'npm_name', 'composer_name',
             'executable', 'executable_path',
-            'tab_width'
+            'tab_width', 'config_file'
         ):
             if key in attrs:
                 logger.warning(
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -448,12 +448,13 @@ class LinterMeta(type):
             'inline_settings', 'inline_overrides',
             'comment_re', 'shebang_match',
             'npm_name', 'composer_name',
-            'executable', 'executable_path'
+            'executable', 'executable_path',
+            'tab_width'
         ):
             if key in attrs:
                 logger.warning(
                     "{}: Defining 'cls.{}' has no effect. Please cleanup and "
-                    "remove these settings.".format(name, key))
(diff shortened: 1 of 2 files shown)


braver:
👌

braver:
@kaste since the milestone is done now, wanna ship? Any notes to go with it? I can also run some tests this week by using it, see if anything comes up, and then ship. ❓

kaste:
Yes, ready to 🚢! But we need notes esp. bc the meat here is done via contributors. 

- mheinzler basically took everything upside down. Everything is now filename based and instead of based on view buffer_ids. We have full multi-file linter support. (And it took us both 2 months.)

- Ibrahim rewrote the executable lookup for NodeLinter. We 🤞 support eslint6, the newest yarn pnp, lerna.

- 'kill_switch' default flip is interesting for users.

- Next milestone will be around performance. We have the same performance for old linters, but if the linter produces a lot of errors bc it lints the whole project we're too slow. (We don't flicker or block, we're just wasting CPU on a thread.)

braver:
👍🏻

PR #1659, opened 2019-09-20
Set `kill_old_processes` to `True` by default

### LINKED ISSUE #1639: Flip `kill_old_processes` switch
We actually wanted to flip the default here to `true`. (And if this works out okay maybe remove the setting.) 

(I have this set to true on my Windows, so it looks proper for that OS at least.)

kaste:
Fixes #1639


braver:
Never saw a negative side effect so let's just roll with it. 🚢

PR #1660, opened 2019-09-20
Expose `lint_mode` as a normal setting to the user

### LINKED ISSUE #1640: Expose final `lint_mode` as ordinary view setting
For 'lint_mode' we have the following dance to get to the final value.

https://github.com/SublimeLinter/SublimeLinter/blob/64cde020583f0e7afc0059048272f1a1a8d1787e/lint/linter.py#L1043-L1044

This should not be necessary as settings already is a `ChainMap` constructed in `get_raw_linter_settings`

kaste:
Fixes #1640

I.e. a plugin author should get the effective `lint_mode` just using

		settings.get('lint_mode')

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -306,15 +306,7 @@ def get_linter_settings(linter, view, context=None):
 
 def get_raw_linter_settings(linter, view):
     # type: (Type[Linter], sublime.View) -> MutableMapping[str, Any]
-    """Return 'raw' linter settings without variables substituted.
-
-    Settings are merged in the following order:
-
-    default settings (on the class)
-    global user settings
-    project settings
-    view settings
-    """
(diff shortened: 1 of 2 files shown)


braver:
So, this simplifies the way a plugin author can get the lint mode? Do we need docs for that? 

Also, should the plugin author get the lint reason? Or is that going overboard?

kaste:
Well, we don't have docs for all the stuff that happened this year 😁 but the way 'lint_mode' works *after* this PR is just expected behavior, and before it was a bug bc settings should always be transparent. It shouldn't be hard to figure out what the 'lint_mode' is set to.

`reason` is actually passed down via `self.context.get('reason')`. (But only in master I think.)

braver:
Excellent 🙆‍♂

PR #1666, opened 2019-09-26
Fix: Inner `path` var should not shadow outer

kaste:
An inner var should not shadow the outer loop var is usually a linter rule. 😁 

Here we also immediately have the error that we compute the 'package-lock.json' path wrong.

### DIFF
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -172,9 +172,9 @@ class NodeLinter(linter.Linter):
 
                     # Since we've found a valid 'package.json' as our 'project_root'
                     # exhaust outer loop looking just for installations.
-                    for path in paths:
+                    for path_ in paths:
                         executable = shutil.which(
-                            npm_name, path=os.path.join(path, 'node_modules', '.bin')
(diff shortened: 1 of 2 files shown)


lint/base_linter/node_linter.py
```diff
@@ -172,9 +172,9 @@ def find_local_executable(self, start_dir, npm_name):
 
                     # Since we've found a valid 'package.json' as our 'project_root'
                     # exhaust outer loop looking just for installations.
-                    for path in paths:
+                    for path_ in paths:
```
braver: I don't like the look of a dangling underscore, but 🤷‍♂ Just `p` because it's short lived?

kaste: Yeah, okay

braver:
Other than that I have no test cases, so I feel like I have to give at least some feedback. To show interest etc 😊 otherwise totally agree with this change, just merge it already 👍🏻

lint/base_linter/node_linter.py
kaste: Single char var is just as ugly. Not important enough.

braver: Excellent 👌🏻

Issue #1663, opened 2019-09-25
Unable to load eslint in 4.13


MaxGuitet:
Since upgrading to 4.13 eslint is no longer loaded. I have a message 
```
SublimeLinter WARNING: Skipping 'eslint' for now which is listed as a devDependency in package.json but not installed.  Forgot to 'npm install'?
```
Going back to 4.12 works as expected.

Any idea ?

kaste:
That's interesting esp. because it only logs 'package.json'. I would expect a full path to 'package.json' here. 

We need more info though. Please turn on debug mode and tell us which eslint binary it uses on 4.12.  Then switch to 4.13 and dump the log here.  Esp. working dir, top folder you have open, file_path you're linting, your OS of course.

kaste:
In the Sublime console: What does `import os; print(os.path.expanduser('~'))` print. Or rather does this point to your actual HOME directory?

cosmocoder:
Linting messages and color hightlights also stopped working for me after upgrading to v4.13, but I don't get the "skipping eslint" warning as mentioned by OP. How do I manually install v 4.12 and see if it fixes the linting?

kaste:
@cosmocoder Please dump the debug output as well.

cosmocoder:
@kaste I have the following in the Sublime Text console after enabling debug mode of SunlimeLinter:
```
SublimeLinter: log_handler.py:74      Logging installed; log level INFO
SublimeLinter: sublime_linter.py:68   debug mode: on
SublimeLinter: sublime_linter.py:69   version: 4.13.0
Traceback (most recent call last):
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 478, in run_callback
    expr()
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 619, in <lambda>
    run_callback('on_activated', callback, lambda: callback.on_activated(v))
  File "/Users/nilok/Library/Application Support/Sublime Text 3/Installed Packages/WordHighlight.sublime-package/word_highlight.py", line 125, in on_activated
    Pref.prev_selections = None
AttributeError: 'dict' object has no attribute 'prev_selections'
Path of tsserver.js: /Users/nilok/Library/Application Support/Sublime Text 3/Packages/TypeScript/tsserver/tsserver.js
Path of tsc.js: /Users/nilok/Library/Application Support/Sublime Text 3/Packages/TypeScript/tsserver/tsc.js
[...]

PR #1667, opened 2019-09-26
Fix: Do not fallback to home dir

kaste:
For the outer loop we always used `paths_upwards_until_home` which
is obviously an error if `start_dir` is not below HOME.

So we check `start_dir` first and instead just go `paths_upwards`
if it is not related to HOME.

### DIFF
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -34,9 +34,9 @@ def paths_upwards(path):
         path = next_path
 
 
-def paths_upwards_until_home(path, home=HOME):
-    # type: (str, str) -> Iterator[str]
-    return chain(takewhile(lambda p: p != home, paths_upwards(path)), [home])
+def paths_upwards_until_home(path):
+    # type: (str) -> Iterator[str]
+    return chain(takewhile(lambda p: p != HOME, paths_upwards(path)), [HOME])
 
 
 def read_json_file(path):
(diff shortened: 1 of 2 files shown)


kaste:
E.g.

```
>>> list(node_linter.paths_upwards_until_home('./foo/b'))
['./foo/b', './foo', '.', '', HOME]
```

Here the input is a relative path which is kind of unexpected input for us but anyway, `HOME` is prepended which is probably not intended. Also notice we yield `''` (the empty string) which probably leads to what we see in #1663 (where we have a plain 'package.json' but expect a full, absolute path).

braver:
I'm not 100% sure what the case is that is fixed here. Of course you can't walk up until home if you're not below home. But linters usually do look for config in parent directories, and if that doesn't yield anything they also check the home dir. So I can see where that comes from.

kaste:
There is probably no use case we fix here, we just define an edge case. Generally we're looking for package.json manifests here, the node resolution if you require something is to just go up and look for node_modules folders. Never go sideways. T.i. in node modules are visible and usable if they're somewhere up even if there are not referenced in the local package.json. They're hoisted.

Practically this is mostly irrelevant bc "nobody" has a package.json in their HOME folder. But e.g. for Windows it is much more likely that you don't work in your home directory but map network/cloud folders to separate drives. Then the algo usually starts on some folder on D or E, goes up, and finally ends on the user folder on C which is just strange.

A further note: You can only define such edge cases when you have tests. They basically drive such code. Nobody wants or can manually test edge cases but with tests it is the other way around: nobody wants to test the happy path, we explicitly test the unhappy paths through the code.

braver:
Oh, yeah, sure 🙄 you don't have package jsons in your home dir. Well, I mean, I do, sometimes if I forget to put it in `~/bin` or something, but I don't expect that to hold relevant information for a project I'm linting.

PR #1669, opened 2019-09-26
Log `start_dir` for `NodeLinter.find_local_executable`

kaste:


kaste:
I'll make a release quick.

kaste:
Can we take over GitSavvy? The admin/original author left the building 🙄 since months.

braver:
> Can we take over GitSavvy? The admin/original author left the building 🙄 since months.

Oh noes! I've been using SVN mostly, or professionally at least lately, so I was out of the loop. But if we're done moving stuff over to Git, I will definitely need GitSavvy again too. I'll check back in a few months. If GitSavvy is still stuck then, we'll just have to go and take it over.

PR #1670, opened 2019-09-27
Fix: file regex of the panel view

kaste:
Since we auto-wrap long messages it is possible that any line can
end with a literal ':'.

https://regex101.com/r/Uzgmhh/1/

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -232,7 +232,7 @@ def get_panel(window):
 def create_panel(window):
     panel = window.create_output_panel(PANEL_NAME)
 
-    panel.settings().set("result_file_regex", r"^([^:]+):$")
+    panel.settings().set("result_file_regex", r"^((?::\\|[^:])+):$")
     panel.settings().set("result_line_regex", r"^ +(\d+):(\d+) ")
 
     syntax_path = "Packages/SublimeLinter/panel/panel.sublime-syntax"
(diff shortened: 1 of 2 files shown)


kaste:
I had this bug last week and stashed this patch. @braver please take a triple look, we don't want to break the obvious.

braver:
Won’t be in the same postal code of my laptop until later next week.

kaste:
The fix of course doesn't work if we show absolute paths on Windows bc they contain a : after the drive letter. 🤔

kaste:
Shit stash.

braver:
Good catch, I wouldn't have gotten that just by staring at the regex. So you need to accept `^[A-D]:` or something.

kaste:
Esp. A and B bc I always do my backups on floppies and sure stop on D. 😁

braver:
Why did I type D and not Z? I don't know. Too many things going on today.
Of course, you're going to look on your DVD burner at D, but not on your netwerkshare that's going to be on M. 

I don't look back fondly on having to deal with this crap in Windows. It's affecting my ability to contribute to this issue :)

kaste:
Then we go test driven

kaste:
Hilarious, looks like `view.find_all_results()` returns pseudo linux paths on my Windows, but Windows paths on the travis linux. 🤣

kaste:
Ok, green

braver:
Excellent 🙆‍♂

PR #1673, opened 2019-10-04
WIP: Boost performance

### LINKED ISSUE #1623: With `show_panel_on_save` panel continues to reopen when results fly in even if the user closed the panel inbetween
With `show_panel_on_save` and multiple linters running on the same view. The panel opens automatically as soon as 'negative' results fly in which works as intended. 

If you now immediately close the panel, slow linters from the same "sweep/trigger" still re-open the panel. We must respect the decision the user made ("Not now, go away").

### LINKED ISSUE #1662: The performance issue
Let's talk about performance. In the following, I measure (as always) with an mediocre laptop in battery mode. We measure propagating errors, i.e. the runtime of the `group_by_filename_and_update` fn (in `sublime_linter.py`) generally using `util.print_runtime`.

v4.13 as the bottom line. 

```
A: (panel closed, active file has 4 errors)
0 errors across 0 files from annotations. 146 errors in the store
group_by_filename_and_update took 24ms [D]

0 errors across 0 files from flake8. 146 errors in the store
group_by_filename_and_update took 22ms [D]

8 errors across 5 files from mypy. 146 errors in the store
group_by_filename_and_update took 82ms [D]
```

```
B: (panel closed, active file has 52 errors)
0 errors across 0 files from annotations. 146 errors in the store
group_by_filename_and_update took 118ms [D]

0 errors across 0 files from flake8. 146 errors in the store
group_by_filename_and_update took 107ms [D]

58 errors across 5 files from mypy. 146 errors in the store
group_by_filename_and_update took 226ms [D]
```

```
C: (panel open, active file has 1 error)
0 errors across 0 files from annotations. 146 errors in the store
group_by_filename_and_update took 66ms [D]

0 errors across 0 files from flake8. 146 errors in the store
group_by_filename_and_update took 47ms [D]

68 errors across 29 files from mypy. 146 errors in the store
group_by_filename_and_update took 731ms [D]   <--- !
```

The code here does not block the UI, but it runs on the (shared) worker thread which means it blocks other plugins and essentially all "async" events to run.

A1 and A2 are often noops bc they don't add new errors; A3 is already an excess for just 8 errors (but across 5 files). For B, B1 and B2 even pay a cost just because count of errors of that file is bigger. All of B is above 100ms. C3 shows an excess.

Luckily this is all code we never optimized a single bit; admittedly we must aim for a 10x improvement.

### LINKED ISSUE #1671: Guard `fill_panel` against hidden panel
In the following line

https://github.com/SublimeLinter/SublimeLinter/blob/8a96c0bf0a73c36b1fd2edc5253508c89788ca78/panel_view.py#L422-L422

`viewport_extent()` will return `0` iff the panel is hidden which in turn breaks the function later in `format_error`.

We usually have checks *before* we call `fill_panel` that the panel is active but of course edge cases ... 

Ideally we don't query the view about its extent here but "just have it" in the app state.

kaste:
Various performance commits. 

```
A 
0 errors across 0 files from annotations. 150 errors in the store
group_by_filename_and_update took 35ms [D]  --> 4ms

0 errors across 0 files from flake8. 150 errors in the store
group_by_filename_and_update took 75ms [D]  --> 23ms

8 errors across 5 files from mypy. 150 errors in the store
group_by_filename_and_update took 113ms [D] --> 16ms
```
```
B
0 errors across 0 files from annotations. 150 errors in the store
group_by_filename_and_update took 166ms [D] --> 4ms

0 errors across 0 files from flake8. 150 errors in the store
group_by_filename_and_update took 127ms [D] --> 25ms

58 errors across 5 files from mypy. 150 errors in the store
group_by_filename_and_update took 292ms [D] --> 23ms
```
```
C
0 errors across 0 files from annotations. 150 errors in the store
group_by_filename_and_update took 83ms [D]  --> 8ms

0 errors across 0 files from flake8. 150 errors in the store
group_by_filename_and_update took 46ms [D]  --> 7ms

70 errors across 29 files from mypy. 150 errors in the store
group_by_filename_and_update took 975ms [D] --> 42ms
```

Must be read commit by commit. 

Solves #1623 for free in a pragmatic way: if the results change it will open the panel again. Otherwise the panel stays closed.

`settings` are now generally cached in memory. Greatly improves `settings.has_changed` which was very cheap and idiosyncratic and now just works as anyone would expect. 

Introduces new event `lint_result_changed` which otherwise is the same as our known 'lint_result' with a cache in front. Notable: Since we usually rely on our settings as a global, t.i. we're never pure anywhere probably, we use a settings 'change-count' as additional cache salt.

Fixes #1623
Fixes #1662 
Fixes #1671

### DIFF
--- a/active_linters_view.py
+++ b/active_linters_view.py
@@ -55,6 +55,10 @@ def redraw_file(filename, linter_name, errors, **kwargs):
     else:
         problems.pop(linter_name, None)
 
+    sublime.set_timeout(lambda: redraw_file_(filename, problems))
+
+
+def redraw_file_(filename, problems):
     for view in views_into_file(filename):
         draw(view, problems)
 
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -157,7 +157,8 @@ class ViewListCleanupController(sublime_plugin.EventListener):
 class UpdateErrorRegions(sublime_plugin.EventListener):
     @util.distinct_until_buffer_changed
     def on_modified_async(self, view):
-        update_error_regions(view)
+        if util.is_lintable(view):
+            update_error_regions(view)
 
 
 def update_error_regions(view):
@@ -172,23 +173,32 @@ def update_error_regions(view):
         if '.Highlights.' in key
     }
 
+    changed = False
     for error in errors:
         uid = error['uid']
         region = uid_region_map.get(uid, None)
-        if region is None:
+        if region is None or region == error['region']:
             continue
 
         line, start = view.rowcol(region.begin())
         endLine, end = view.rowcol(region.end())
-        error.update({
-            'region': region,
-            'line': line,
-            'start': start,
-            'endLine': endLine,
-            'end': end
-        })
+        if (
+            line != error['line']
+            or start != error['start']
+            or (endLine != error['endLine'] if 'endLine' in error else False)
+            or end != error['end']
+        ):
+            error.update({
+                'region': region,
+                'line': line,
+                'start': start,
+                'endLine': endLine,
+                'end': end
+            })
+            changed = True
 
-    events.broadcast('updated_error_positions', {'view': view})
+    if changed:
+        events.broadcast('updated_error_positions', {'filename': filename})
 
 
 def head(iterable):
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -473,8 +473,10 @@ def prepare_gutter_data(view, linter_name, errors):
             continue
 
         scope = style.get_icon_scope(error)
-        pos = get_line_start(view, error['line'])
-        region = view.line(pos)
+        # We draw gutter icons with `flag=sublime.HIDDEN`. The actual width
+        # of the region doesn't matter bc Sublime will draw an icon only
+        # on the beginning line, which is exactly what we want.
+        region = error['region']
 
         # We group towards the optimal sublime API usage:
         #   view.add_regions(uuid(), [region], scope, icon)
@@ -498,8 +500,7 @@ def prepare_highlights_data(view, linter_name, errors, demote_predicate):
         mark_style = style.get_value('mark_style', error, 'none')
 
         region = error['region']
-
-        selected_text = view.substr(region)
+        selected_text = error.get('offending_text') or view.substr(region)
         # Work around Sublime bug, which cannot draw 'underlines' on spaces
         if mark_style in UNDERLINE_STYLES and SOME_WS.search(selected_text):
             mark_style = FALLBACK_MARK_STYLE
@@ -530,10 +531,6 @@ def prepare_highlights_data(view, linter_name, errors, demote_predicate):
         by_region_id[region_id] = (scope, flags, regions)
 
     return by_region_id
-
-
-def get_line_start(view, line):
-    return view.text_point(line, 0)
 
 
 def undraw(view):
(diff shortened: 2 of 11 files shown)


kaste:
I don't like the complicated throttler in panel_view so I reverted e803a44 and the previous throttler 2cd32b7. 

Now, 5284df0 and the new tip b76c6b0 have the same performance. I think all commits are safe, easy to grasp, except e30f67a is (maybe?) tricky. 

With the performance now we basically hit the wall (with warm cache), in that the low ms 5, 6, 8 are spent in `group_by_filename_and_update` itself and not in the event listeners/view anymore.

braver:
Nifty. Are you ready for a review here or want to do more work first?

kaste:
This is good enough for review. Sure I can beautify and add some comments and better commit messages ... but I think this works and should be safe.

braver:
Huh, I was under the impression that accessing settings via the ST API was essentially free. Maybe there is a trick to it, I'm not sure, it's been a while. Maybe `load_settings` is not free, whereas get any particular value will get it from memory directly. Not doing a `load_settings` if you don't need to sounds like a good idea anyway.

So, yeah, looks good indeed. I like how you improved the panel view, it's actually a lot simpler to read whereas e803a44 was a bit much. 

I'll run with it for a few days, see if any errors turn up.

kaste:
I maybe go for the extra mile and open separate PRs here. Looks like these are good improvements/patches with clear boundaries.

Also looks like that while we have on_lint_result which fires unconditionally we could just invent lint_result_changed. Usually a view doesn't need to update for the same result set. (This wasn't important for single file linters bc it is at least likely that the report changes for the file you're editing, but for multi file linters it is not so. You edit one file and the other ten files reported probably stay the same.) Don't need to do this immediately but it sounds reasonable.

kaste:
No separate PR but I actually introduce 'lint_result_changed' bc that's really nicer and makes sense. 

Please look at `highlights_view` bc I compute the gutter region slightly different. I don't see a difference visually.  Maybe you know why we computed a 'line-region' for these icons.  On my Sublime, it works with the 'normal' region just as well. 

LGTM

braver:
Maybe something about multiple errors on a line, or otherwise overlapping regions?

kaste:
Good idea! The line comes from SL3 where the actual region was used to filter overlapping errors, but we just use the reported line for that now.

braver:
I’ll try and put it through the ringer later today. Other than that the branch has been really solid for me.

braver:
Ok, it looked good enough for what I tried now, but I need to try something more complicated with eslint. Will try to break it tomorrow and will report back.

braver:
Ok, most of this looks and behaves well. I can't remember how to trigger overlapping multi-line errors. I seem to remember that happening with TSLint sometimes, but I haven't done any TypeScript for some time now and don't have any projects lying around. 

That said, there is one problem: the tooltips on the gutter no longer work. On the regions themselves they're fine but this doesn't work anymore (screenshot taken on master while hovering over the line number):

Screen Shot 2019-10-16 at 22 20 44

kaste:
You're really a good tester. 👏 👍

braver:
🙇🏼‍♂️

kaste:
Fixed.

braver:
Nice! Let's 🛥 this!

kaste:
Workflow question: How do you checkout after I force-push?

braver:
Well usually I just pull and then notice it doesn’t fast-forward. So then I switch to master, remove the branch, and checkout the branch again.

kaste:
🙌 

(After force-push, I `reset --hard` to upstream, but also not without this oh shit, I didn't want that moment.)

braver:
Wanna ship this now as a minor release?

kaste:
I'm on it right now

braver:
🎉

PR #1674, opened 2019-10-14
Add setting to show/hide Lint This View in Context Menu

olavk:


kaste:
We don't want to introduce a setting for that. This is easy to automate as proposed by you and discussed in the issue.

PR #1675, opened 2019-10-21
Clear `deprecation_warning` cache after reloading plugin

kaste:
This is for convenience. If a user fixes a linter plugin, she wants
to see all deprecation warnings again and again, but hopefully none
of course.


braver:
👍

PR #1676, opened 2019-10-21
Fix NodeLinter test when run locally

kaste:
For these tests, we don't run in a new window but just reuse the
current one. The current one of course usually has a working dir
attached, and if so we will issue another `logger.info` about where
we start searching for an executable.

On the other side, on the server/CI we work with an empty window.

Since it's faster to not open new windows all the time, we fix the
test by mofifying our assertion.

Introduced by 92437f6dd6af9e9313d752f06342f16677b57f79


braver:
OK

PR #1677, opened 2019-10-22
Fix: Goto should participate in Sublime's move-history

kaste:
Actually, a comment here suggest that this works but Sublime's
inbuilt `go back/forth` functionality was broken since 333ebc2.

We participate with the `history_list` by manipulating it
imperativly.

Notes:

- We convert the command to a `TextCommand` bc that  is a better fit.

- After that, we *technically* don't need the
  `_sublime_linter_move_cursor` command (bc we're already in a
  `TextCommand`) but since `move_to` is a function we could call from
  everywhere we keep the invariant local for now. :shrug:


kaste:
That's an old bug, and I never had the time to file or fix it, although I use this command and 🙄 every time it did not work.

braver:
Oh, darn. I don't even know what this `go back/forth` thing does. Is it useful?

kaste:
Well I have it on 

```
  { "keys": ["alt+left"], "command": "jump_back" },
  { "keys": ["alt+right"], "command": "jump_forward" },
```

so I use it very often. E.g. if you jump to a definition (symbol?) (don't remember the Sublime terminology here) you can, well, just jump back 😲 

(Well, they could have fine-tuned this over the years 🙄.  Bet there are dozens issues open on the tracker.)

braver:
Oh that’s really handy. 8 years and I’m still learning new stuff.

braver:
Wow, goto back/forward is really handy!

On master, going back does *something*, but not what I would expect. With this PR it does what I would expect. I do notice that clicking in the output panel seems to kind of clear the goto history or something. That's probably ok, clicking in built-in output panels like find-results also doesn't co-operate with goto back/forward.

So, all good! Nice fix. 🚤 !

kaste:
Yeah, the panel double clicking does not cooperate.  Nothing we can change afaik, we just register the regexes.

PR #1681, opened 2019-11-07
Iterate panel again

### LINKED ISSUE #1415: output panel only for current view

### LINKED ISSUE #1582: In the panel show 'affected' files nearby active file

### LINKED ISSUE #1679: Panel updates too often for multi-file reporters
In the performance PR #1673 I reverted https://github.com/SublimeLinter/SublimeLinter/commit/2cd32b7a6bc0d34b538aa8a71cd2930569da4c88 (and its successor https://github.com/SublimeLinter/SublimeLinter/commit/e803a44a38d70d7ac8581ff98f94858cd2df30d4) bc they looked complicated and unnecessary.  But we forgot that these commits weren't sole performance optimizations but visual optimizations.  Now, without them the panel (can) again look like this:

aBj98NYsQl

Basically, it looks like the errors streaming in and we just have too many updates which is totally ugly and not *just too slow*. 😞  

I mark as bug because it is really a visual bug.

kaste:
Fixes #1415 
Fixes #1582 
Fixes #1679 

So again the panel gets some more ❤️, and it's awesome.

1. As shown in the video in #1679 we needed a "throttler". 
2. Finally rebased #1556 
3. Simplified hack to hold the viewport stable.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -534,35 +534,24 @@ def update_panel_selection(active_view, cursor, draw_info=None, **kwargs):
 
 #   Visual side-effects   #
 
+
 def update_panel_content(panel, text):
     if not text:
         text = "No lint results."
-    panel.run_command('_sublime_linter_update_panel_content', {'text': text})
+    panel.run_command('_sublime_linter_replace_panel_content', {'text': text})
 
 
-class _sublime_linter_update_panel_content(sublime_plugin.TextCommand):
+class _sublime_linter_replace_panel_content(sublime_plugin.TextCommand):
     def run(self, edit, text):
-        """Replace a view's text entirely and try to hold the viewport stable."""
         view = self.view
-        x, _ = view.viewport_position()
-
+        _, y = view.viewport_position()
         view.set_read_only(False)
         view.replace(edit, sublime.Region(0, view.size()), text)
         view.set_read_only(True)
-
-        # We cannot measure the `viewport_position` until right after this
-        # command actually finished. So we defer to the next tick/micro-task
-        # using `set_timeout`.
-        sublime.set_timeout(
-            lambda: view.run_command('_sublime_linter_pin_x_axis', {'x': x})
-        )
-
-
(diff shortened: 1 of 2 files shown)


braver:
Oh, I need to set aside some time to play with this. Sounds good!

kaste:
Hope 🤞 this works because it's super awesome. That's the one PR **I** wanted for 6 months.

braver:
This is 🔥🔥

I wanna try if I can throw some more complicated scenarios at it, but some first tests are really positive.

braver:
If had to nitpick anything here, it's that now it's really nice to keep the panel open all the time, but sometimes I have a slow linter and when I open a file it will say "no lint results" until the linter is done. It's too easy to interpret this as "all green", instead of "results inconclusive, come back later". 

Functionally though, it feels rock solid and smooth to me.

kaste:
Yeah, if you make a good UI change, and especially something instant without flicker, new enhancement requests arise immediately.

kaste:
(Have you tried: Make the panel bigger; make the whole window smaller. The filename stays. 😲)

braver:
> make the whole window smaller. The filename stays

wut 😱

PR #1682, opened 2019-11-07
Fix: 'Lint' command is only available for normal buffers

### LINKED ISSUE #1680: Context menu "Lint this view" shows up on scratch views
Of course you need a linter installed matching the syntax. Easy to reproduce with the 'annotations' linter. Then you can actually 'lint' GitSavvy interfaces and even our own error panel. 😲

kaste:
Fixes #1680

Overall we omit the menu entry only if the underlying buffer is
e.g. a scratch view or panel. On views for which no linter is
installed (aka 'assignable') we just gray out the menu entry.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -289,7 +289,13 @@ class sublime_linter_lint(sublime_plugin.TextCommand):
     """A command that lints the current view if it has a linter."""
 
     def is_enabled(self):
-        return any(elect.runnable_linters_for_view(self.view, 'on_user_request'))
+        return (
+            util.is_lintable(self.view)
+            and any(elect.runnable_linters_for_view(self.view, 'on_user_request'))
+        )
+
+    def is_visible(self):
+        return util.is_lintable(self.view)
 
     def run(self, edit):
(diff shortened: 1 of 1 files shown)


braver:
Oh, that's very nice!

.. I have a bunch of other packages installed that should do this actually.

kaste:
This is okay for the context menu.  For the Command Palette, grey always means not in the list. Since users check these commands to see if SL is somehow present, it might be better to register a separate command just for the Command Palette which could be always present and then flash the status bar or give some good advice on selection. 🤔

braver:
Yeah, that is tricky about the command palette. You can't tell why something is not there. But there are lots of commands in the palette that are contextual, so this fits the pattern. I also think we already have pretty good indicators in the statusbar that there are linters active (so kicking off a linter makes sense). And if someone wants to find out if SL is installed it should run "list packages" instead of trying to see if a certain command is there. 

That said, it's probably a good idea if the command were listed in the palette if a view could be linted in theory (i.e. it's not a scratch view), and then we flash the statusbar because there are no linters.

PR #1683, opened 2019-11-07
Log when no linters match a view

### LINKED ISSUE #1668: Log if there are no linters installed, or none can be selected for the given view
As long as there are no linters at all, we could emit a `warning`. If none match the views syntax, we should still log an `info`. 

Otherwise users see as last debug message e.g. `"Delay linting 'eslint' for 0.3s"` and then silence which is not very helpful.

kaste:
Fixes #1668

I decided against a warning when no linters are installed. This was actually never a problem in the issues. And if we *warn*, we also must throttle otherwise we would spam the console etc. 

So, I just `log.info` which only prints if in debug mode. It prints of course when no linters are installed at all, or when the user fiddles with the syntaxes/selectors and oit doesn't match.  The last case is the typical problem when you try to find your own selector.

### DIFF
--- a/sublime_linter.py
+++ b/sublime_linter.py
@@ -334,6 +334,9 @@ def lint(view, view_has_changed, lock, reason):
     This function MUST run on a thread because it blocks!
     """
     linters = list(elect.assignable_linters_for_view(view, reason))
+    if not linters:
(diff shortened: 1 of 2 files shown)


braver:
👌

Issue #1685, opened 2019-11-07
Refactor `util.which` to use `shutil.which`


kaste:
`shutil.which` actually does the right thing plus handles absolute paths. T.i. it is basically a combination of our `util.can_exec` and `util.which`. 

After refactor, `util.can_exec` should become unused and could then be removed.

PR #1686, opened 2019-11-07
Maybe resolve executable set by user

### LINKED ISSUE #1635: Allow to define just the executable name in `executable`
This comes from https://github.com/SublimeLinter/SublimeLinter-gcc/pull/17#issuecomment-514578619 but I actually used this as a tip'n'trick a lot.

If you set `executable` to a `str` we check (validate! LOL) the value using `can_exec`. Probably just to give a error message. We don't do this if you set an array. E.g. `['flake8']` will just pass and *maybe* work (if it is in PATH), but the naive `'flake8'` will never pass because we require an abs path in that case. 

Two options here: a) We actually use `which` on the input, which will return a given abs path just as is, and otherwise resolves it. And then print the error message if it can't resolve and returns `None`.  b) We don't check at all, and for the error message we compute a good one if the `popen` call actually fails.

kaste:
Fixes #1635

We actually only ever expected absolute paths in the 'executable'
setting. Of course, users try short executable names here, and
wonder if it fails.

'executable' can be a plain string or a list of strings. In the latter
case, the real executable is the first item in the list.

We now either check if the given 'real' executable is there and
actually executable if it is an absolute path, or try to resolve
it with the usual `which` call.

As always, the error handling and messages make everything relatively
long.

### DIFF
--- a/tests/test_command_generation.py
+++ b/tests/test_command_generation.py
@@ -168,17 +168,13 @@ class TestExecutableSetting(_BaseTestCase):
             cmd = ('fake_linter_1',)
             defaults = {'selector': None}
 
-        settings = {'executable': 'my_linter'}
-        result = ['my_linter']
+        EXECUTABLE = 'my_linter'
+        RESULT = ['resolved']
+        settings = {'executable': EXECUTABLE}
 
         linter = FakeLinter(self.view, settings)
-        # XXX: We probably don't need to test `can_exec`
-        # - Popen will also throw and show the error panel
-        # - User could just set e.g. 'linter.exe', and the OS will use PATH
-        #   to resolve that automatically
-        # - We don't check for arrays, see below
-        with when(util).can_exec('my_linter').thenReturn(True), \
-             expect(linter)._communicate(result, ...):  # noqa: E127
+        when(util).which('my_linter').thenReturn('resolved')
+        with expect(linter)._communicate(RESULT, ...):
             linter.lint(INPUT, VIEW_UNCHANGED)
 
     def test_executable_is_set_to_an_array(self):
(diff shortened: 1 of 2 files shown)


kaste:
Reg: #1685 

Here we have separate if branches using `util.can_exec` and `self.which` (which is an alias for `util.which`).  I first confused this and just used `which` because `shutil.which` would have handled both cases. 🙄  I think a refactor in `util` here or later can make this better.

braver:
> We actually only ever expected absolute paths in the 'executable' setting.

It's mostly designed to overcome issues where for some reason the PATH does not have the name of the executable as the linter plugin defined it, mostly in cases where ST fails to load the PATH fully. So, in those cases trying to resolve the executable will fail again, because that's the reason the user used that setting in the first place. 
Now, perhaps another case is that the linter program has a different name on your system (because you're using multiple versions, or otherwise have a non-standard setup). That might also be a good case for this setting. 
Personally I don't think it's a problem to expect the full path even in that case. Just take the ambiguity out of it. The user is experiencing a corner case and probably has that information. 

That said, since we also accept arrays here, I expect `'linter'` and `['linter']` to end up running the same implementation. So if we do something on an array, we should do the same on a string, but perhaps only if that makes sense. E.g. we don't try to find the executable if the path to it is already given unambiguously, like when it starts with `/` or `~/`.

So, a lot of words to agree with you. Just reminding myself of why the feature is the way it is and making sure we're on the same page. Also, merge if you want to, I don't have the setup to test this any better than you probably already did. I might be able to test tomorrow, but not sure.

tests/test_command_generation.py
```diff
@@ -168,30 +168,124 @@ class FakeLinter(Linter):
...
+        when(linter_module.logger).error(...)
+
+        try:
             linter.lint(INPUT, VIEW_UNCHANGED)
+        except linter_module.PermanentError:
+            pass
+
+        verify(linter_module.logger).error(
+            "You set 'executable' to 'my_linter'.  "
+            "However, 'which my_linter' returned nothing."
```
kaste: > "However, 'which my_linter' returned nothing."

This could be better. Can you give me a hint. Like "However, 'which my_linter' returned nothing.\nTry setting an absolute path to the binary. Also see troubleshooting guide."

tests/test_command_generation.py
```diff
@@ -168,30 +168,124 @@ class FakeLinter(Linter):
...
+        when(linter_module.logger).error(...)
+
+        try:
+            linter.lint(INPUT, VIEW_UNCHANGED)
+        except linter_module.PermanentError:
+            pass
+
+        verify(linter_module.logger).error(
+            "You set 'executable' to '/usr/bin/my_linter'.  However, "
+            "'/usr/bin/my_linter' does not exist or is not executable. "
```
kaste: This one is probably okay.  I mean the hint would be most of the time `Look for typos`.

kaste:
Everything you said. 

In the end it is probably about good error messages.  If we decided we never call `which` here, we still would help the users and our issues if we would tell `You set executable to 'x'. However, we expect an absolute path here.`. 

Now here we try `which` because some compile their own linters, or more generally, in the system language world it is more common to have several "same" compiled linters. (g++, c++, ...)

That is probably okay as an enhancement as long as it doesn't change the way we handle the last resort, namely the user gives us an absolute path. (That's the only important part: if all algorithm fail, there must be *one* simple way to tell the computer what to do.)

TLDR; we should just discuss possible error messages.

tests/test_command_generation.py
braver: 👍 A good error message also proposes a possible solution and it's sorely missing here.

braver:
> we should just discuss possible error messages.

totally

braver:
👌🏻

PR #1688, opened 2019-11-08
Refactor `util.which` and the functions around it

### LINKED ISSUE #1684: Remove deprecated `util.communicate`
`util.communicate` is deprecated since v4.3.0  and internally unused. Remove it because we're not a library but an app.

### LINKED ISSUE #1685: Refactor `util.which` to use `shutil.which`

kaste:
Fixes #1684 
Fixes #1685 

These are straight forward refactor-commits, no intent to change any behavior.

### DIFF
--- a/lint/base_linter/python_linter.py
+++ b/lint/base_linter/python_linter.py
@@ -109,7 +109,7 @@ def find_python_version(version):
     # type: (str) -> Optional[str]
     """Return python binaries on PATH matching a specific version."""
     requested_version = extract_major_minor_version(version)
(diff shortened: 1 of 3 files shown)


kaste:
image still possible even after 1 1/2 year.

braver:
The only constant is change 👌🏻

PR #1691, opened 2019-11-11
Ensure file properties in the context come from given view

### LINKED ISSUE #1690: Switching from an unsaved view to a saved view can have inconsistent context
Here I trigger a lint in an unsaved js buffer, and quickly switch to the saved file linter.py

image

```
  Running...

  C:\...\Sublime Text 3\Packages\SublimeLinter-eslint  (working dir)
  > type <buffer 8869> | C:\...\eslint.CMD --format json --stdin --stdin-filename "C:\...\SublimeLinter-eslint\linter.py"
```
Note `type <buffer 8869>` but `$file` gets expanded to the new, active view.

Relevant code https://github.com/SublimeLinter/SublimeLinter/blob/be839b11c32c60ebba0fa398cc9308c5a66f01bb/lint/linter.py#L354-L381

We actually compute `filename` related information on our own because it changes over time.  But unfortunately, if filename is empty I don't erase whatever `window.extract_variables()` has to offer which in this case already reflects the new active view. 

More generally, we pass the `view` from the event handler down but we freeze the context only after the delay time, t.i. in `lint()` not in `hit()`. Possible solution freeze `(view, context)` "immediately" and either do this on the main thread or if on the worker `assert` that both are consistent.

kaste:
Fixes #1690

We use `window.extract_variables()` and took it as-is for the context.
We knew we had to overwrite all file properties to get some consistency,
but we never *erased* possible file properties if the given view is
actually unsaved.

That's a bug which can be exposed *iff* the linter uses e.g. `$file`
as part of its command. The impact is relatively small because at this
time basically no linters report filenames still, and then the backend
just fills in the correct canonical filename by itself.

Typical info in the console looks like

```
> type <buffer 8869> | eslint.CMD --filename "...\linter.py"
```

indicating that buffer `8869` is linted with the name `linter.py` which
is inconsistent.

The fix is to just remove all file related properties from
`extract_variables()` unconditionally.

Note:

Avoiding the race between getting the view (in the event handler) and
actually freezing its context is impractical: We don't want to this
on the main thread because it would a waste. We don't want to abort
if we detect this inconsistency, we really just want the fixed context
in that case which leads to the same code here.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -351,6 +351,18 @@ def get_raw_linter_settings(linter, view):
     )
 
 
+def _extract_window_variables(window):
+    # type: (sublime.Window) -> Dict[str, str]
+    # We explicitly want to compute all variables around the current file
+    # on our own.
+    variables = window.extract_variables()
+    for key in (
+        'file', 'file_path', 'file_name', 'file_base_name', 'file_extension'
+    ):
+        variables.pop(key, None)
+    return variables
+
+
 def get_view_context(view, additional_context=None):
     # type: (sublime.View, Optional[Mapping]) -> MutableMapping[str, str]
     # Note that we ship a enhanced version for 'folder' if you have multiple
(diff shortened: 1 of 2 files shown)


kaste:
This is actually easy to test once you know what's going on. I even added tests. Just merge and forget about it.

braver:
d6650a286f59d211d1387e7db7a09ccf

Issue #1617, opened 2019-06-21
In ST4 test for dangles


kaste:
Pretty sure newer Sublime regresses here. Whenever we reload a view which has regions drawn they become dangles, and also the viewport scrolls to the top nowadays. Also swap line up and down doesn't move the region with the selection but kinda invalidates them by turning them to dangles. Pretty annoying. 

(Easily observable: if you have the panel open, all errors are moved to position 1:1.)

![u7Uj2XVA8x](https://user-images.githubusercontent.com/8558/59923024-9d4f3400-9432-11e9-8d85-1a6302d073f5.gif)

kaste:
Upstream https://github.com/SublimeTextIssues/Core/issues/2877

braver:
Yeah I noticed them too. Not so often, but definitely more dangles than before.

braver:
These are not even labeled yet upstream, so no way of knowing the status. 

Unrelated, but I thought I'd give you a heads-up, if you're not on the Discord channel, that a v4 is coming for Sublime and it will ship Python 3.8. So it's kinda early and under the radar now, but that might offer some opportunities for improvement etc.

kaste:
I think I have a standing record with upstream issues not getting a single comment.

TBH, Sublime may have >1000 "issues" but here in SL we maybe have seen ~3 or so. We just don't have layers of code just to fix "their" bugs.

kaste:
But they sure need a DevRel. If the developers of GitSavvy, SL, or LSP have problems, they should get help.  And they should look at these repos esp. when v4 is probably a breaking change. (I sure don't want pressure v4 suddenly being there and fixing this code base asap.)

braver:
Well, for v4 they are definitely looking at the major packages and it has fixes specifically to improve LSP. So, I'm reading that as good signs. There is a dev release if you want it, but it's kinda alpha and only disclosed via the discord channel, ie. to the hard core group. SL "just works" though, looks like we're all good here. If you want I can dig up the download link for you. It's not a big secret, they just don't want it featured on hackernews right now.

braver:
I changed the title. I feel like we should just revisit this when v4 drops. There has been a lot of work in this area and maybe it got better, maybe it got worse. But it doesn't make sense to do anything right now in v3, or in an alpha release of v4 when it still might change. But it is historically a re-occuring thing so it makes sense to be vigilant.

[...]

PR #1692, opened 2019-11-11
Revert broadcasting `lint_result_changed` cache layer

kaste:
Generally, we can only apply such broad cache layers if the render
functions are "pure", i.e. only draw the highlights if we actually
have different errors from before, otherwise they're already in place.

This breaks for the highlights often because of #1617 or rather the
underlying https://github.com/SublimeTextIssues/Core/issues/2877

I.e. whenever Sublime invalidates all drawn regions (usually because a
file gets reloaded), we can retrigger a lint but still get the same
results as before. In that case we *need* to draw because all regions
are actual dangles but the cache would prevent this. Basically our
error store and the view state are out-of-sync. This is solvable
if we can solve #1617.

For the panel, the cache is just not applicable at all. The intention
was to solve #1623 "for free" in a pragmatic way. But that's just not
the case, and actually we're breaking the `show_panel_on_save` feature.

The usage in `status_bar_view` was never important performance wise
so we can, for now, just revert the whole thing.

Partially reverts 816904c87f8eae497872fc422b07dc43781ff7af
Reopens #1623

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -74,7 +74,7 @@ def plugin_unloaded():
             undraw(view)
 
 
-@events.on('lint_result_changed')
+@events.on(events.LINT_RESULT)
 def on_lint_result(filename, linter_name, **kwargs):
     views = list(all_views_into_file(filename))
     if not views:
--- a/panel_view.py
+++ b/panel_view.py
@@ -81,7 +81,7 @@ def unzip(zipped):
     return tuple(zip(*zipped))  # type: ignore
 
 
-@events.on('lint_result_changed')
+@events.on(events.LINT_RESULT)
 def on_lint_result(filename, linter_name, reason=None, **kwargs):
     # type: (FileName, LinterName, Reason, Any) -> None
     LINT_RESULT_CACHE[linter_name].append((filename, reason))
--- a/status_bar_view.py
+++ b/status_bar_view.py
@@ -44,7 +44,7 @@ def plugin_unloaded():
             view.erase_status(STATUS_MSG_KEY)
 
 
-@events.on('lint_result_changed')
+@events.on(events.LINT_RESULT)
 def on_lint_result(filename, **kwargs):
     if State['active_filename'] == filename:
         draw(**State)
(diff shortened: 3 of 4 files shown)


braver:
So would this bring back dangling highlights, or does it simply mean we don't have a really good solution? I mean, usually not having a cache is a mostly a performance hit, but do we regress functionally here? I don't see any immediate downsides here, but the dangling stuff is quite subtle.

kaste:
We regress with the cache.  

As soon as we get dangles, we *always* need to redraw. Sure most of the time, we have different results from the linter, Sublime reloads the file because you reverted changes or whatever so there is a good *chance* that it has different errors, and then it will draw of course because new errors can't match the cache. But if we have the same results after dangling (you revert something at the end of the file but had import errors on the top)... it is possible that we have dangles *until* the user edits enough so that the linter reports *different* errors. It is not super likely, it surely doesn't happen *all the time* but it can. 

Because it is rather subtle a user will likely roll eyes, feel insecure about the software. You know, it's not right but it heals itself after editing.  Unlikely we will get reports about such things because these are common in this software world. 

tldr; The view state and the store are out of sync after the dangles. T.i. we're not pure 
(in the simple sense of the view reflects the data) and if we're not pure we can't cache. So imo we have to revert (until we solve #1617).  It's that simple: if only *we* write the regions we're safe of course, but if Sublime modifies the view as well and we don't detect it 🐛

braver:
> Because it is rather subtle a user will likely roll eyes, feel insecure about the software. You know, it's not right but it heals itself after editing.

Totally agree here. My point was more that it's hard to test for if you don't know what to look for. The original dangles and zombies we only found after scratching our heads for a few weeks trying to understand what we were seeing. So I find it hard to test what the effect is of this PR. 

The general idea that we cannot cache here is totally clear, so I'm all on board with the intent of the PR. So the question is are you confident here (the changeset is small enough we can basically read it and know that it reverts the right things), or if you want me to also test this for a few days to try and find quirks?

kaste:
Actually, I did not reproduce the bug. I just thought: shit I forgot the dangles! 😁  (I think our store actually knows where the regions/dangles are because in the gif at #1617 it shows that SublimeLinter updates the panel and shows all errors on 1:1 position, which is already super magic and awesome, but our cache doesn't produce a new hash after that.) I just know that I wasn't full-capacity while doing the commit in the first place which also counts.

There is nothing to test here. If there is no typo, we just subscribe to the old event we always used.

braver:
Ok then. Let's just roll with it then.

PR #1694, opened 2019-11-11
Ensure `active_view` is from given window

### LINKED ISSUE #1693: Panel shows last active file from a different window
image

An edge case: On Window 1, **panel must be closed** and the active file must be scratch view, e.g. a GitSavvy interface, something we don't lint. Go to window 2, lint a normal file with some errors.  Go back to window 1, open the panel, it will add the last active file from window (but always stating "No results.").

kaste:
Fixes #1693

We currently only store the `active_view` and `active_filename`
globally; but we can have a panel per window.

This change ensures the `active_filename` comes from the window
we want to draw otherwise we possible show "No lint results." for
a file actually belonging to a different window.

### DIFF
--- a/panel_view.py
+++ b/panel_view.py
@@ -474,8 +474,14 @@ def fill_panel(window):
     if vx == 0:
         return
 
+    active_view = State['active_view']
+    if active_view and active_view.window() == window:
+        active_filename = State['active_filename']
+    else:
+        active_view = None
+        active_filename = None
+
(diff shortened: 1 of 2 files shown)


kaste:
Note to future self: 

On travis `on_activated_async` never runs. Because of that the State never updates to reflect the newly created view. Here the fix is to just "mock" or provide the complete `State` object. (This actually a good solution for this case because we have full control.)

PR #1695, opened 2019-11-17
Log output of a failed command

asfaltboy:
in addittion to the exception output of subprocess.check_output,

I found it very useful to see the full output in debug, as was for me
when checking an error that occurred due to wrong version of python.
Pipenv (or rather it's dependency "click") when run on macos 10.15 is
missing some required "locale" env variables in sublime, which was
fixed in versions 3.7.5 / 3.8. Ref: https://bugs.python.org/issue18378

### DIFF
--- a/lint/util.py
+++ b/lint/util.py
@@ -219,7 +219,9 @@ def check_output(cmd, cwd=None):
         )
     except Exception as err:
         logger.warning(
-            "Executing `{}` failed\n  {}".format(' '.join(cmd), str(err))
+            "Executing `{}` failed\n  {}\nFailed command output: {}".format(
+                ' '.join(cmd), str(err), str(err.output)
+            )
         )
         raise
     else:


kaste:
Hi 👋! 

Actually, I didn't know about `err.output`. I would include it just in the warning message. Don't think users should turn on debug for that. (These are also rare, e.g. `python -V` failing is really uncommon, unlikely we spam the console if we do.)

kaste:
This is unfortunately more complex. Currently I get this output:

```
>>> SublimeLinter.lint.util.check_output(['python', '-Vasd'])
SublimeLinter: WARNING: Executing `python -Vasd` failed
  Command '['python', '-Vasd']' returned non-zero exit status 2
Failed command output: b"Unknown option: -a\r\nusage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...\r\nTry `python -h' for more information.\r\n"
```

You see it's `bytes` and not converted to universal newlines. You can generally use `process_popen_output` right below in the same file for that. Indenting 2 spaces would look better as well. We usually do this in SublimeLinter for a block of content.  

mypy also complains:

```
mypy: error
attr-defined - "Exception" has no attribute "output"
```

so we must access `err.output` with care.

braver:
Is this still going anywhere?

PR #1696, opened 2019-11-20
add discord badge

jrappen:
see commit for details


kaste:
There are no details since the commit message is empty. 🤷‍♂

PR #1721, opened 2020-04-03
Refactor highlights

### LINKED ISSUE #1617: In ST4 test for dangles

kaste:
Fixes #1617 

This is a rewrite of the highlights code. The old code was basically upside-down because it was driven by "*we* control the regions" so we always just render our data. 

But in fact, Sublime controls the regions, it moves them around when we write and edit the code, maybe turns them into Dangles! or even Zombies if we hit *undo*. 

This looks like an example how difficult it can get to untangle a wrong architecture or implementation idea: we have > 40 commits just rewording and juggling old code, old code which was at least mostly working.

Still, this is not done because I know see in the code where we don't handle edge-cases. T.i. we now have a new design space more visible we can tackle later: For example, if we have a warning and an error on a line, resulting in one (red) gutter icon; shouldn't the icon remain or get blue (for a warning) if the user deletes or edits the error? Right now the icon just disappears. 

For the reader or next hacker: Each RegionKey is a proper dataclass `Squiggle` or `GutterIcon`, well not, it is a subclass of `str` because the old code just used strings and holding it as a string is just the way forward while refactoring without breaking too much. (Basically we can still use the key verbatim in `view.add_regions(key, ...)` calls.)


braver:
I try to do no coding whatsoever this week, but will go over outstanding PR’s next week and do some testing.

kaste:
This PR is actually pre-christmas code I forgot 🤷 I was always on that branch so it should probably work.

braver:
I'm basically running with one local branch where I merged everything for the corona sprint. I noticed 2 things:

- I installed an additional linter. It showed up in the status bar automatically and works out of the box. I then uninstalled it again, but it stuck around in the status bar until I restarted ST. 
- Today I had a dangling highlight. A squiggly underline stuck around on the last character of a line, and it no longer belonged to an error. I don't know what I did or how it happened. It's in a really slow setup over a network share, with a not-so-fast linter (phpcs). I would not go away until I re-opened the file. Just thought I should let you know that it happened, I don't know if that information is useful 🤷‍♂️

kaste:
I don't think "uninstalling" was ever supported because we actually don't have `plugin_unloaded` handlers for the plugins.  Package Control actually removes the package and directory, but we have the registered class still in memory, and it actually runs!  I don't know if we can listen for "uninstall" events for "alien" plugins.  If you have the `toggler-addon`, at least "disabling" a package works without restart because the plugin hooks into that. (Disabling moves the plugin to "ignored_packages" which is a readable setting.)

The dangle. Hard to believe.  It is absolutely possible if you switch SublimeLinter branches because the (ever-)store gets corrupted, and if we detect that we just wipe it.  (T.i. I don't migrate the store, neither backwards nor *forwards*. I don't migrate even forwards means if we ship this, the reloader will not save you from either a restart or just re-opening all views, or you're just lucky and don't use "Undo" before the "Undo" stack filled again....)

That said, there are some races still, I don't think you can exploit them easily; but then a typo can make all the planning and thinking 💣; compared to that, the chance to run into migration problems is *very* high.

braver:
> I don't think "uninstalling" was ever supported

Now that you mention it, it does seem unlikely it ever was. And it doesn't need to be. So, disregard that.

> The dangle. Hard to believe.

It is. And I don't really believe it myself unless I see it again and can reproduce it. Really unsure about what caused it. 

Just in general, all open PR's together are really solid. Let's merge them into master and then discuss what other work needs to be done?

PR #1722, opened 2020-04-03
Improve error logging for `util.check_output`

kaste:
Although we catch any `Exception`, it is likely that we have to deal with a `Popen` exception here. In that case the `error` has probably the `output`of the command attached, and we just nicely format it for the user.

Fixes #1695


kaste:
Added tests

PR #1724, opened 2020-04-03
Export our error classes

### LINKED ISSUE #1713: Export `PermanentError` and `read_json_file`
Reminder: Both are used *public* e.g. in the eslint plugin. Proper export them under `SublimeLinter.lint`. 

`PermanentError`and `TransientError` are probably no-brainers. 

About `read_json_file`: Although we have the policy to not export anything beyond core, and reading a json file is clearly a utility function, that particular fn has a built-in cache. Thus plugin authors _should_ really use our function here so that we share that cache.

kaste:
Exporting `PermanentError` and `TransientError` is a no-brainer because
both are control flow exceptions for the linter plugins and used in the
wild.

~~About `read_json_file`: Although we have the policy to not export
anything beyond core since we're not a library, and reading a json file
is clearly a utility function, that particular function has a built-in
cache. Thus plugin authors *should* really use our function here so that
we share that cache.~~

Fixes #1713

### DIFF
--- a/lint/__init__.py
+++ b/lint/__init__.py
@@ -12,7 +12,7 @@ from . import (
 from .const import WARNING, ERROR
 from .util import STREAM_STDOUT, STREAM_STDERR, STREAM_BOTH
 
-from .linter import Linter, LintMatch
+from .linter import Linter, LintMatch, TransientError, PermanentError
 from .base_linter.python_linter import PythonLinter
 from .base_linter.ruby_linter import RubyLinter
 from .base_linter.node_linter import NodeLinter
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -39,6 +39,8 @@ def paths_upwards_until_home(path):
     return chain(takewhile(lambda p: p != HOME, paths_upwards(path)), [HOME])
 
 
+# `read_json_file` is maybe used by plugins. Check `eslint` and
+# `xo` for example.
 def read_json_file(path):
     # type: (str) -> Dict[str, Any]
     return _read_json_file(path, os.path.getmtime(path))


kaste:
In a future, `read_json_file` can be moved to `util.py` of course.

braver:
What is the use case for a plugin to use our function here?

kaste:
Currently `read_json_file` is used by eslint and xo to read the used `package.json` (again). The function is currently local to `node_linter` because only there we have fancy config searching (we only did this for eslint v6 because it is too stupid (?) to climb the directories up although all other linters can do this; actually: I don't know). 

So, `node_linter` finds a suitable `package.json` and reads it by doing so.  We then store its *path* into `self.context['project_root']` but otherwise throw it away. Later eslint (and its sibling xo) had PR's which again read this same `package.json` and iirc look for other installed dependencies (plugins for eslint). 

For these plugins, they either use `read_json_file` with a warm cache at this point, basically getting the contents of `package.json` without *much* overhead. (That's what they currently do.)  Or we attach the actual contents of `package.json` to `self.context` under some key we need to invent. Then the plugins would read just from that basically without any overhead.

Since I either did or reviewed said PR's, I was okay with using `read_json_file` at that time. Would you prefer strictly not exporting functions, instead adding useful stuff to the `context` object.

braver:
I just thought reading a json file is really generic, so if we’re going to provide that feature to plugins there needs to be a real benefit over stuff in the standard lib. The caching is a real benefit, but not as a generic function to read just any json file, but because you want to read some data that the plugin already needed anyway. Storing that data in self.context maybe makes that also easier to understand and attractive for plugin authors. 

I guess it also depends on if you want to lean towards doing a lot of the thinking for plugin authors, or towards giving them raw utils to work with. 

Another consideration might be that some linters use yaml or other formats for configuration that a plugin might have a need for. We could hide how we got that information and provide all plugins a similar way of retrieving it. If that covers the real usecases for read_json_file...

So, there are some thoughts. Hope you find them useful.

kaste:
I'm inclined to not export `read_json_file` then.  Reading any json file can be done with Vanilla Python, using the private function is just a tiny bit edgier and more performant for plugin authors, it can be replaced quickly if we ever break that. 

We cannot foresee the future here, as always, so if there is more need to read other "manifest" data coming from PHP, Python linters, ... we will only learn if the need arises actually. (Just imagine "eslint" would actually get better at automatic discovery, and we would then read the "package.json" just for compat with our own layer. LOL!)

braver:
> I'm inclined to not export read_json_file then. Reading any json file can be done with Vanilla Python, using the private function is just a tiny bit edgier and more performant for plugin authors, it can be replaced quickly if we ever break that.

Agreed. There is maybe no generic need and only one serious usage in the wild. If we ever break ESLint I'm sure we'll hear about it, so let's leave it at that.

PR #1728, opened 2020-04-05
Multiline messages and misc. fixes

### LINKED ISSUE #1716: Allow and support multiline messages
We currently smash "\n" in messages into " " (spaces). Usually type checker produce lengthy, informative messages and we should support their formatting. (I had this for Haskell, but flow, ts, and mypy would also benefit.)

Since the panel already wraps lengthy lines, and with it technically produces multiline messages, I hope this to be a simple change.

### LINKED ISSUE #1725: No popup tooltip for errors with custom types
E.g.

image

This is a haskell "suggestion".  We currently only popup for "error" and "warning" types although we allow any string here.

### LINKED ISSUE #1726: Status bar does not support custom error types
For example, it will print `(ok)` if there are only custom errors; otherwise it just shows (and counts) warnings and errors.

I suggest simplifying this one: We show just `(n)` if we have also custom types in the set of errors. For backwards compat we show `(n|n)` if we only count errors and warnings. We only show `(ok)` if the we really have no reported problems.

### LINKED ISSUE #1727: "Busy" indicator freezes if you open a panel while its spinning
Our cheap "busy" spinner usually freezes if you open the command palette, or for example place the cursor in a panel. You have to activate the "original" view again to get rid of it.

- it stops spinning immediately
- it does not clear after the linter finished

kaste:
This is on top of "refactor-highlights".

Fixes #1716 
Fixes #1725

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -916,8 +916,8 @@ def join_msgs(errors, show_count, width):
             <div>{messages}</div>
         '''
 
-    tmpl_with_code = "{code} - {msg}"
-    tmpl_sans_code = "{msg}"
+    tmpl_with_code = "{code} - {msg_line}"
+    tmpl_sans_code = "{msg_line}"
 
     all_msgs = ""
     for error_type in (WARNING, ERROR):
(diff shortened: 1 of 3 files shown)


kaste:
Multiline example:

image

(Just imagine this was on one line before.)

Status bar example:

image

active_linters_view.py
```diff
@@ -48,8 +49,10 @@ def redraw_file(filename, linter_name, errors, **kwargs):
             State['assigned_linters_per_file'][filename].add(linter_name)
 
         we_count = count_problems(errors)
-        if we_count == (0, 0):
+        if we_count == (0, 0, 0):
             problems[linter_name] = '(ok)'
+        elif we_count[2] != 0:
+            problems[linter_name] = '({})'.format(sum(we_count))
```
braver, line 55: We have a template in the settings for w+e, so we cannot switch to a hardcoded template here: it should be a setting too, and then it might as well account for all three error types.

kaste: Moved to #1730

kaste:
Reverted!

PR #1730, opened 2020-04-09
Honor custom error types when counting for the status bar

### LINKED ISSUE #1726: Status bar does not support custom error types

kaste:
Fixes #1726

~~We now count [w]arnings, [e]rrors and all [o]ther types. Also make "non-errors"
available as the sum of warnings and others. Trivially add "total" for the total
sum.~~

~~The counter template switches to named variables (e.g. `{total}`) from the
limited positional filling. T.i. we switch from a tuple as data format to a
dict.~~

~~The default format uses `non-errors` in-place of `warnings` to avoid the case
that both shown counters are zero but the panel still shows "problems", for
example "notes" or "suggestions".~~

Completely revamped above approach during design phase. Consider reading the full thread.

Redesign status bar widget. Deprecate "statusbar.counters_template" setting. The setting is unused now.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -152,10 +152,9 @@
     // - never: disable this feature.
     "show_panel_on_save": "never",
 
-    // Display counters in the status bar.
-    // The {}'s will be replaced by warnings and errors respectively.
-    // Example alternative: (w:{}, e:{})
-    "statusbar.counters_template": "({}|{})",
+    // DEPRECATED: Not in use! Remove the setting from your User settings
+    // to avoid future warnings.
+    "statusbar.counters_template": "",
 
     // Show the messages for problems at your cursor position.
     // - {message} will be replaced by the actual messages.
--- a/active_linters_view.py
+++ b/active_linters_view.py
@@ -20,7 +20,6 @@ if MYPY:
         'assigned_linters_memo': DefaultDict[FileName, Set[LinterName]],
         'failed_linters_per_file': DefaultDict[FileName, Set[LinterName]],
         'problems_per_file': DefaultDict[FileName, Dict[LinterName, str]],
-        'activated_views': Set[sublime.ViewId],
         'linters_per_file_memo': DefaultDict[FileName, Set[LinterName]],
         'running': DefaultDict[sublime.BufferId, int],
         'expanded_ok': Set[sublime.BufferId],
@@ -34,7 +33,6 @@ State = {
     'assigned_linters_memo': defaultdict(set),
     'failed_linters_per_file': defaultdict(set),
     'problems_per_file': defaultdict(dict),
-    'activated_views': set(),
     'linters_per_file_memo': defaultdict(set),
     'running': defaultdict(int),
     'expanded_ok': set(),
@@ -67,6 +65,16 @@ def on_finished_linting(buffer_id, **kwargs):
     if buffer_id in State['expanded_ok']:
         # Prolong "expanded" state
         show_expanded_ok(buffer_id)
+
+
+def on_first_activate(view):
+    # type: (sublime.View) -> None
+    if not util.is_lintable(view):
+        return
+
+    show_expanded_ok(view.buffer_id())
+    filename = util.get_filename(view)
+    draw(view, State['problems_per_file'][filename])
 
 
 def show_expanded_ok(bid):
@@ -108,26 +116,6 @@ class sublime_linter_unassigned(sublime_plugin.WindowCommand):
 class sublime_linter_failed(sublime_plugin.WindowCommand):
     def run(self, filename, linter_name):
         State['failed_linters_per_file'][filename].add(linter_name)
-
-
-class UpdateState(sublime_plugin.EventListener):
-    def on_activated(self, view):
-        if not util.is_lintable(view):
-            return
-
-        vid = view.id()
-        if vid in State['activated_views']:
-            return
-
-        State['activated_views'].add(vid)
-
-        show_expanded_ok(view.buffer_id())
-        filename = util.get_filename(view)
-        draw(view, State['problems_per_file'][filename])
-
-    def on_close(self, view):
-        # type: (sublime.View) -> None
-        State['activated_views'].discard(view.id())
 
 
 @events.on(events.LINT_RESULT)
@@ -252,3 +240,21 @@ def throttled_on_args(fn, *args, **kwargs):
             action()
 
     return program
+
+
+ACTIVATED_VIEWS = set()
+
+
+class OnFirstActivate(sublime_plugin.EventListener):
+    def on_activated(self, view):
+        # type: (sublime.View) -> None
+        vid = view.id()
+        if vid in ACTIVATED_VIEWS:
+            return
+
+        ACTIVATED_VIEWS.add(vid)
+        on_first_activate(view)
+
+    def on_close(self, view):
+        # type: (sublime.View) -> None
+        ACTIVATED_VIEWS.discard(view.id())
--- a/active_linters_view.py
+++ b/active_linters_view.py
@@ -39,6 +39,7 @@ def plugin_unloaded():
     events.off(redraw_file)
     events.off(on_begin_linting)
     events.off(on_finished_linting)
+    events.off(on_actual_linters_changed)
 
     for window in sublime.windows():
         for view in window.views():
@@ -153,8 +154,7 @@ def redraw_file(filename, linter_name, errors, **kwargs):
     else:
         problems.pop(linter_name, None)
 
-    if actual_linters_changed(filename, set(problems.keys())):
-        set_expanded_ok(filename)
+    remember_actual_linters(filename, set(problems.keys()))
 
     sublime.set_timeout(
         lambda: redraw_file_(
@@ -164,6 +164,11 @@ def redraw_file(filename, linter_name, errors, **kwargs):
             expanded_ok=filename in State['expanded_ok']
         )
     )
+
+
+@events.on('actual_linters_changed')
+def on_actual_linters_changed(filename, linter_names):
+    set_expanded_ok(filename)
 
 
 def count_problems(errors):
@@ -266,13 +271,18 @@ if MYPY:
     U = TypeVar('U')
 
 
-ACTUAL_REPORTING_LINTERS = {}  # type: Dict[FileName, Container[LinterName]]
 ASSIGNED_LINTERS = {}  # type: Dict[FileName, Container[LinterName]]
 
 
-def actual_linters_changed(filename, linter_names):
-    # type: (FileName, Container[LinterName])  -> bool
-    return not distinct_mapping(ACTUAL_REPORTING_LINTERS, filename, linter_names)
+def remember_actual_linters(filename, linter_names):
+    # type: (FileName, Set[LinterName])  -> None
+    previous = persist.actual_linters.get(filename)
+    current = persist.actual_linters[filename] = linter_names
+    if current != previous:
+        events.broadcast('actual_linters_changed', {
+            'filename': filename,
+            'linter_names': linter_names
+        })
 
 
 def assigned_linters_changed(filename, linter_names):
--- a/active_linters_view.py
+++ b/active_linters_view.py
@@ -10,7 +10,7 @@ from .lint import events, persist, util
 
 MYPY = False
 if MYPY:
-    from typing import DefaultDict, Dict, Hashable, Iterator, List, Set
+    from typing import DefaultDict, Dict, Iterator, List, Set
     from mypy_extensions import TypedDict
 
     FileName = str
@@ -19,7 +19,6 @@ if MYPY:
         'assigned_linters_per_file': DefaultDict[FileName, Set[LinterName]],
         'failed_linters_per_file': DefaultDict[FileName, Set[LinterName]],
         'problems_per_file': DefaultDict[FileName, Dict[LinterName, str]],
-        'linters_per_file_memo': DefaultDict[FileName, Set[LinterName]],
         'running': DefaultDict[sublime.BufferId, int],
         'expanded_ok': Set[sublime.BufferId],
     })
@@ -31,7 +30,6 @@ State = {
     'assigned_linters_per_file': defaultdict(set),
     'failed_linters_per_file': defaultdict(set),
     'problems_per_file': defaultdict(dict),
-    'linters_per_file_memo': defaultdict(set),
     'running': defaultdict(int),
     'expanded_ok': set(),
 }  # type: State_
@@ -103,7 +101,7 @@ class sublime_linter_assigned(sublime_plugin.WindowCommand):
         State['assigned_linters_per_file'][filename] = set(linter_names)
         State['failed_linters_per_file'][filename] = set()
 
-        if not distinct_pair(filename, linter_names):
+        if assigned_linters_changed(filename, linter_names):
             on_assigned_linters_changed(filename, buffer_id)
 
 
@@ -191,9 +189,7 @@ def all_views():
 def draw(view, problems):
     if persist.settings.get('statusbar.show_active_linters'):
         bid = view.buffer_id()
-        previous = State['linters_per_file_memo'][bid]
-        current = State['linters_per_file_memo'][bid] = set(problems.keys())
-        if current != previous:
+        if actual_linters_changed(bid, set(problems.keys())):
             show_expanded_ok(bid)
 
         if (
@@ -260,11 +256,28 @@ class OnFirstActivate(sublime_plugin.EventListener):
         ACTIVATED_VIEWS.discard(view.id())
 
 
-PAIRS = {}  # type: Dict[Hashable, object]
+if MYPY:
+    from typing import Container, TypeVar
+    T = TypeVar('T')
+    U = TypeVar('U')
 
 
-def distinct_pair(key, val):
-    # type: (Hashable, object) -> bool
-    previous = PAIRS.get(key)
-    current = PAIRS[key] = val
+ACTUAL_REPORTING_LINTERS = {}  # type: Dict[sublime.BufferId, Container[LinterName]]
+ASSIGNED_LINTERS = {}  # type: Dict[FileName, Container[LinterName]]
+
+
+def actual_linters_changed(bid, linter_names):
+    # type: (sublime.BufferId, Container[LinterName])  -> bool
+    return not distinct_mapping(ACTUAL_REPORTING_LINTERS, bid, linter_names)
+
+
+def assigned_linters_changed(filename, linter_names):
+    # type: (FileName, Container[LinterName])  -> bool
+    return not distinct_mapping(ASSIGNED_LINTERS, filename, linter_names)
+
+
+def distinct_mapping(store, key, val):
+    # type: (Dict[T, U], T, U) -> bool
+    previous = store.get(key)
+    current = store[key] = val
     return current == previous
--- a/active_linters_view.py
+++ b/active_linters_view.py
@@ -19,8 +19,8 @@ if MYPY:
         'assigned_linters_per_file': DefaultDict[FileName, Set[LinterName]],
         'failed_linters_per_file': DefaultDict[FileName, Set[LinterName]],
         'problems_per_file': DefaultDict[FileName, Dict[LinterName, str]],
-        'running': DefaultDict[sublime.BufferId, int],
-        'expanded_ok': Set[sublime.BufferId],
+        'running': DefaultDict[FileName, int],
+        'expanded_ok': Set[FileName],
     })
 
 
@@ -46,21 +46,22 @@ def plugin_unloaded():
 
 
 @events.on(events.LINT_START)
-def on_begin_linting(buffer_id):
-    # type: (sublime.BufferId) -> None
-    State['running'][buffer_id] += 1
+def on_begin_linting(filename):
+    # type: (FileName) -> None
+    State['running'][filename] += 1
 
 
 @events.on(events.LINT_END)
-def on_finished_linting(buffer_id, **kwargs):
-    if State['running'][buffer_id] <= 1:
-        State['running'].pop(buffer_id)
+def on_finished_linting(filename):
+    # type: (FileName) -> None
+    if State['running'][filename] <= 1:
+        State['running'].pop(filename)
     else:
-        State['running'][buffer_id] -= 1
+        State['running'][filename] -= 1
 
-    if buffer_id in State['expanded_ok']:
+    if filename in State['expanded_ok']:
         # Prolong "expanded" state
-        show_expanded_ok(buffer_id)
+        show_expanded_ok(filename)
 
 
 def on_first_activate(view):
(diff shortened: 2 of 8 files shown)


kaste:
Extracted now from #1728 for further discussion.

braver:
That looks like a very comprehensive solution. It’s a lot of flexibility, but any less would seem to require arbitrary choices on our end and leaves the solution incomplete. I like this, it’s even worth breaking compatibility over IMO (and I really wonder why we didn’t do it this way in the first place).

braver:
I like this a lot, it's a good solution. I guess we need to keep support for the old format until we do a big number release, no way around that. We should remember to put a note in the release message about this anyway.

braver:
One thought I had though, was that we do not have a setting for the "(ok)" state. I would like to be able to put a space between the linter and (ok) for instance. Also, you could perhaps set it to an empty string to leave it out (if there are no counters it's ok), or leave out the parentheses or change them to something other than `()`. Now you can do whatever you like with the counters, but not with the "(ok)", and that feels kinda uneven.

braver:
For instance:

```diff
diff --git a/SublimeLinter.sublime-settings b/SublimeLinter.sublime-settings
index 1a5db45..b8d6b96 100644
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -160,6 +160,9 @@
     // Example alternative: ({total})
     "statusbar.counters_template": "({non_errors}|{errors})",
 
+    // Format "ok" (ie. empty counters) in the status bar.
+    "statusbar.ok_template": "(ok)",
+
     // Show the messages for problems at your cursor position.
     // - {message} will be replaced by the actual messages.
     // - {linter} will be replaced by the linter reporting the error.
diff --git a/active_linters_view.py b/active_linters_view.py
index 249fee4..5fc1589 100644
--- a/active_linters_view.py
+++ b/active_linters_view.py
@@ -50,7 +50,7 @@ def redraw_file(filename, linter_name, errors, **kwargs):
 
         we_count = count_problems(errors)
         if we_count['total'] == 0:
-            problems[linter_name] = '(ok)'
+            problems[linter_name] = persist.settings.get('statusbar.ok_template')
         else:
             tpl = persist.settings.get('statusbar.counters_template')
             try:
diff --git a/resources/settings-schema.json b/resources/settings-schema.json
index 098e4b9..311e798 100644
--- a/resources/settings-schema.json
+++ b/resources/settings-schema.json
@@ -138,6 +138,9 @@
         "statusbar.counters_template":{
             "type":"string"
         },
+        "statusbar.ok_template":{
+            "type":"string"
+        },
         "statusbar.messages_template":{
             "type":"string"
         },
```

If we're doing the setting, maybe we should just go all the way down the rabbit hole. Silly as this is, I must admit a setting here would make me happy. 😄

kaste:
It is uneven but with settings this leads to bike-shedding without solving the design problem. (Imagine a website where the user has a setting to have round borders on a specific box.) If the user doesn't like `()` then she uses `[]` or `<>`? We also have the `(erred)` state. In a future refactoring we need the busy spinner here in the brackets as well (currently we run the linters in a sweep per view but the code direction is to run the linters independently, autonomous; we can then tell the user: "mypy" is still running but "flake8" is already ready.)

I also wanted this to be as compatible as possible and actually not let the user know the change at all since we now know there is a maybe solvable design problem here. We may solve it which immediately would result in another breaking change. (I also think and foresee that the design problem is not solvable with templates but with code.  It is similar to the crowded, irritating error panel we had; just settings would not have made the panel better.)

I already dislike "non_errors" tbh. It is confusing, users don't wan't "warnings" and "non_errors" at the same time. The likely combination a user chooses is "warnings" and "errors" which leads to the bug that the sum of both counts doesn't have to match the actual reported problems the panel shows. So "warnings" can only go with "others" but then we would show 3 numbers which is cognitively and visually too much. This rather reduces to "total", "errors", and "non_errors" as variables but they still don't combine t.i. a user can't combine all three or "errors" plus "total". That doesn't make sense outside of bike-shedding and bloat. 

From using this thing over the last year, I actually found `eslint(erred)` and `flake8(ok)` to be the two *really* good states, and that's probably because they are readable. That would mean `(ok)` is inherently better than `(0|0)` or `(0)`. The counter itself is often useful when I don't have clean files, then I look at the initial count, remember that, and try to leave that file edited with the same amount; that happens too often although it is basically a broken setup.

That being said I don't have a better design idea. 🤷‍♀️

braver:
> That being said I don't have a better design idea.

Haha, you're right: this is definitely a good example where we have a setting because the designer gave up. I think it's actually a pretty hard design problem because the design elements are abstract and generic and can be used in different ways for different reasons. That said, by resume tells me I'm a designer, so maybe I should put in some more effort here so we can get rid of the settings.

braver:
The functionality provided by the status bar is a quick summary of the number of errors. The count of errors per file is only available in this view (line tooltips have counts per line). It answers the questions "what linters are running", "is my file ok" and "how bad is it" at a glance. 

Error severity levels (aka "error types") are important here, because in many setups errors break the build whereas warnings (or "infos") don't. Splitting those levels out improves the answer to "how bad is it": lots of warnings maybe isn't so bad but one or two errors is. 

What levels might show up for a given user probably changes from linter to linter, maybe from one project to another. 

A grand total is perhaps nice to have, but it hides information that you cannot get elsewhere. Also, if you have set up different error levels, why do you want a grand total? Just "because we can" is not a good enough answer, I don't see the real use case here.

For levels that don't have errors the count (0) doesn't add much: it's the deviations from 0 that matter. It may improve glanceability to keep the numbers in the same place though. 

For this to work, the information needs to be concise and glanceable, but also unambiguous. We've seen that even avid users of SL (the maintainers no less) cannot remember which number is for errors and which for warnings. 

Cases:
- erred (failure)
- ok (errors = 0)
- errors (errors > 0, warnings > 0, others > 0) 

Proposal:
- `linter (erred)`
- `linter (ok)`
- `linter (e:1)`, `linter (w:1 e:1)`, `linter (w:1 i:1)`, etc.

I.e.:
- We leave out types where the count is zero.
- We prefix the counts so that they are unambiguous.
- We keep the current order (warnings, errors) and add "other" to the end? (Makes more sense to do other, warnings, errors?)
- Use `i` for "others" because it is visually very distinct from `e` and "others" are usually "info" or something similar. 
- A space between the linter and the counters and spaces between the counters because whitespace makes the groups of info easier to read quickly.
- No settings because this is excellent for everyone. 🤞

braver:
It's also fine the parentheses this way:

Screen Shot 2020-04-14 at 22 25 46

Screen Shot 2020-04-14 at 22 23 42

kaste:
Very helpful! Pushing a new WIP state to play with. 

Some words:

- A visual change like this is for a part of the users always a breaking change. It is usually better to make it small and iterate because of the push-back we will receive. I go with `()` parentheses because that's basically the ™️ look right now. (Adding the `e` and `w` and replacing the `|` separator.)

- To better scan the status bar, we move linters having errors to the end; t.i. "ok" linters come first.  (This was personally the most useful or interesting observation you gave: highest severity comes last. Obviously no change if you just run *one* linter for example `eslint` like most user probably do.)

- Give up `erred` and replace it with a `?`. (I also tried `!`.) So, basically `flake8?` if we have problems running the linter. (This idea comes from emacs.) Although it is somewhat cryptic, it stands out very nicely; on the other side it is short, and if the user doesn't want to fix the problem (which I do; e.g. I don't want to setup eslint right now) stays mostly out of the way. 

- Give up `(ok)` per linter as well, just omit it and only show the linter name. (Also from emacs.) (E.g. `flake8 mypy` means both are running and are okay.) 

- Compress the output to a simple `ok` if *all* configured linters pass. (But show the linter names initially, e.g. when you open the file or disable/enable linters.) (This is a tradeoff; if it shows `ok` it is a super clear indicator, very easy to grasp; but you loose the ability to *always* tell which linter is actually running.)

braver:
>  I go with () parentheses because that's basically the ™️ look right now

Ok.

> To better scan the status bar, we move linters having errors to the end

👍 

> Give up erred and replace it with a ?

Smart!

> Give up (ok) per linter as well, just omit it and only show the linter name

🤔 That sounds like a good idea, but I'll have to try and see

> Compress the output to a simple ok if all configured linters pass

I hope the initial flash of linter names is enough. I'm not sure about this yet.

kaste:
The "ok" stuff comes from your paragraph. (Took me some iterations to see and understand.)

> For levels that don't have errors the count (0) doesn't add much: it's the deviations from 0 that matter. It may improve glanceability to keep the numbers in the same place, but if everything is unambiguous and concise, this is of secondary importance at best. 

Basically, if you have a counter `Inbox (12)` you have some unread messages but you usually don't do `Inbox (0)` or "Inbox(💪)" but go to just "Inbox" straight. 

> And what people really want to see is (ok).

That lead to "if everything is ok" the linter names also don't matter anymore, go to just "ok". That also means the status bar info gets shorter and shorter the more it gets green.

braver:
> That also means the status bar info gets shorter and shorter the more it gets green.

Yeah, that's nice, but I think this might be lost:

> It answers the questions **"what linters are running"**, "is my file ok" and "how bad is it" at a glance.

kaste:
Yeah, sure. I had to add the "initially (on open) show expanded" otherwise it wouldn't work *at all*. We could also say "Show expanded for one minute"; or: "Expand if the user opens the error panel". (Typically, I think, you would expand on hover/focus; either literally expanding, animated "divs" on a website, or using tooltips, although tooltips are usually discouraged.)

The question "what linters" could be answered by the panel explicitly. If everything is "ok", the panel shows "No lint results". We clearly have enough room here to say "No lint results. Running mypy, flake8."

braver:
> We clearly have enough room here to say "No lint results. Running mypy, flake8."

Fair point.

For me the flash of linter names before going to "ok" doesn't work reliably right now. Many times it's too fast for me to even notice. And other times I still see the linter name after several seconds.

kaste:
It's the cheapest implementation I could think of, just to get a feeling of how it could be. 😁

kaste:
I also state that it works when enabling or disabling linters. Well, no. 🤫

braver:
To me this does kinda make it unclear exactly what is "ok"? Because I'm me and I'm reading this PR I understand it's SublimeLinter that's telling me it's ok, but that's very implicit.  

Screen Shot 2020-04-17 at 18 14 14


And maybe it's only me, but I find it hard to get used to not being able to see what linters are running. I have so many different projects going on, I don't know what's "ok" if I miss that first flash.

kaste:
Wow, that's a shitty implementation. I'm on an always red project so I did not see how bad this is.

It shouldn't flash at all but stay until you edit and new results from the linters come in. 

You can (easily?) comment the if statement in the render function. I made at least this rather simple because I always switched A/B. Can post later if you don't find it.

kaste:
I think it's at least roughly as advertised now.

braver:
Ok, I’ll try it out next week and report what it feels like.

braver:
So, it isn't a time-based flash, but it's on the second lint "sweep"? It might feel better if it behaved kind of like the temporary indentation detection feedback. But it feels kind of *off* that information (which linters are active) is there when I'm reading or reviewing code, or simply looking up a call signature or something, but it disappears when I'm editing. To me it doesn't improve the feedback whether the file is ok or not (the numbers really stick out when there are error counts), and it saves only a few characters (and actually only saves those characters if the file is clean, *and* when I'm editing). Also, the linter names are unambiguous: they're names of programs I know and recognize. A simple "ok" without any context is ambiguous: *what* is ok, where does that feedback come from, and where is the information that was just there a second ago? 

"all linters ok" might be better if you have a lot of linters running. But then I'd just show that right away, and not flash the linter names at all. 

I like the linter names without counters if there are no counters, that's a great change. It makes a clear statement that the linters are there and ready to catch your errors, but that there isn't anything to report right now. I like that there isn't an implicit state change that changes the fundamental layout of our little piece of the status bar, but stuff is added to it when errors are found. The "ok" is concise and maybe that is all you need to know most of the time, but it's maybe *too* little information and the state change feels wrong to me. 

I guess I don't understand the goal well enough here?

braver:
I *am* wondering if we need to think about manual or on-save-only linters more. Is "linter hasn't run yet" now not the same as "linter has run but reports clean", ie. just the linter name?

kaste:
I actually just coded without further planning but having all our writings open. Reduction to `ok` was the last step, I already had the the other dynamic parts in place. Especially I liked "the red linters come after green linters" sorting, and that the info gets shorter the better it gets. Also probably that the parentheses go away; it's red if you see one, otherwise green. 

Okay, with this we go from

image

to 

image
 
We can clearly just merge this and be done. 

Having this I thought, okay in the first pic, the red state, we state e.g. `mypy` to say "mypy is okay, but flake8 has a warning". In the second, green pic we only say "mypy and flake8 are running". This is of course sometimes useful info but if you're working on the same file, the same project for hours or weeks, you sure know these are running, and actually SublimeLinter is pretty damn stable, which it wasn't as we first released. 

It's like you pay 40$ for Norton Security to get 3 icons in the taskbar. At some point you trust the utility that it runs, and don't need the icons to assure yourself of the investment. So, I thought, maybe the active linters were always also a form of direct marketing: "Hej, here we are."

Of course, it then came from

> And what people really want to see is (ok).

which is really literally what I did here. It is just the clearest form of saying "you're green", just showing the active linters has this somewhat implicit. (Sure I tried "all ok", "all linters ok"; but then this is odd when there is only one linter running, also "all linters ok" would be even longer than "eslint ok". Then why mention that at all? The button is just "green", not "all 3 checks passed" hence you're green).

That being said, I'm not a fan of the current expanding/flipping logic. I started with only having "ok" for the green state but then there was no option at all to see which linters are currently configured. If we had this info in the panel I would have just used that. (I just opened an obscure file in a new project and needed to see where the "ok" is coming from. I only had this question once that day.) Of course when you start the session, or really diable/enable/configure some linter/your project it is also really convenient to see it in the status bar. In contrast: if your in a coding tunnel, one commit after another, you just need the "ok" as easy as possible. 

One metric for the success of the redesign would be: Do I push less often simple flake8 errors to the server?  

> I am wondering if we need to think about manual or on-save-only linters more. Is "linter hasn't run yet" now not the same as "linter has run but reports clean", ie. just the linter name?

Unfortunately, this broke long time ago. We actually only show the linter names after a "sweep" not when "assigned".  It would be okay to fix this with the current "initially expanded" form though, we would show the "assigned" linters on start, and later "ok".  But, you're correct we need to have this in mind. Also: Can we put the busy spinner to the actual slow linter. Or how can we say: flake8 already ready, mypy results may change.

kaste:
So the current logic is, "Show expanded until after the first sweep every time you open a new view (and when changing the config)". Maybe "Show per file type, per window/project, the expanded info until x (and also if the config changes of course)" T.i. the default is the reduced "ok". (Either for n "sweeps" or n seconds.)

kaste:
Tip of the branch now has the timer based reduce to ok solution (like the "indentation detection feedback"). Reduce to ok after 3 seconds but wait for the slowest linter to finish. With this in effect, it is relatively clear where the "ok" comes from, because what we have is basically an animation. Enabling/disabling linters immediately shows the full list of "assigned" linters, again for 3 seconds.

braver:
🎉 At first glance, this is great. I expect that it's just perfect if you know what linters you're running and don't change it a lot. It's a shame there is no tooltip on the statusbar to discover the linters that are active for that view. I do feel like there needs to be a way to discover that, but it's ok if it's kinda hidden. Maybe we can flash the names again if you manually trigger a sweep? (or list them in the panel of course)

kaste:
This would be a nice and usually rather simple addition for SL4 if we had `on_click` and `on_hover` for the keyed status messages

```
view.set_status(KEY, message, on_hover=, on_click=)
```

`on_click` would be very nice to toggle the panel which is otherwise relatively hidden for mousers. `on_hover` would be a function which returns a string or `Html`.  

LSP could also open the panel, GitSavvy could open the status view, MAXPAYNE could toggle the MAX state. 

Ok, this is probably a very good second iteration. (Esp. since we started with my sloppy, just reduce to one number implementation just to fix the bug where we state "ok" although the linters had notes/suggestions.) 

Def the panel should show the additional info.  ("No lint results" is also *very* ambiguous; it can mean "We did not run yet" or "The configured linters did not report any problems"; we also had the lengthy discussion in another thread where a user suggested that we should tell that we actually excluded some errors (via our filter_errors setting): basically "(32 problems suppressed by "filter_errors")".)

SublimeLinter.sublime-settings
```diff
@@ -152,10 +152,13 @@
...
-    // The {}'s will be replaced by warnings and errors respectively.
-    // Example alternative: (w:{}, e:{})
-    "statusbar.counters_template": "({}|{})",
+    // Format counters in the status bar.
+    // Available variables are:
+    // - {errors}, {warnings}, {others} counting errors by their type
+    // - {non_errors} which is (warnings + others)
+    // - {total}
+    // Example alternative: ({total})
+    "statusbar.counters_template": "({non_errors}|{errors})",
```
braver, line 161: let's get rid of  this

active_linters_view.py
```diff
@@ -25,33 +28,57 @@
...
+            sorted_keys = (
+                tuple(sorted(counts.keys() - {'w', 'e'}))
+                + ('w', 'e')
+            )
+            parts = ' '.join(
+                "{}:{}".format(error_type, counts[error_type])
+                for error_type in sorted_keys
+                if error_type in counts and counts[error_type] > 0
+            )
+            problems[linter_name] = '({})'.format(parts)
```
braver, line 81: I think we might as well do without the parentheses here. Go big or go home. It's been a while since we did a big release that changed stuff. Changing this will also signal to users that we made a serious redesign of the status bar info and not just took away their template setting.

kaste:
That's a fucking beast I push here.  Took me several hours to get the behavior we now want to work properly in "manual mode" or for "saved-file-only" linters.

kaste:
Refactored, should work the same. 

Naturally we would put the "busy" spinner herein.  Maybe a short and concise "ok?" if it takes long would be enough.  If we're in "expanded" mode, showing all the linter names, we could just add the same or almost same spinner we already have (since we currently don't know which linter is actually slow.)  (Do I remember correctly that the main graphics problem for the status bar animation were that we cannot expect that the user uses a monospace font. So simple " ", "." switching (blinking cursor etc.) do already fail because the width of the whole text pumps. ?)

active_linters_view.py
kaste: image

image

Well, I rather go home before breaking.  

Just removing the parentheses actually feels like going back to step 0, the design phase.  Since I removed all other delimiters (no ", " between the linters), we start all over again.  "mypy n:2 e:3 annotation w:3"? It is really easy right now to scan for the "("/")" signs.  Especially round parentheses (circles, badges) are very common design lang for this, "Inbox (1)" usually is used for one unread message, whereby "Inbox 1" indicates its name, like there is also an "Inbox 2".

Typical badge, of course with lots of whitespace because the web always uses lots of whitespace, basically the only currently known design language is whitespace: 
image

braver:
The status bar is indeed not monospace. Our entry in the bar comes before others, so if the length of ours changes the entire thing "jumps".

We should not use `?` to mean both a failure and "busy". I like that it's short though. Well, anything we do here changes the length of our piece of the status bar, so anything we do is good as long as it doesn't change constantly I guess. You might as well collapse to `ok` here because in this state everything is up in the air anyway. I don't think the animation is essential, but I do like the clarity of the current "busy indicator": it's very much like "oh it's still busy". 
So, perhaps `ok...?`.

active_linters_view.py
braver: Oh, that's a really good example and a very good reason for keeping the parentheses. I totally agree now 👍

kaste:
Over the last week, this felt very good.  It is so much clearer if the file is ok or not okay.  Because "ok" is so short now, you also just get more dopamine out of it.  This should be better for everyone *if* or *as soon* there is enough confidence or trust in the tools and configurations.  (Probably not perfect for new users though. Still would prefer if the error panel would show the running linters.)

kaste:
> We should not use ? to mean both a failure and "busy".

This would only work in the "ok" state because there is no "failed ok", we always show *which* linter failed "mypy?" means mypy failed.  "ok?" currently has no meaning.

braver:
> Over the last week, this felt very good.

Yes, same here. I regulary have two linters running and it's actually really nice to glance down and just see "hey, everything is ok" instead of having those two separate stats there. 

> Probably not perfect for new users though. Still would prefer if the error panel would show the running linters.

I just set up a new laptop and had to set up the linters again from scratch. I found the first flash of linter names, and of the course the error feedback, to be sufficient to get up and running again. I can imagine novice users, or perhaps situations where linters come and go depending on what your package.json has, where it would be nice to be able to figure out what linters are running... It's the first item in [usability heuristics](https://www.nngroup.com/articles/ten-usability-heuristics/) for a reason: visibility of system status. And we're abstracting and thereby hiding some of that system status. Maybe we're imagining the importance of that information, but I do feel like it's something that the user should be able to figure out somehow (without starting debug mode): "what linters are active right now?". So, the output panel would be good. 
Or perhaps a command that expands the linters in the status bar?

braver:
> This would only work in the "ok" state because there is no "failed ok", we always show which linter failed "mypy?" means mypy failed. "ok?" currently has no meaning.

Sure, but we're making the question mark into an indicator of trouble. "ok + trouble" is indeed meaningless, that's true. But the question mark doesn't mean "wait a minute", I think the ellipsis is a better fit here, to say "more is coming", or "hang on". If it switched from `ok` to `ok...` that seems like a pretty clear "busy" indicator to me. (also use the actual ellipsis character, not 3 periods, might save some horizontal space too)

kaste:
Haha, when you make huge edits, sure I don't get a new email notification. 

The ellipsis char is probably not ASCII, will see.  The "?" is not an icon and as such not just an indicator for something.  "ok?" is an (elliptical) sentence. It could expand into "Am I (still) okay?", "Are you okay?/Is everything okay?", basically expressing doubt: "I'm not sure if everything is okay."  I also read "mypy?" as a sentence, but here the linters are names, so more like "Allen?", typically "Are you there?", usually the interaction got struck, you request something but don't get an answer. The ellipsis is of course to shorten something, "there is more", or to indicate a gap or pause. 

We could have an extra command here. What I don't like if I type at the top of the viewport into Command Palette, and then at the bottom I get the answer.  Then we can just open a quick panel which has the better/bigger font and some HTML preview area if we had an idea what to put there.

braver:
> or to indicate a gap or pause

Exactly! A pause. ... and then there it is.

> What I don't like if I type at the top of the viewport into Command Palette, and then at the bottom I get the answer.

I don’t like that either. We could do an output panel that says something like:

Linters for <file name>:
- eslint
- jshint

braver:
I just had the thought that if we add counters to this panel, there will be the expectation for those counters to be live... so that might be more trouble than it's worth.

braver:
How about we add the active linters as a list to the header of the existing output panel? That way it's easily called up in a way that you may already use. And we don't need to create anything new.

Screenshot 2020-05-16 at 18 03 51

Or like this:

Screenshot 2020-05-16 at 18 05 38

kaste:
Yes!

Likely it is either

```
panel_view.py:  (eslint, jshint)
```

or

```
panel_view.py:
  No lint results. Running eslint, jshint.
```

because the syntax (for the highlighting) seems easier.  I would go for the latter because it's just plain english.

kaste:
I pushed a solution yesterday. Currently our auto-scroller requires exactly two rows for clean files. This is hard-coded. Would be nice if I don't need to change that.  So that's "  No lint results.  Running ...".  I pushed a (ridiculous?) syntax change as well. We *need* a syntax change otherwise it highlights the "8" in "flake8" as a line number 😲🤣. What is actually the scope for ... well just *text*? "entity.name" for the linter names? How would I capture the ", " between them as "punctuation"?

braver:
I would simply capture the entire line as a comment:

```yml
    - match: '  No lint results.*$'
      captures:
        0: comment
```

kaste:
The good old *comment*.

braver:
First off, this is excellent and ready to ship. We can bike shed for months, but let's not. 

Like... maybe not show the linters that have nothing to report?

Screenshot 2020-05-17 at 21 58 20


... 🤔

kaste:
LOL

braver:
🎉

braver:
I guess this concludes the corona sprint?

I'll try and write up some release notes, anything other than this particular PR that needs to be mentioned?

braver:
Oh, you already did the release.. no message to the users this time? I thought the status bar change warranted a message... ah well, next time maybe.

kaste:
Yeah, I really like the silent updates. I think users should subscribe to the releases here in GitHub. Of course I first started to write a message but there is nothing to promote here really. We don't have new commands or settings. Everything is automatic. And the curious can just look it up. 🤷

braver:
I guess I'm the only one left who really likes the update messages from PackageControl 😄 

We should clean up all the old messages one day.

PR #1731, opened 2020-04-09
Continue spinning if the user activates a special view

### LINKED ISSUE #1727: "Busy" indicator freezes if you open a panel while its spinning

kaste:
Fixes #1727

The busy spinner immediately stops spinning if the user opens or
activates a different view. This is only the right thing if the selected
view is a non-special view.

The fix is to check `util.is_lintable` before accepting the new view
as the "active" one. `is_lintable` filters all panels for us.

We switch to `on_activated` (t.i. the version on the UI thread) because
this reduces status bar repaints (flicker) if you really `<tab>` to a
new view while the spinner is showing.

(Basically being "sync" here we avoid one repaint; without it Sublime
would first update the status bar info to match the new view, then we
would kick in and update the spinner; with it we do this in one paint.)

### DIFF
--- a/busy_indicator_view.py
+++ b/busy_indicator_view.py
@@ -6,6 +6,19 @@ import time
 from .lint import events, util
 
 
+MYPY = False
+if MYPY:
+    from typing import Dict, Optional
+    from mypy_extensions import TypedDict
+
+    Filename = str
+    LinterName = str
+    State_ = TypedDict('State_', {
+        'active_view': Optional[sublime.View],
+        'running': Dict[sublime.BufferId, float],
+    })
+
+
 INITIAL_DELAY = 2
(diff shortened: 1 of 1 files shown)


kaste:
Extracted from #1728

braver:
We do the state update in two places here, also on `plugin_loaded`. In theory, the `sublime.active_window().active_view()` on `plugin_loaded` could also be a non-lintable-view: e.g. plugin might take a while to load and in the meantime you may have switched to a panel already or something. So maybe we need that guard here as well?

PR #1741, opened 2020-05-27
ST4074 supports underline styles on white space

### LINKED ISSUE #1740: mark_style: squiggly_underline
**SL Settings:**
The underline styles are replaced with outlines when there is whitespace in the problem region, because underlines aren't drawn on whitespace (ST issue https://github.com/sublimehq/sublime_text/issues/137).

ST issue [https://github.com/sublimehq/sublime_text/issues/137](https://github.com/sublimehq/sublime_text/issues/137) was solved, can you fix that from you side?

braver:
This still prefers outlines over multi-line errors.

If I'm correct we do not need to do anything about the demote logic. It's fully configurable and default "off", so users of newer beta ST versions can adjust if they want.

---
fixes #1740

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -210,9 +210,10 @@
             // - "none"
             // - "fill", "outline",
             // - "solid_underline", "squiggly_underline", "stippled_underline"
-            // The underline styles are replaced with outlines when there is
-            // whitespace in the problem region, because underlines aren't drawn
-            // on whitespace (ST issue #137).
+            // In ST < 4074 the underline styles are replaced with outlines when
+            // there is whitespace in the problem region (ST issue #137).
+            // In newer versions underlines are replaced with outlines when
+            // there are newlines in the problem region for readability.
             "mark_style": "outline",
 
             "scope": "region.redish markup.error.sublime_linter"
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -67,6 +67,9 @@ FALLBACK_MARK_STYLE = 'outline'
 WS_ONLY = re.compile(r'^\s+$')
 MULTILINES = re.compile('\n')
 
+# Sublime > 4074 supports underline styles on white space
+# https://github.com/sublimehq/sublime_text/issues/137
+SUBLIME_SUPPORTS_WS_SQUIGGLES = int(sublime.version()) > 4074
 
 State = {
     'active_view': None,
@@ -293,8 +296,13 @@ def _compute_flags(error):
     # type: (LintError) -> int
     mark_style = style.get_value('mark_style', error, 'none')
     selected_text = error['offending_text']
-    # Work around Sublime bug, which cannot draw 'underlines' on spaces
-    if mark_style in UNDERLINE_STYLES and SOME_WS.search(selected_text):
+    if SUBLIME_SUPPORTS_WS_SQUIGGLES:
+        # Work around Sublime bug, which cannot draw 'underlines' on spaces
+        regex = SOME_WS
+    else:
+        # In newer Sublime versions, we still prefer outlines over multi-line errors
+        regex = MULTILINES
+    if mark_style in UNDERLINE_STYLES and regex.search(selected_text):
         mark_style = FALLBACK_MARK_STYLE
 
     flags = MARK_STYLES[mark_style]


kaste:
How do I install ST4 without affecting my current setup but also carrying all settings and plugins over from my current setup. I admit I fear the break, I hate alpha testing if I don't own the code. I especially need all my own plugins.

braver:
These are the download links:

* https://download.sublimetext.com/sublime_text_build_4074_x64_setup.exe
* https://download.sublimetext.com/sublime_text_build_4074_x64.zip

You can set it up side by side with ST3, with ST3 data in `%APPDATA%\Sublime Text 3` and ST4 data in ` %APPDATA%\Sublime Text` (ie. un-numbered, they're going to drop the major release version numbers). This is the explanation I could find, there may be more, but it's not very well documented right now:

> ... the rule is (on Windows, for example), if %APPDATA%\Sublime Text exists, then ST4 will use that for session files, but if it doesn't, and %APPDATA%\Sublime Text 3 does, then it'll use the latter: this will be the case for most people

For my uses ST4 has been very stable, I've had to make zero changes to how I work or which plugins I use. But I *do* give it a day after each update before I upgrade (the beta's don't auto-upgrade), because sometimes they do break (but it's always caught and patched in no time so far). But I can see how you don't want to go there at all.

kaste:
I installed ST4 and have 3 issues so far. Had no luck disabling almost all plugins, so I have to make a clean install...

braver:
Dang, that's rough. I use almost nothing, and then mostly stuff I maintain myself, so for me it was easy.

kaste:
Auto complete wasn't working in any sane way.  I have SL and GitSavvy on the fresh install.  Auto complete changed like completely.  I think it works as intended.  I actually had "auto_complete: false" in ST3, and usually just used "insert_best_completion" on `TAB` which cycled through all possible completion variants and triggered "auto"-complete manually via `ctrl+space` because I can't have popups while typing.   

It looks like this is just gone. I now have "auto_complete: true" and "mini_auto_complete: true" which is slick but a bit noisier than before, but there is no cycling through all the possible completion items anymore. It behaves like "insert_best_completion" and then be done with it.  Whereby in ST3 it was "take_the_next_best_completion" and next and next.  Odd enough I can't complete import statements, t.i. the to be imported identifiers.

braver:
> Auto complete changed like completely

Yes they completely overhauled it. I was never an AC power user, so I don’t have all the details, but for me it comes up with useful things a lot more often than before, and the UI is much better.

kaste:
Before:
image

After:

image

image

braver:
Great 👍🏻 🎉

kaste:
We're slightly heading into bike-shedding land here. I'm of course in a totally biased mood because ST4 doesn't work. But of course we're underlining spaces here, and there is a reason probably why ST3 decided not to do that, this is undoubtedly a topic in typography.  And I can see that finding or reading underlines between operators (these are typical ws errors for style linters) is harder than before. 

Again, the main problem might be that I need to revert to ST3 til tomorrow morning to get productive again...

braver:
> there is a reason probably why ST3 decided not to do that

I don't remember any reasoning here, but that it was a bug or side effect from the way they used to render whitespace. It was certainly never a feature that when you request to underline a region that ST would refuse to do so.

>  And I can see that finding or reading underlines between operators (these are typical ws errors for style linters) is harder than before.

Sure, a straight underline here is asking for trouble wrt. underscores. So you'd use squigglies, or if you find you run into this a lot you use outlines. And you'd make them a contrasting color to your text (yellow or red) so that they are clearly visible.

The question here is not if underline-like styles are good per se, but if we should mix them up to work around certain problems. I certainly welcome the underlines on whitespaces, in squiggle style. And I like that we're keeping the advantage that we discovered, where multiple outlined regions merge if they cross newlines. 

Again, I don't think the question is if this makes anyone want to switch from outlines to underlines now. The default is still outlines and many are probably comfortable and happy that way. The question is whether this solves the original problem we had with underlines (it does) and if at the same time we don't lose any value we created here (it doesn't). 

It's pretty shitty you have such problem with ST4. It sucks when you can't be productive and have to chase down breakage. We haven't heard anything about how far along the project is, how stable it seems or when it might release. I'm just going to merge this one and we can bike-shed and revisit when at some point in the future there is an ST4 build that works for you. It'll probably be a few more months before this reaches a lot of people anyway.

kaste:
Now that you wrote this I remember I actually styled these typical ws errors from flake8 to be light, low contrast because these are drawn as outlines. 🙄

braver:
It has been said we have too damn many options 😉

PR #1742, opened 2020-06-03
Fix by ignoring

### LINKED ISSUE #1739: Commands to disable errors for a line or file
it was interesting, if the Sublimelinter, suggested options to cover an error or ignore it.

Something of this kind:

Screenshot 2020-05-27 at 11 26 13

kaste:
Fixes #1739 

Implement a command which in turn opens a quick panel from where the user can select a specific error "code" aka rule name violated on the *line* the users cursor currently is.  (<-- one sentence)

The command can be called from the Command Palette "FBI". 

Currently only understands flake8 and mypy errors. 

- [x] just do the fix if there is only one fix
- [x] if the cursor is on one error only select errors and fixes for that error, probably resulting in an immediate fix if there is only one action available then
- [x] for selections on one line only select selected errors
- [x] don't show the help message in the footer if none of the errors actually has a fix
- [x] disable command for multiple cursors/selections
- [x] for eslint, if you select a block, t.i. multiple lines, suggest a disable/enable block
- [x] add stylelint which is basically the same as eslint but maybe has some other domain preferences. For example the preferred mode is probably disable this line here.
- [x] better quick fix label/description text
- [x] sort actions in the quick panel


kaste:
Interesting how this gets complicated: when working with mypy and flake8; mypy ignores must come first.  This is not implemented.  

Otherwise support for multiple ignores is baked in.

kaste:
Can you please push the tooltip mockup code you had aka the design? I'll do the wiring then, should be straight forward.

braver:
The really simple way to go about this would be to do it like this and add a button next to "Copy", which would then trigger the quick panel just like the command. Of course, it's also a bit lame because the list of applicable errors is right there in the tooltip anyway.

```diff
diff --git a/highlight_view.py b/highlight_view.py
index 7120d44..40adebf 100644
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -853,7 +853,7 @@ TOOLTIP_STYLES = '''
         color: var(--yellowish);
         font-weight: bold;
     }
-    .copy {
+    .actions {
         margin-top: 0.5em;
     }
 '''
@@ -862,7 +862,7 @@ TOOLTIP_TEMPLATE = '''
     <body id="sublimelinter-tooltip">
         <style>{stylesheet}</style>
         <div>{content}</div>
-        <div class="copy"><a href="copy">Copy</a></div>
+        <div class="actions"><a href="copy">Copy</a> <a href="ignore">Ignore</a></div>
     </body>
 '''
```


> Can you please push the tooltip mockup code you had aka the design?

Well all I did was add a hardcoded character to an existing template ([highlight_view.py line 940](https://github.com/SublimeLinter/SublimeLinter/blob/fix-by-ignoring/highlight_view.py#L940)): 

```py
tmpl_with_code = "{code} - {msg_line} 🙈"
```

We need to factor in that not all linters will have an action for "ignore". Of course, this gets `html.escape`-d later on. Anyway, that doesn't work.
I pushed a mockup that seems like a somewhat reasonable start to this feature. I'm not in love with it visually yet, but it's not terrible.

kaste:
I only need the design because the href is variadic here because it references an Action object in the end.  Initially I thought we go with the ape but maybe I don't even have that emoticon/image. Do we have hover tips aka descriptions over links?

kaste:
And: what is the best action for eslint. I usually go with disable next line. You probably noticed that the quick panel supports arbitrary many actions per error but for the tooltip we only have one link thus we must choose which.

braver:
I would say disable-next-line is always preferable, and makes the most sense with the interaction we’re creating here.

The ape is a bit much if you have several errors. And I don’t think most Linux have up to date emojis, although 🙈 is actually a really early one and should be there on most systems... 
I’ll work on the design. But you can probably already hook up the command, I’m not going to completely overhaul the tooltip.

kaste:
Ah, I thought the ape is an inline image of the emoticon because I never saw colored emoticons in Sublime Text on my Windows machine.

kaste:
image

I will force-push you're commits because how the escaping is done is already too confusing.  

You added the link to the first line, above I did it on the last line.  First line is also okay, but here design mockups would really help I think because trying different design is not done in a minute. We have generally long messages, or even multiline messages. If we take the link on the first line, maybe we go tabular layout, t.i. the linter name left col, the message middle col, the links right. (Currently we have a two column layout.) And that for sure cannot be done in just 5 minutes.

braver:
Good point. I’ll mockup a design and we can talk about that and do implementation later. I have an idea that might work easily, will let you know later today.

kaste:
You can also checkout again because the links are now working. A bit odd interaction because you can hover over an error without the cursor being on it. We then apply the fix without moving the cursor of course. After that undoing, moving the cursor, editing jumps the viewport back to the cursor.  All that is probably expected behavior, but I felt more comfortable and in control using the quick panel. (Maybe it's just the magic when there is (automatic) editing of the buffer and you don't see a cursor.)

braver:
Because the messages might be long and wrap, I'd like to stay closer to the error code. Usually the ignore comment will include the error code, so in my mind it's all about the code and the message is just gravy.

Screen Shot 2020-06-04 at 20 09 35


And I'd like to try and do something with an icon. ⌫ ("erase to the left") is pretty nice, but we can't be sure what that looks like on OS's we can't test. But I like it a lot. And we have some space at the bottom of the tooltip to explain what it does.
Screen Shot 2020-06-04 at 20 06 40


--- 

I hacked it in, it doesn't work. Also, the additional html breaks the character count and thereby the text wrap. But this is the patch:

```diff
diff --git a/highlight_view.py b/highlight_view.py
index ce8484d..ddad676 100644
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -853,8 +853,16 @@ TOOLTIP_STYLES = '''
         color: var(--yellowish);
         font-weight: bold;
     }
-    .copy {
+    .footer {
         margin-top: 0.5em;
+        font-size: .92em;
+        color: color(var(--background) blend(var(--foreground) 50%));
+    }
+    .ignore {
+        text-decoration: none;
+    }
+    .ignore .icon {
+        font-family: sans-serif;
     }
 '''
 
@@ -862,7 +870,7 @@ TOOLTIP_TEMPLATE = '''
     <body id="sublimelinter-tooltip">
         <style>{stylesheet}</style>
         <div>{content}</div>
-        <div class="copy"><a href="copy">Copy</a></div>
+        <div class="footer"><a href="copy">Copy</a><span> | Click highlighted errors to ignore</div>
     </body>
 '''
 
@@ -970,13 +978,17 @@ def join_msgs(errors, show_count, width, pt):
             first_line_prefix = "{linter}: ".format(**error)
             hanging_indent = len(first_line_prefix)
             if error.get("code"):
-                first_line_prefix += "{code} - ".format(**error)
+                if error["linter"] == 'flake8':
+                    first_line_prefix += "<a class='ignore' href='dinges'>{code}&nbsp;<span class='icon'>⌫</span></a> - ".format(**error)
+                else:
+                    first_line_prefix += "{code} - ".format(**error)
+
             first_line_indent = len(first_line_prefix)
 
             lines = list(flatten(
                 textwrap.wrap(
                     msg_line,
-                    width=width,
+                    width=210,
                     initial_indent=(
                         " " * first_line_indent
                         if n == 0
@@ -993,7 +1005,7 @@ def join_msgs(errors, show_count, width, pt):
             if action:
                 id = uuid.uuid4().hex
                 quick_actions[id] = action.fn
-                lines[-1] = lines[-1] + ' <a href="{}">Ignore</a>'.format(id)
+                lines[-1] = lines[-1] + ''.format(id)
 
             filled_templates += lines
```

kaste:
This looks fantastic. ❤️ I actually love both. Maybe I wouldn't highlight the code, or make it clicky, as we still could use that area to open documentation about the rule.  (Newer eslint now ships a rule url and description.) But that's thinking about a future.

braver:
> Maybe I wouldn't highlight the code, or make it clicky

I like the large click target, just the single character might be a bit small. But it does look tidy like this:

Screen Shot 2020-06-04 at 22 44 46

I like this the best today, an icon is unobtrusive. But maybe it needs the explanation and that seems a bit lame. Maybe I change my mind later that it should be `(ignore)` in stead of ⌫, but that doesn't change the design in an important way. 

What does ⌫ look like on your system?

```diff
diff --git a/highlight_view.py b/highlight_view.py
index ce8484d..cf32010 100644
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -853,8 +853,14 @@ TOOLTIP_STYLES = '''
         color: var(--yellowish);
         font-weight: bold;
     }
-    .copy {
+    .footer {
         margin-top: 0.5em;
+        font-size: .92em;
+        color: color(var(--background) blend(var(--foreground) 50%));
+    }
+    .icon {
+        font-family: sans-serif;
+        text-decoration: none;
     }
 '''
 
@@ -862,7 +868,7 @@ TOOLTIP_TEMPLATE = '''
     <body id="sublimelinter-tooltip">
         <style>{stylesheet}</style>
         <div>{content}</div>
-        <div class="copy"><a href="copy">Copy</a></div>
+        <div class="footer"><a href="copy">Copy</a><span> | Click <span class='icon'>⌫</span> to ignore an error</div>
     </body>
 '''
 
@@ -970,13 +976,17 @@ def join_msgs(errors, show_count, width, pt):
             first_line_prefix = "{linter}: ".format(**error)
             hanging_indent = len(first_line_prefix)
             if error.get("code"):
-                first_line_prefix += "{code} - ".format(**error)
+                if error["linter"] == 'flake8':
+                    first_line_prefix += "{code}&nbsp;<a class='icon' href='dinges'>⌫</a> - ".format(**error)
+                else:
+                    first_line_prefix += "{code} - ".format(**error)
+
             first_line_indent = len(first_line_prefix)
 
             lines = list(flatten(
                 textwrap.wrap(
                     msg_line,
-                    width=width,
+                    width=210,
                     initial_indent=(
                         " " * first_line_indent
                         if n == 0
@@ -993,7 +1003,7 @@ def join_msgs(errors, show_count, width, pt):
             if action:
                 id = uuid.uuid4().hex
                 quick_actions[id] = action.fn
-                lines[-1] = lines[-1] + ' <a href="{}">Ignore</a>'.format(id)
+                lines[-1] = lines[-1] + ''.format(id)
 
             filled_templates += lines
```

kaste:
Looks lovely.

image

kaste:
That's basically feature complete. The rest is refactoring. 

- I'm unbelievable unconstructive with the action descriptions shown in the quick panel.  Please help me out here.

Usually you go tightly integrated in the first round, you just need more examples to get a proper API.  Is there a need even?

kaste:
The show stopper for putting this feature on the plugins is that we have no tests then which is highly problematic for regex based stuff we do here. Regexes always make trouble. 

This would easier if we already had all the patterns we need. E.g. search a line for a given ignore comment and expand it.  This is such a pattern and will be a core feature. But we need some battle testing here.

Please show other linters with documentation links so I can add them here first. And take a look at the descriptions for the quick panel.

braver:
There are actually a few that allow both comments before the line and after. I think whatever we do here needs to include a linter setting, so that you can customize per project which you want. Or maybe you should be able to pick each time, but that complicates the command. I'd say usually you pick one style for a project and you don't mix. 

It seems like a linter should provide a string template for the before-line-comment, and/or end-of-line-comment, and a string to join multiple errors with. 

For those linters where you probably want to re-enable the error later on, we should just only do the before-line-comment and leave the re-enabling to the user, because where to put that second comment is tricky (if you even want to).

This is for pretty much everything in the org (I skipped the deprecated stuff):

**bandit**
https://bandit.readthedocs.io/en/latest/config.html#suppressing-individual-lines
add `  # nosec` at end of line to disable everything

**chktex**
http://www.nongnu.org/chktex/ChkTeX.pdf (6.1.4)
add ` % chktex ##` at end of line, ## is the number of the message to be suppressed
e.g. ` % chktex 9`
multiple: ` % chktex 8 chktex 9`

**clang**
n/a apparently (it's just the compiler with a flag)

**coffeelint**
of course coffeescript is essentialy a dead language
http://www.coffeelint.org/#usage
add `# coffeelint: disable={code}` before the line
and then preferable of course also
`# coffeelint: enable={code}` some time later (I guess the syntax usually allows you to add it to the next line?)
it's not clear how to suppress multiple errors

**cppcheck**
http://cppcheck.sourceforge.net/manual.pdf (6.5) (why PDF??)
support needs to be enabled in the command using the `--inline-suppr` flag
either before the line, or at the end of the line:
`// cppcheck-suppress {code}`
multiple:
// cppcheck-suppress[{code},{code}]

**cpplint**
n/a as far as I can tell

**erlc**
n/a apparently (it's just the compiler with a flag)

**eslint**
covered this already

**flake8**
covered this already

**flow**
https://flow.org/en/docs/linting/flowlint-comments/
either before the line
`// flowlint-next-line {code}:off`
or at the end of the line
` // flowlint-line {code}:off`
not clear how to do multiple, but I'd guess
` // flowlint-line {code}:off {code}:off`

**gcc**
probably n/a: a compiler with a flag

**ghc**
probably n/a: a compiler with a flag

**gjslint**
https://github.com/google/closure-compiler/wiki/@suppress-annotations
essentially a dead project by the way...
before the line:
`/** @suppress {{code}} */`

**golangcilint**
https://golangci-lint.run/usage/faq/
doesn't look like it has anything 

**golint**
can't find any actual docs

**gotype**
n/a seems like

**govet**
n/a seems like

**haml-lint**
https://github.com/sds/haml-lint#disabling-linters-within-source-code
before the line: 
`-# haml-lint:disable {code}, {code}`
and then enable again later
`-# haml-lint:enable AltText, LineLength`

**hlint**
https://github.com/ndmitchell/hlint#ignoring-hints
I don't understand enough about haskell to be able to read the docs...

**html tidy**
http://www.html-tidy.org/documentation/
I don't think it has any, it essentially tries to find parse errors and it doesn't make sense to suppress those errors

**javac**
again a compiler

**jscs**
dead project

**jshint**
https://jshint.com/docs/#inline-configuration
add this before the line:
`/* jshint {code}: false, {code}: false */`
and I guess later on probably
`/* jshint {code}: true, {code}: true */`
the nice/weird thing is that this is function scoped?! 

**jsl**
https://sourceforge.net/projects/javascriptlint/
looks kinda dead to me, can't find any docs

**luacheck**
https://luacheck.readthedocs.io/en/stable/inline.html
before the line:
`--luacheck: no {code}`
seemingly also allows a general disable at the end of a line
` --luacheck: ignore`

**phpcs**
https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file
before the line, or at the end of a line
`// phpcs:ignore {code}`

**phplint**
http://www.icosaedro.it/phplint/index.html
can't find anything in the docs

**phpmd**
http://phpmd.org/documentation/suppress-warnings.html
comment before the class
Maybe this isn't a good match for what we're trying to do here, or that depends on how warnings are reported... I dunno
```
/**
 * @SuppressWarnings({code})
 * @SuppressWarnings({code})
 */
```

**pug-lint**
Doesn't look like this offers anything 

**pycodestyle**
pycodestyle tries to be strictly pep8, there are about 5 types of errors that allow inline ignores with `# noqa`. essentially if you want options use flake8

**pyflakes**
ignores don't make sense for what pyflakes reports about. it also points users to flake8 if you want options

**pylint**
http://pylint.pycqa.org/en/latest/user_guide/message-control.html#block-disables
at the end of a line:
`# pylint: disable={code}`
not sure how you'd do multiple... `# pylint: disable={code},{code}`?

**ruby -wc**
again a compiler

**shellcheck**
https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-one-specific-instance-in-a-file
before the line:
`# shellcheck disable={code},{code}`

**stylelint**
https://stylelint.io/user-guide/ignore-code
probably a lot like eslint?
at the end of a line:
` /* stylelint-disable-line {code}, {code} */`
or before a line
`/* stylelint-disable-next-line {code}, {code} */`

**tslint**
https://palantir.github.io/tslint/usage/rule-flags/
before the line
`// tslint:disable-next-line:{code} {code}`
or at the end of a line
` // tslint:disable-line:{code} {code}`

**xmllint**
don't think so

braver:
Oh, to support multiple errors in a comment you need to be able to read back the comment. That’s going to be tricky. However, it goes against our entire model to have linter specific code in core. If we add a linter-level feature it should be available to all plugins. The regexes should be trivial enough to allow that. Or we find a way for linters to have tests around this in some way. But I don’t feel like we can ship with linter-specific implementations for more than an alpha test.

kaste:
There is some misunderstanding here, at least it reads a bit like we're fighting. I just work on the first iteration, and for that to be useful (to test) we need to support some linters. Basically the linters *we* use plus maybe find the most prominent linters which are interesting because not having them would be lame. (For example, shipping something without eslint support is always lame.  From the conversations/issues we know that rubocopy, go linters, and clang (c lang family) have some traction in Sublime Text. I don't think there is a top php linter.)

So already with this here, we have a strawman implementation of the feature at least for python. The implementation should be deep enough to not suck, t.i. more than just the happy path; that's why I support extending existing comments and do my best to respect given whitespace etc. I don't want rolling eyes when using this feature. 

The feature has two UI's. First a quick panel which can show arbitrary many fixes *per line*. Second, *one* icon in the tooltip *per error* always linked to the best-action-available.  

We need to question if we put that icon near the `code` or: Can't there be a linter without codes with the best action always being "ignore that line" (basically `# noqa`). Then, can we have some standard for the quick panel descriptions. like "Disable flake[E123] for this line." How could we do this for eslint et.al. where we have "disable-line" and "disable-next-line"?

Again, we're still doing the first iteration here that *really* works and doesn't suck. And we want to be sure that we don't make decisions that are just wrong (for example requiring a "code" maybe). (Honestly, it works *really* well.)

Another two things: We make some micro-decisions here which can otherwise lead to an explosion of options. For example, I prefer "# noqa: E123, F234", but "noqa: E123,F234" or "noqa:F234,E123" are equivalent. Likewise, I prefer "// disable-line ..." but "/* disable-line ...*/" is also okay. If we were to handle these variants via options we get in hell. (Here, a first fast deploy would get meaningful reactions that our model is to simple to be useful.)

There are no "linter_settings" here (by design) because we don't have linter *instances* in the views. We only have the linter classes. Everything can be changed in software but we don't have any options available for the end user in this first iteration.

braver:
Sure, we still have a bunch of things to work/think through before we’re anywhere near shipping anything. 
I do want to make sure we’re on the same page about wether this is going to be something that plugins can opt in to, or if everything needs to go through the core package. Now we’re still experimenting so of course it’s just this PR and nothing in the linters themselves. But we _are_ working toward something that can also work for any contrib linter as well right?

kaste:
Sure!

Maybe I should instead of the quick panel use the ... well other panel thing, the new one, for which we have html previews baked in. And in the preview I show how the transformation would look like. Obviously, because it's a preview. 🙄

kaste:
It's maybe coming from "If there is even a need for an API?".  That still holds. We should probably not go full API before we get early feedback and some critique.  I mostly fear the bike-shedding options war we might face. Whitespaces, comment style preferences, "it should be NOQA not noqa, you morons", the best action should be x not y.

braver:
By the way, if I come across as strong headed here, I know I kinda am, but for news features like this I'm also totally up for doing a 180 degree turn on whatever I think right now. 

>  I mostly fear the bike-shedding options war we might face.

Yes, that *is* a concern. However, I'm pretty sure for a lot of "micro decisions" we need to make there is one solution that is best, that I can defend from a UX/human-centered point of view. So, like with previous features where we have been able to find an opinionated solution, I think we should try and find one here too. 

For instance:

> I prefer "# noqa: E123, F234", but "noqa: E123,F234" or "noqa:F234,E123" are equivalent.

Your preference here is obviously correct. The spaces make the statement easier to read for humans. Of course the computer doesn't care and all 3 are equivalent technically. But code is written to be read by humans primarily, and therefore a space after `:` and each `,` is required. Anything less is sloppy or lazy.

For instances like eslint, stylelint etc. where you can choose between before-the-line and end-of-line comments, it becomes murkier. But I think it defendable to opt for end-of-line where possible. Comments in between the lines break into the flow of reading the code. Those comments should be reserved, where possible, for information that adds value for the reader. The fact that some linter warnings are ignored is mostly irrelevant. Now, this might mean some of those linter-disable comments are apparently lost at the end of a long line, or the comment itself might trigger a line-too-long error. And some users might want to opt for a before-the-line comment in those cases. But we're not stopping users from adding comments by hand, we're only trying to help. 

Just to illustrate, that with some effort and discussion we should be able to prevent having settings. I think it's a good approach to add useful features in a simple and practical way, and not to try and act like a portal to every possible feature a linter might have. Maybe that makes us opinionated, but I'm a designer, I'm trained to be opinionated in a way that's grounded in the user experience. So give me some more decisions to make, and I'll try and make them ;)

braver:
> I don't think there is a top php linter.

PHPCS as far as I can tell from colleagues. It's pretty common to use that to adhere to predefined standards (we're using PSR2, PSR12 is more recent). I definitely want PHPCS supported here. 

> We need to question if we put that icon near the code or...

Good point. There isn't always a code, and we should be able to make this work for "disable everything for this line". A good example is bandit.

highlight_view.py
```diff
@@ -989,4 +1029,382 @@ def sort_by_type(error_type):
...
+        view = self.view
+        window = view.window()
+        assert window
+        sel = [s for s in view.sel()]
+        if len(sel) > 1:
+            window.status_message("Only one cursor please.")
+            return
+
+        if not sel[0].empty():
+            window.status_message("Only cursors no selections please.")
```
braver, line 1062: I think it's reasonable to take `sel[0].a` even if the selection is a range?

highlight_view.py
```diff
@@ -989,4 +1029,382 @@ def sort_by_type(error_type):
...
+
+
+class sl_fix_by_ignoring(sublime_plugin.TextCommand):
+    def run(self, edit):
+        view = self.view
+        window = view.window()
+        assert window
+        sel = [s for s in view.sel()]
+        if len(sel) > 1:
+            window.status_message("Only one cursor please.")
```
braver, line 1058: How do you feel about moving these checks into an `is_enabled()` check? This would remove the command from the list. In the context menu it would be greyed out. That doesn't explain "why", but the feedback in the status bar is also very easy to miss.
Of course, eventually we might be able to support multiple cursors as well (ie. "simply" loop through all cursor positions and handle them one by one)?

braver:
Proposal for the quick panel:

I'm thinking the goal is: "just **ignore** this particular **error** right now". We reach that goal by **disabling** a **rule**. This is a UI, so we talk about what the user wants to achieve, not the implementation detail. So we say "ignore this error at my cursor":
Screen Shot 2020-06-06 at 17 20 48


The goal is to let you select the error to ignore. Therefore we list the errors in the same format as we would do elsewhere so you can find the one you're looking for. The format of the comment is less irrelevant at this point, although it *is* a nice forecast of what will happen. Usually though, the effect should be pretty visible.
Screen Shot 2020-06-06 at 17 20 59

I would also propose we simply apply the action if there is only one candidate, ie. we skip the quick panel:
```diff
-        window.show_quick_panel(
-            [action.description for action in actions],
-            on_done
-        )
+        if len(actions) == 1:
+            on_done(0)
+        else:
+            window.show_quick_panel(
+                [action.description for action in actions],
+                on_done
+            )
```

kaste:
FWIW, gcc, clang, rubocop, phpcs, all of them support inline ignoring, everthing or per code/rule.  State of the art in SL land: none of them parse the code in the regex. 🥇

highlight_view.py
kaste: A range (over multiple lines) would be a perfect match to suggest "disable for this block". E.g.

```
// disable
... selected block
// enable
```

Otherwise a range should probably only select the "selected" errors. (Currently we take `.a` but actually expand to the whole line.)

highlight_view.py
kaste: The `status_message` calls are just short hacks.  For example in the case of `status_message("No errors here.")` I usually show the quick panel with an error message and a `noop` callback so that both `ESC` and `Enter` just closes the panel. 

Here, you're correct I wasn't sure if we can support multiple cursors. If not, I would prefer `is_enabled()`. At the moment I don't think multiple cursors or even multiple selections lead to a good experience.

kaste:
About the proposal for the quick panel: I really wanted to offer or allow multiple fixes per error.  T.i. in the quick panel you can for example choose between `disable-line` or `disable-next-line`, or you can do just `noqa` without the code, or maybe for the file. I don't want to restrict the set here for no good reason because of course there will be demand for some flexibility. (It's just in the tooltip we fold down to one action because we only have one icon.) 

"SublimeLinter: Quick Actions" would be very open. (Let's face it if this is an open API, eslint could expose the quick fixes, they get shipped in the JSON, and these are text edits.)

highlight_view.py
braver: > At the moment I don't think multiple cursors or even multiple selections lead to a good experience.

It probably isn't. I usually do prefer trying to support multiple cursors even if it probably doesn't ever happen in the wild, it's just such a corner stone of the Sublime Text experience that everything you can do with one also works with n cursors.

braver:
> It's just in the tooltip we fold down to one action because we only have one icon.

Well, the tooltip can trigger a quick panel to and expose the options. I think we should try for symmetry across the different ways we expose this functionality.

> I don't want to restrict the set here for no good reason because of course there will be demand for some flexibility.

I think we could do something opinionated if that simplifies the interaction. But perhaps for some linters (e.g. flake8) it's already simple because only one format is supported anyway. And then for other linters like eslint maybe we offer all the options (I need to think about the wording):

Screen Shot 2020-06-06 at 22 30 14

This would then list all the applicable errors, or simply apply the comment if there is only one.
The flow for the tooltip would be the other way around: you first pick the error you want to ignore from the tooltip, and then it presents you with this list of options: for the file, for the line, for the line via next-line.

If it's not clear from the text what I mean let me know and I'll mock it up.

> or you can do just noqa without the code

I don't see the use case for that. If you're using a linter at all, why would you ignore everything a linter might warn you about for a single line? I think maybe the only reason for that is because you can't be bothered to look up the error code and to type the entire statement. We make it really easy to not be sloppy like that and add several specific error codes just by clicking some buttons.

braver:
> FWIW, gcc, clang, rubocop, phpcs, all of them support inline ignoring, everthing or per code/rule. State of the art in SL land: none of them parse the code in the regex.

Well, let's fix that ;) https://github.com/SublimeLinter/SublimeLinter-phpcs/pull/51

kaste:
Don't go the route with multiple entries in the main Command Palette. We're not fast enough to enable or disable them.

braver:
> We're not fast enough to enable or disable them.

Ah, didn't consider that. How about this?

flow

(Maybe for the context menu we should do "error where I clicked" rather than "error at my cursor", I'd have to look it up if that's possible)

kaste:
Well, that's a bit overdone and predicting the future.  If we really get a convoluted quick panel in the future we have options but we're not here.  It is also against the core design idea I said okay to.  We generally want one "best-action", and for that we have the icon, and we generally have a quick panel where we naturally have lots of space if and when the need arises. 

- Going from first hover, then click an icon, then switch to a panel is convoluted. (The panels are also keyboard focused UI.)
- We clearly said we don't want complicated (GitSavvy style) panel sequences (wizards). (It is also usually not needed because the quick panel has fuzzy matching, and seeing a couple of entries is no cognitive load.) (I actually redo most GitSavvy panel wizards because these are annoying to use.)

Technically, these require "tagged" actions or fixes. In other words they need a type basically saying "I'm a on-the-same-line-fix", and then a custom type for the rest because there is always a "other" tag. I don't have tags in the design, and don't want to introduce them.  Currently, an `Action` (maybe named `Fix` after refactoring) is a "description" plus a fixer-function. I assume that actions with the same description are the same and can be reduced/filtered.  (For example, in the case we have a only "noqa" capable linter (t.i. without codes), we would reduce possible multiple noqa descriptions/actions to one in the panel.) 

An concrete `fix` is a function maybe of `(Error, View) -> List[Edit]`, maybe it takes multiple errors.  Anyhow, these are generic and very capable and open, combined with that we want an _open_ API "Ignore Error At Cursor" et.al. is no fit here because a fixer can instead of ignoring also apply a correct fix. So while "FBI" is also not future-proof, it is way more playful and humane and correct because we may fix the whole line and probably must support fixing a whole block/selection. In the end, in the future, it may come down to boring "Quick Actions" ("qa", related to "noqa") or "Quick Fixes". 

For "eslint" and disable-this or -next-line.  For short lines we can use "this-line", but otherwise prefer "disable-next-line". No options, always nice, *one* best action. But in the panel, we maybe see that the line or selection has multiple codes, *then* we could suggest "ignore all", one more action, not even applicable to the tooltip. You see, I don't want an explosion of actions and options, that's the main design space, but with the quick panel we always have an option to smooth things out.

braver:
Ok, I think I get your point mostly (not sure I understand everything you wrote there...). 

I think the main thing to take from that design in my previous post is this: symmetry across all actions, regardless if you go through the tooltip, the context menu, or the command palette. If SublimeLinter can help you ignore one or more errors, it shouldn't matter which UI you use. 

I don't like the wizard either. I don't believe we should offer "ignore x for the entire file". And I think we should offer 1 action: either "this-line" or "next-line", but not both. But you asked for flexibility and this is what I think that flexibility should look like. 

> But in the panel, we maybe see that the line or selection has multiple codes, then we could suggest "ignore all", one more action

I don't think we need to do this. Typically, you don't have multiple errors you want to ignore for a single line. Typically, you want to ignore one rule that you cannot adhere to. It's good to have a quick action for that. If for some reason you *do* have multiple errors, you can simply do the quick action twice. That keeps the feature simple. Anything more and the complexity in code and UI explodes for something hardly anyone will ever use. 
Example: I don't see why we would need a quick action that does "ignore E113 and E999". Just ignore E113, and then ignore E999. And "ignore everything" in this case really is just bad practice and we should not offer a quick action for that. 

So, coming back to your original question

> I really wanted to offer or allow multiple fixes per error. T.i. in the quick panel you can for example choose between disable-line or disable-next-line, or you can do just noqa without the code, or maybe for the file

I don't think we should be doing any of that. Just "fix this single error in the best way" is enough IMO. Sure, the quick panel could maybe offer more options, but I personally think that it's easier to understand how SL works, and easier to use overall, if it's the same in the panel as in the tooltip etc.

braver:
Again, I'm not 100% sure about any of this, just thinking aloud. 

I thought some more about this:

> For short lines we can use "this-line", but otherwise prefer "disable-next-line".

What constitutes a short line? I think the primary problem of sticking an "ignore this line" at the end of the line is that that might trigger a new error if the line becomes too long. If "too long" means > 80 or > 120 depends on how you set up the linter. And we don't want to get into the situation where _we_ have to account for how long a line becomes with the addition of a comment.  We can't really even know what the line length limit is. But it's kinda lame if the action to ignore one error triggers another. 

So maybe there is no choice but to offer the choice to the user here. And that would mean offering that choice also in the tooltip. 

Or maybe it is not that big of a problem because now we made it so easy to ignore errors.

kaste:
Well, that's all easy solvable. Short lines are actually really short because the eslint comments are always or easily over 30 chars. You don't put such a comment at eol just to exhaust your theoretical max line length. Nobody sets a wide max line length and really uses it for the essential code which is conditionals et.al. We set a wide line length to allow some lines usually strings esp. urls... How many and what lines in SL are over 80? We also actually know how long the line will be after adding the comment. Tldr; it should be easy to make the code better and avoid an option.

The noqa example or the possible action to suppress multiple errors in one go... Don't take that for python literally. We may have none noqas in SL at all. You don't use noqas with flake8 that often, most errors you really just fix. But this might be different in other languages. I remember js when you have short shim sections in the code written in es1999 style basically in violation to the other code. ... This isn't necessarily important for the concrete code we might ship as first version, it is some loud thinking of how we shape the API. 

Mostly we should focus on the current features. Are they good, good enough to ship, are there rolling eyes interactions? Imo both python fixers are good enough, and beyond completely naive because we can extend and merge multiple errors. 

For eslint I would appreciate a disable on this line for really short lines. There is a bit rolling eyes if the line only has 30 chars, maybe being a simple assignment. 

Todo otherwise:

- just do the fix if there is only one fix
- if the cursor is on one error only select errors and fixes for that error, probably resulting in an immediate fix if there is only one action available then
- for selections on one line only select selected errors
- don't show the help message in the footer if none of the errors actually has a fix
- disable command for multiple cursors/selections
- for eslint, if you select a block, t.i. multiple lines, suggest a disable/enable block
- add stylelint which is basically the same as eslint but maybe has some other domain preferences. For example the preferred mode is probably disable this line here.

braver:
> Short lines are actually really short because the eslint comments are always or easily over 30 chars

Fair enough. Nobody keeps their JS at max 80 chars, 120 is more likely. Same with PHP. Even in python land 80 is rare, but then the comments are also really short. A typical noqa (`  # noqa: E221`) is 14 chars.

So, pick a number... 120 - 40 = 80 chars. 
Or if you're doing strict 80 chars python, minus 15 chars for a noqa + code = 65 chars.

So let's say a "really short line" is 65 chars? Seems reasonable to me. Maybe make a setting here so you can tweak it to preference and we don't have a magic number in the code.

> add stylelint which is basically the same as eslint but maybe has some other domain preferences. For example the preferred mode is probably disable this line here.

yeah, lines in a stylesheet are typically very short, unless you're some asshole who puts multiple properties on one line 😉 

> Mostly we should focus on the current features. Are they good, good enough to ship, are there rolling eyes interactions?

I think the implementation is pretty good indeed. The only thing I have a problem with really is just the quick panel. It should really list the errors, not the fixes, because I don't know what 221 or 111 is off the top of my head, but I do know for instance that I want to ignore the multiple-spaces error. Basically we need this: https://github.com/SublimeLinter/SublimeLinter/pull/1742#issuecomment-640078423 (second screenshot).

kaste:
I made some progress.  I have the following problem with the quick panel descriptions. 

Part of the domain here is that we can have multiple *same* errors, t.i. with the same error code/rule on the same line, or in the same selection.  E.g.

```
import foo, bar

panel output:
167:8  flake8:F401    'foo' imported but unused
167:13 flake8:F401    'bar' imported but unused
```

There is obviously only one action here, namely to put `# noqa F401` at EOL, which will in turn remove *both* errors. That's no problem, I just combine/filter here, we do line based or maybe for some linters block based "ignoring/disabling" so some post processing of the possible actions is just .... Done.

For the description, I then currently only show "Disable F401" or something like that because these are the invariant parts here whereby the actual messages of the errors differ in "foo" and "bar". (For "unique" actions, I show "Disable F202 too much whitespace" (not literally but I show the message plus the rule etc.)  Any better idea? 

Also, the error message can be just too long, not in flake8 but think of typescript or mypy. Generally this is pluggable atm, not all linters have to agree on the same output format.  E.g. 

image

This is already too wide, and Sublime adds `...` in the middle here not really helping. Just "Disable [return-type]" is rather short, sure a truncated message doesn't help either. I mean you literally can't read it, it would rather function as a hint.  Should we take the effort and trim here at our own after 40 chars? Omit it?

Can we have coherent descriptions at least for the three linters we currently only support?  (For example, for python I have flake8 and mypy errors in the same view all the time, would be nice if the descriptions would look homogeneous of course.)

PS Generally, very nice feature we already have, didn't noticed that I actually write these disable comments that often.

TLDR; The only info we always have is the rulename plus linter name.  If we just use "Disable [no-return]", "Disable F304", everything looks similar, coherent, even if we combine/merge/filter multiple actions. Workflow would be anyway look at the error in the output panel or the tiny status bar, then decide to disable, then open the quick actions panel. At this point the decision has been made, and either the status bar or the error panel are still open showing the error info. 🤷

braver:
> Generally, very nice feature we already have, didn't noticed that I actually write these disable comments that often.

Same here!

We really need to include the error message, because it’s not user friendly to have to either remember or look it up. Don’t make me think.

Do we really need to deduplicate? From the functional perspective it makes sense, but for the user not necessarily. If we keep the deduplication, we need to add a little explanation, and we can still keep one of the messages as a reminder/hint. Maybe 

```py
Ignore {linter}: {code} - 2 errors, e.g.: {msg}
```

And then clip the entire line at 80 or something.

braver:
All this is mostly for flake8 and such with numbered error codes. For eslint or stylelint, or phpcs, the code is usually descriptive enough.

kaste:
I just push the WIP.  Take a look at how it behaves if you select multiple errors, multiple lines.

braver:
Oh, that's really sweet! I love how it handles a selection across multiple lines. With some effort we can maybe do the same for multiple selections?

I also see now how deduplication is necessary here. I do think it would be nice if we could announce in the quick panel how many errors are covered by each fix. 

I think the list in the quick panel would be more neat if you sort the entries alphabetically. That would also make entries for each linter sit together, which probably looks better.

kaste:
Iirc I sort by error count then alphabetically. But not having the error count displayed. 🙄

But how would we go about it. "solves 2 problems". "solves 3 errors on 2 lines". Quick panel entries are shit to design really.

braver:
I’d phrase it like “solve [E113] 3 times”:

```py
Ignore {linter}: {code} (2x) - {msg}
```

Why sort by count first? Because that’s mostly likely the one you want?

kaste:
I think I thought about only showing the "common" actions when selecting multiple lines. Basically the idea was to reduce the actions to only common ones if there are any, in the hope it would reduce to one action which then would be applied automatically because if there is only one action to show we just apply it. However, I never implemented this.

kaste:
This is now a real showcase PR.  I tried "block commenting" and used eslint as an example.  I also added some "real" fixes for flake8 to see how this would work, and how much code this would be.  The only thing missing, what I would like to see, is using the "preview" functionality of the `ListInputHandler` to show the actual changes a particular fix would do to the file.   

In general the API is okay, and anyone could just use it straight in any plugin.

E.g. for a real fix:

```python
@fix("flake8", {"E262", "E265"})
def fix_e262(error, view):
    # type: (LintError, sublime.View) -> Iterator[TextRange]
    yield TextRange("# ", error["region"])
```

You can put such a snippet in the User folder if you like.  The error here is missing whitespace in a comment for example `  #foo`, and literally we replace the squiggled error region with the literal `# `.  

Inline-disable or ignore fixers are registered with `@quick_action_for_error` which is a misnomer but anyhow.  

```python
@quick_action_for_error("mypy", except_for={"syntax"})
def fix_mypy_error(error, view):
    # type: (LintError, sublime.View) -> Iterator[TextRange]
    line = line_error_is_on(view, error)
    code = error["code"]
    yield (
        extend_existing_comment(
            r"  # type: ignore\[(?P<codes>.*)\]",
            ", ",
            {code},
            line
        )
        or maybe_add_before_string(
            "  # ",
            "  # type: ignore[{}]".format(code),
            line
        )
        or add_at_eol(
            "  # type: ignore[{}]".format(code),
            line
        )
    )
```

This is basically straight forward, and since these are only functions it is always easy to compose them, t.i. maybe to make them even simpler for the end-users. Note the `except_for="syntax"` because we never want to ignore syntax (or some other) errors, this is baked in already.

One interesting point might be that these fixers *must not* modify the view, they just return descriptions of necessary edits.  That's a feature not a restriction, and should be okay with todays programmers.  (That's not OOP here.)

If we potentially have different kinds of fixers (Ignorers and RealFixers) we probably need to tag the `Actions` so that we could show different icons in the tooltip for example, or in general for sorting and filtering.  That's doable. 

Now, I don't want to ship with 4 real fixers for flake8 because that's probably more confusing.  But these are just damn useful.  I've since added a key binding `ctrl+k, ctrl+f`, and that's insanely fast and good to use, esp. *if* there is only one action applicable, t.i. the panel doesn't even show up. 

Ok, enough said, this is all very useful already, and very open.

braver:
Just before I dig in further, are you proposing to actually implement custom handwritten regex based fixers? I.e. not just the “ignore this thing” comments, but actually modifying the buffer to fix an error? I’m strongly opposed to this approach. For all languages that have linters that are any good, there exist fixers that operate on the AST, and usually also formatters, and 9 times out of 10 there is a very good LSP implementation. We should not expect that linter plugins for SL to write fixes by hand. It will only cover a small subset of possible errors, and will always be lame compared to other available options. If this were a good way to build fixers, which I don’t think it is, who is going to write and maintain them?

If we want to support fixers, and I’m interested in doing this if done properly, we should take a similar approach to how we do linters: we pass a bit of code (a file, a buffer, a part of a file perhaps) off to an external process and let that do its job. Maybe we can update the buffer in a smart way to make it flow well, which has always been the blocker to go down this path.

braver:
I’d definitely need to see more than that trivial implementation for “real fixers”. There is a wild number of possible style errors you can configure in eslint. How do you propose someone add their own fixers for that? Also, again, why? Eslint can fix most of those already and it will read your config so it “just works”.

braver:
Eh I don’t want to be overly negative here. It’s a showcase so we can talk about stuff right?

Can we spin the fixers out to an add-on so it doesn’t blur the lines of what SL is supposed to do for people? 

I get the usefulness here, it’s like magic when it works. But I really don’t see how this is going to solve a significant portion of the formatting issues I run into daily across several languages and projects. Usually there is a fixer/formatter available that is also aware of the requirements specific to each project, and I just need a better way to run that. Not a new way to create fixers myself. 

The ignore comments are 🔥 though. Can I try and hook some up for phpcs? RE: https://github.com/SublimeLinter/SublimeLinter-phpcs/pull/51

kaste:
Lol, you're focusing on the wrong end.  I implemented the real fixers to showcase what's doable with the API.  The low level API is how we register fixers or quick actions, everybody can do this without touching core or plugin code. Then I show that by limiting to return or yield edit descriptions we basically have an open system for fine buffer modifications. This also seems solid and error free. At this point it looks like it's fast and without user annoyances like jumping cursors or glitches. It also doesn't crash. These are the important parts. The feature has to work without rolling eyes, and it must be open and extensible.

The 4 real fixers are also actually impressive because there are so fast, glitchfree, undoable. The eslint pun, well eslint ships the fixes in the JSON response...

Anyway, I think I'm okay with the low-level system design. Here the question was if I should straight from the start tag the actions. I'm also interested in using the preview area but had no time to try this.  I don't have time to implement actions for other linters especially linters I don't use atm. I would probably toy with flake8 real fixers but I don't have time for that either really. But I usually go the extra mile for flake8 at some point, maybe Christmas.

The UX is also probably okay.  I wonder if we should just reduce the available actions aggressively. For example, if the user selects a region just fix everything in it.  The user can still undo or edit manually.

braver:
> Lol, you're focusing on the wrong end

Ok! 😅 I was hoping I was. It's a good showcase of what the API allows, that's for sure. 

> I wonder if we should just reduce the available actions aggressively. For example, if the user selects a region just fix everything in it.

I am also thinking in that direction. 👍  If you want fine grained control over what to fix you go via the tooltip. All the other commands just fix/ignore everything in the current selection. It's not like you'll have 10 errors and want to selectively ignore a few of them. Typically you have zero errors except for one or two you can't/don't want to fix.

lint/quick_fix.py
```diff
@@ -0,0 +1,597 @@
...
+        view = self.view
+        window = view.window()
+        assert window
+
+        if not quiet:
+            if len(self.view.sel()) != 1:
+                window.status_message("Not implemented for multiple selections")
+                return
+
+        sel = view.sel()[0]
```
braver, line 65: In stead of requiring 1 selection and having the kinda ugly `quiet` parameter here, would it be feasible to just loop over `view.sel()` and foreach naively? I believe ST prevents having overlapping regions here so there don't seem to be any crazy corner cases to worry about. My first very simple attempt at doing a loop here doesn't work, but what do I know? 😄 

Alternatively, we could still remove the `quiet` parameter. Use `is_visible` in stead of `is_enabled` and simply return `return len(self.view.sel()) == 1` to prevent the command showing up in the palette. And when the command is still called with more than one selection, e.g. via a keybinding where `is_visible` doesn't come into play, we can throw the status message.

Default.sublime-commands
```diff
@@ -24,6 +24,12 @@
         "command": "sublime_linter_toggle_highlights"
 
     },
+    {
+        "caption": "SublimeLinter: Fix By Ignoring",
```
braver, line 28: I'm still partial to the phrase "SublimeLinter: Ignore Error at Cursor", because ignoring `!=` fixing, and I think it's clearer about what's going to happen.

I'd also add this to the context menu, because it makes sense to have it there "in context":

```json
[
    {
        "id": "sublimelinter",
        "caption": "Linter",
        "children": [
            {
                "id": "sublimelinter-lint",
                "caption": "Lint This View",
                "command": "sublime_linter_lint"
            },
            {
                "id": "sublimelinter-fix",
                "caption": "Ignore Error at Cursor",
                "command": "sl_fix_by_ignoring"
            }
        ]
    }
]
```

lint/quick_fix.py
```diff
@@ -0,0 +1,597 @@
...
+            "",
+            solves=[error]
+        )
+        for error in errors
+        if error["code"] in only_for or not only_for
+    )
+
+
+# import zoo, foo
+# import bar
```
braver, line 266: 🤔

braver:
Pretty sweet, I was quickly able to set up a fixer/ignorer for phpcs:

```py
@quick_action_for_error("phpcs")
def fix_phpcs_error(error, view):
    # type: (LintError, sublime.View) -> Iterator[TextRange]
    line = line_error_is_on(view, error)
    code = error["code"]
    yield (
        extend_existing_comment(
            r"\/\/ phpcs:ignore (?P<codes>[\w\-\/.]+(?:,\s?[\w\-\/.]+)*)",
            ", ",
            {code},
            read_previous_line(view, line)
        ) or insert_preceding_line(
            "// phpcs:ignore {}".format(code),
            line
        )
    )
```

Because the error codes here are super long, the comments would easily push a normal line over 120 characters, so I opted for `insert_preceding_line`. In this particular case I don't see the need to offer options here or whatever, this is excellent.

I think this interaction is also really ok:
Screen Shot 2020-06-27 at 19 41 43
We *could* simplify, but I don't think I want to. The intended interaction is never to select a lot of lines and then say: automagically ignore every error in here. Usually linters have a different way to ignore errors for a range of code (e.g. "disable" at line A and then "enable" again later at line B). What I think we should focus on is, hey I have a specific error in here that I can't/don't want to fix, I ignore that one for me. And that's what this already does really well.

kaste:
Yeah, each fixer is kind of a regex hell because really they should try to `extend_existing_comment` (probably no other editor does this, iirc VSCode does not). But otherwise it's straight forward.  

It is really not necessary to further reduce the actions for the first release.  We're not a team here, we're only two people.  We can't ship a completely shaped up feature.  In some sense we're (maybe already) overdoing it here.  (That's also different from the lenghty status bar update PR. There, we had actually looked two years at the same status widget so we just had more experience what it actually is and does.)

lint/quick_fix.py
kaste: LOL, you didn't see all the fixups for the ws errors I initially committed. You need some test area.

lint/quick_fix.py
kaste: Actually, `view.sel()[0]` is hardcoded in the code. T.i. the code somewhere assumes there is only one selection. 

Didn't know the `is_visible` thing but I actually don't want to show the quick panel for "No actions." as well.  We try to do this in GitSavvy, and I like it.  For strict key bindings try to flash the status bar with a message to not disturb; on the other hand, if the user is in the panel area, for example using the Command Palette, open a noop panel with an explanation. (Although the `quiet` etxra argument is a hack here, I never needed in GitSavvy.) (An example would be trying to push: if you use a keyboard shortcut, just flash "Already up to date", but if you use the panels display the same info in a noop quick panel.)

braver:
> In some sense we're (maybe already) overdoing it here

Yeah, probably. Everything is mostly guess work now, I need to use this for maybe a year and then come back to it. 

So, what do you want to do? Merge this as it is and announce it as a beta? We can maybe add a few other cases as people request them. And then when we're sure about what we need we can redesign the API and only _then_ expose it to the community of plugin writers? 

There are some UI things I want to try when you're basically done here. So if we merge maybe also give it a few weeks before we tag a release?

Default.sublime-commands
kaste: For the context menu, we had an issue that we should *not* add anyting to it.  And although we didn't follow the suggestion "for reasons", I'm actually behind it.  We should not mess with the context menu IMO. (And for example Sublime Text itself should *not*, **really not** add all these Sublime Merge ads here.  It's not adware or shareware.  What huge stupid decision is that?  What users think of other software doing this?) Instead we should add this to the README.

"Fix by ignoring" was always only playful for the draft version here.  But I'd go with the neutral "SublimeLinter: Quick Actions".

braver: > We should not mess with the context menu IMO

Fair enough. The thing is nog enough as it is. However, we already have an entry in there, and this expands that single entry to expose multiple features. So, it's not making the situation worse IMO.

I don't think the context menu is an important part of interacting with ST, but I think this entry belongs there. Or we backtrack and remove the `Lint This View` entry. I could get behind that for a big release like we're planning here (this is at least a 4.5 if not a 5.0).

I totally agree about the Sublime Merge thing by the way. At least have the decency to group them so we don't have 4 entries sitting there doing nothing. 🤷‍♂️

kaste:
We should at least close merge this to a dev branch because the conversation here is unproductive. 

Sure I can rename all day long, and it is also kinda important because these top level decorators will eventually land in user land. At least we need to name the main Command otherwise users need to fix their key binding at some point.

Well, we probably should ship a binding to Ctrl+kf.

braver:
> Well, we probably should ship a binding

We shouldn’t ship a key binding IMO. I’m even more opposed to packages shipping with keybindings than context menus. They tend to clash because there aren’t enough keys on the keyboard for all packages. They’re also hard to unset and even harder to debug. We _should_ suggest something in the readme though, and maybe put some comments in the keybindings file that people can copy paste.

Let’s make a 5.0 branch and merge to that. We can make some more PR’s over the coming weeks to finalize everything. I can do some renaming on user-visible stuff, write some docs, etc.

kaste:
I pushed to `dev`.  Note that this not for long-living. It's just that this PR with 70 comments became unreadable. 

I don't see why this is a "5" if that matters, it's 4.16, we don't deprecate anything here. It is addition only, and unlikely we ship a full feature either.  (The multi files feature last year was more work honestly.)  

I also don't need to test this any further.  It never crashed from the first commit on.  And the more subtle usability issues only come with real time.  I hope I don't have to see the quick panel in real life that often, and expect when testing it never popped up.

braver:
Yeah, the numbers are just marketing anyway, even if we follow semver for whatever reason. But I always feel a big number justifies taking features away and you know how much I like to remove features 😉

I just need the dev branch so I can open a PR for some docs, UI tweaks, etc. It’s certainly stable enough to ship as it is.

PR #1744, opened 2020-06-10
Fix: Use "outline" again as fallback in ST3

kaste:
Well, #1741 is basically wrong on every line, don't use drugs and then
code.

- The Sublime change has been introduced *with* v4074 so it's `>=`
- The conditional must be flipped

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -65,7 +65,7 @@ SOME_WS = re.compile(r'\s')
 FALLBACK_MARK_STYLE = 'outline'
 
 WS_ONLY = re.compile(r'^\s+$')
-MULTILINES = re.compile('\n')
+MULTILINES = re.compile('\n(?=.)')
 
 # Sublime >= 4074 supports underline styles on white space
 # https://github.com/sublimehq/sublime_text/issues/137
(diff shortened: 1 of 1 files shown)


braver:
Since we're still on this topic... I pushed a tweak. MULTILINES now matches errors reported just at the end of the line. That is, they don't actually cover multiple lines. Example:

Screen Shot 2020-06-17 at 22 15 52

So, MULTILINE means there is actually something _after_ the newline. This the same error, now with the lookahead in place:

Screen Shot 2020-06-17 at 22 15 42

I checked that it still works as intended for typical errors where eslint highlights entire blocks. I actually think it's kinda odd that it works this way. We're highlighting after the last character, and apparently there's a newline character there. Maybe it wouldn't work this way outside Sublime Text, who knows. 🤷

kaste:
That makes sense.  I'm a bit confused why you used a lookahead here although it shouldn't matter. I can't think of any edge case where this would be wrong either, so we merge.

braver:
Oh, a lookahead is perhaps a bit too fancy here, we’re not capturing anything here. Ah well... there’s too many ways to do something.

PR #1752, opened 2020-06-28
The Quick Actions feature

kaste:
Let's keep this as a TODO list

- [x] Undo "real fixers" for flake8 I accidentally merged


braver:
Cool, looks like we're ready to roll here. Great stuff. 🚀

braver:
I use this from time to time, not intensively but just real day to day work.. and it's perfect all the time 🚀 . Let's just ship this and let other people play with it. And complain that we need to support more linters. 😄

kaste:
I use it every day once.  It is super helpful esp bc these "rules" or "error codes" are so hard to type.  (Esp. the flake8 codes; and flake8 is *permissive* with wrong codes ignoring everything or so 🙄.)

You for the release notes.

braver:
Added some notes and a new screenshot:

messages/4.16.0.txt
```diff
@@ -0,0 +1,9 @@
+SublimeLinter 4.16 - Quick Actions (beta)
+
+We're building a new feature to quickly ignore specific error messages.
+At first we will ship actions for eslint, stylelint, flake8, mypy and phpcs.
+
+Want to see actions for your favourite linter?
+Please open an issue so we can investigate it.
+
+Eventually this will become an API so that plugins can add their own.
```
kaste, line 9: Well, yeah, LOL, that's too short.  We should say how you access the feature as an end-user. How a possible key binding should look like.  (T.i. how the main command is named.)  

Actually the API is very good to begin with, we just don't want to make any deprecation commitments at this early stage.  So we're tightly coupled because of that, and because at least in the beginning we need unit tests.  It is virtually impossible to write non trivial regexes without unit tests.  

On the other side, users can just import the decorator `@ignore_rules_inline` and use it even in User plugins for fun and testing.  The tricky part is what you write in such fixers, without the helper functions `extend_existing_comment` or `add_at_eol` it is challenging for non SublimeText freaks.  Maybe we need a lot of such helper functions, maybe not.  

So users can play with it, and just open PR's because maybe it is rather easy (like what you did with `phpcs`, anyone can write this). TL;DR **I don't want to investigate** it, which basically means I would have to write them anyway. Ideally we review good written PR's, *with* unit tests, and click merge. *With* tests so I don't have to test manually, we just merge.  The tests prove how the output will look. I don't want to review non-readable regexes, I see in the tests if they work. (And I would take that for the actual linters as well: copy paste the actual output of some linter, and prove that it gets transformed into our `LintError` correctly.)

braver: > Well, yeah, LOL, that's too short.

Actually, I don’t think so. People tend to ignore these messages, don’t even like to get them usually, and they cannot access the info after they close the message. It should just be a heads-up, with all the actual info in our readme, in the documentation. And how to use it should just be evident by using it. Nobody reads manuals anyway, and the update notification should definitely not be a manual.

I do get what you’re saying about PR’s though. I kinda doubt anyone will write tests though. Adding a quick action is trivial, writing tests isn’t, and isn’t something everyone knows how to do. But we’ll see what we’ll get. I’ll change the info to reflect that we want PR’s, not issues. Hopefully we can move the responsibility for these things to plug-in authors ASAP, so we don’t even have to be involved.

braver:
> We should say how you access the feature as an end-user

The screenshot does that quite excellently. And of course there is a command pallete entry called "Quick Action", we don't have to explain that. And of course there is the keymap and we have a generic explanation of how to add stuff to that already in the README as well. 

I have updated the README and update notification to specify that we want PR's, not issues. 

I didn't write any tests for the PHPCS thing, and honestly I don't know where to start on writing one, let alone running it (although maybe I can just let it run in CI). I'm not against it, I'd rather do these things driven by tests, but I don't know how to do it for our thing. Maybe the PHPCS action was trivial enough, but no regex ever is trivial. Anyway, maybe we need some more docs on that?

braver:
Shit, you still have any focus left? I'm worn to a thread and I don't even have a holiday to look forward to. If it looks like I don't have the energy for something, ie. that I'm even more lazy than is normal for me, it's probably because it's true.

Anyway, I'm really glad we wrapped this feature up and that it works as well as it does. I actually get a ton of utility from it. I'm really proud of it and I didn't even do anything final. Kudos to you man, great job (as always).

kaste:
focus is maybe a stretch.  No concerto, no theater, no movies, nor football.  Basically whatever you're into it's a no. It's some overwork but really also overboredom.

PR #1753, opened 2020-06-28
Fix/Ignore tweaks

braver:
Improve UI, consistency in messaging, and double check / 🚲-shed about everything user-facing.


kaste:
- improve quick actions via tooltip view

image

I actually used the icon as separator between the rulename and the message which is completely lost here. 

Vs. 

image

The icon looks too small here.  It's almost just a nicer "X".

- action singular: we will apply a single action

Ah well, we *show* multiple actions (otherwise we just apply the one action), and even then, I wrote wild code which actually combines multiple actions, we really may apply multiple actions.  That's bike shedding on the highest standard, though. 

- add context menu sub-entries

image

1. Is this intended that the "Quick Action" appears and is not even greyed although no action is available at the cursor?  
2. Should a context command not read the actual click position. T.i. we must implement and handle "want_event"?  
3. Should we not show a `show_popup_menu` here?

quick_actions.py
```diff
@@ -13,7 +13,7 @@
     LintError = persist.LintError
 
 
-class sl_quick_actions(sublime_plugin.TextCommand):
+class SublimeLinterQuickActions(sublime_plugin.TextCommand):
```
kaste, line 16: We use lowercased `_` names now because searchability. Looks like I did not convert all command names.

kaste: Of course, `sl_` was a mistake.

README.md
```diff
@@ -19,6 +19,11 @@ Settings are mostly documented in the [default settings](https://github.com/Subl
 - Additional information is in our docs at [sublimelinter.com](http://sublimelinter.com/).
 - Read about all the changes between 3 and 4 [here](https://raw.githubusercontent.com/SublimeLinter/SublimeLinter/master/messages/4.0.0.txt). 
 
+## Quick Actions (beta)
+
+We're building a new feature called Quick Actions, that will allow you to quickly ignore specific error messages. At first SublimeLinter will ship actions for [eslint](https://github.com/SublimeLinter/SublimeLinter-eslint), [stylelint](https://github.com/SublimeLinter/SublimeLinter-stylelint), [flake8](https://github.com/SublimeLinter/SublimeLinter-flake8) and [phpcs](https://github.com/SublimeLinter/SublimeLinter-phpcs).
```
kaste, line 24: And of course, `mypy`

menus/Context.sublime-menu
```diff
@@ -1,7 +1,23 @@
...
+                "caption": "Toggle Highlights",
+                "command": "sublime_linter_toggle_highlights"
+            },
+            {
+                "id": "sublimelinter-fix",
+                "caption": "Quick Action",
+                "command": "sublime_linter_quick_actions"
+            }
+        ]
     }
```
kaste, line 22: If you know, how users could remove or modify this, maybe add a comment.

quick_actions.py
braver: Oh, lowercased is also good, I always thought the use of snake case in one context and camel case in another over complicated things.

README.md
braver: 🙆‍♂️

braver:
This is why I want the icons in front: with longer error codes, and codes of different lengths, they otherwise end up all over the place:

Screen Shot 2020-06-28 at 20 48 08

I agree this is a better icon in general (this is the one we had before, but the other way around). I added an em-dash to separate the code from the message again.

braver:
Example quick panel in its current state, I'm pretty happy with this:

Screen Shot 2020-06-28 at 21 01 41

menus/Context.sublime-menu
braver: Removing context menu entries works via ST's standard override protocol. It's not easy, and not well known outside a handful old timers probably, but it's powerful.

- In Packages create a directory with the same name as the package you want to override (in this case "SublimeLinter", but this also works for AlignTab or ColorConvert to name a few).
- Using [PackageResourceViewer](https://packagecontrol.io/packages/PackageResourceViewer) open the Context.sublime-menu file from that package, and save it in the new directory you just created. 
- This file will now replace the one from the installed package. You can override every single file in any package this way.
- To remove the context menu, change the `children` key to an empty array `[]`.

To add entries, create a Context.sublime-menu file in the Packages/User directory (aka the "User package"), create an entry with the `id` that matches the one we ship, and create a child entry you want to add (e.g. `sublime_linter_panel_toggle`).

```json
[
    {
        "id": "sublimelinter",
        "caption": "Linter",
        "children": [
            {
                "id": "sublimelinter-panel",
                "caption": "Show Panel",
                "command": "sublime_linter_panel_toggle"
            }
        ]
    }
]
```

---

I'll add this to our docs.

braver:
> Is this intended that the "Quick Action" appears and is not even greyed although no action is available at the cursor?

Not really. Just a side effect that the `sublime_linter_quick_actions` has no `is_enabled` that handles this. It's a good idea to add this, also because it hide the command palette entry if there is nothing for that to act on. That's maybe more something you should add though, I don't know how to efficiently derive "there are no errors here".

> Should a context command not read the actual click position. T.i. we must implement and handle "want_event"?

That would seem to make more sense, but I'm not 100% sure it works that way, i.e. that it passes the click position at all. I believe you only get to trigger the command, and act on the selection, but that opening the context menu doesn't change the selection.

> Should we not show a show_popup_menu here?

Like this?

```json
{
    "id": "sublimelinter-fix",
    "caption": "Popup",
    "command": "sublime_linter_line_report"
}
```

... I think that might be overdoing it? But people can add anything they like. Which is a big reason to create the container entry with multiple context menu entries. 
https://github.com/SublimeLinter/SublimeLinter/blob/fix-ignore-tweaks/docs/customization.rst#context-menu

kaste:
I fix the test failure.  Looks like the order I return the actions can vary on each invocation, the quick panel entries are safe but inner generator? of actions yields randomly.

kaste:
Love the quick panel design with the "--" Thank you 😭 I hate designing quick panels. 

It says `(3x) -- Lorum` but `(2x) e.g. -- Lirum`. Shouldn't the latter be `(2x) -- e.g. Lirum`? You're in the really ugly code part here, so just asking if that was too complicated to do.

kaste:
That's basically

image

Vs.

image

braver:
> Shouldn't the latter be (2x) -- e.g. Lirum

Yes, that’s what I wanted to do but the code was too scary 😂

kaste:
> ... `show_popup_menu` ...

No, I literally meant instead of the quick_panel use the Sublime function `show_popup_menu` *iff* the user interacts with the context menu.  (`show_popup_menu` opens a custom context menu.)

kaste:
These look like

image

braver:
Oh that thing! Like the autocomplete menu? I never think of that as a mouse driven menu. I guess I’d have to try it. Doesn’t look any better than the quick panel though.

docs/customization.rst
```diff
@@ -0,0 +1,42 @@
...
+
+- In Packages create a directory with the same name as the package you want to override
+  (in this case "SublimeLinter", but this also works for AlignTab or ColorConvert to name a few).
+- Using `PackageResourceViewer <https://packagecontrol.io/packages/PackageResourceViewer>`_
+  open the Context.sublime-menu file from that package,
+  and save it in the new directory you just created.
+- This file will now replace the one from the installed package.
+  By the way, you can override every single file in any package this way.
+- To remove the context menu, change the ``children`` key to an empty array ``[]``.
+
```
kaste, line 20: I just installed PRV.  Looks like I can edit away the SublimeMerge ads.  But also looks like it's way easier (or is it only with the default packages). I just used "PRV: Open Resource", then navigate through the quick panel ~~hell~~ (it's super slick, actually) into `Default > Context` which opens the file with the correct path set, ready to edit and save. Away with the ads.

kaste: PRV does not recurse into sub directories.  We should move the two menu files out of the sub dir "menus" to the top level.  I never find them anyway?

kaste:
I won't dive into the context menu that much, I'm not used to it anyway.  Just wanted to show the option here.  That being said, I don't like the "quiet" option now.  I want to show the panel with "No actions" if we're using the Command Palette or context menu, but just flash the status message if used from a keyboard binding.  The latter works, the former also just flashes the status.  Obviously, the naming "quiet" was stupid to begin with.  

Generally, we don't have the time to compute available actions whenever the user opens the command palette, (even if we're fast that doesn't scale if the user has lots of packages, I think we should not even start this) so at least this entry has to be on all the time.  And if the eye is on the top in the command panel, you can't do nothing and just flash the status message.

kaste:
That being said, it could be also natural to *always* show the quick panel when interacting with the "Command Palette" even if there is only one action to select.  (You can still "ESC" here.)  For a shortcut, you sure want to just apply if possible.

kaste:
I actually also disagree with hiding/removing the entry for multiple selections.  Usually you interact with the "Command Palette" using acronyms, so after a while I just enter "qa<enter>".  Expecting the obvious.

Now I also programmed SublimeText that "qa" is "Quick Action" but I get

image

This is usually not what I want.  I entered "qa" because I *thought* there is an action available but my eyes were on the wrong line, I somehow confused where my selection actually is, or **the squiggles are just outdated**.

docs/customization.rst
braver: Excellent. IIRC you can browse into sub directories with PRV. Regardless, we should move the menus to the top level, I can never find them either.

braver:
Hm, that’s a good point. I tend to have the same problem, punching in an acronym and not getting what I expect because the command is not available for some reason. 

I think it’s a good idea to always have the quick panel. It’s simple enough and with one action you just hit return a second time and you’re done. If there are no actions the quick panel is also a good place to say so because your eyes are already there.

Before we cover too many subjects on one PR again, maybe merge this one and then tackle the quick panel interaction in a new PR?

braver:
Ok, I think I'm done here for now. Maybe you want to make a proposal for the quick panel interaction?

kaste:
Dude, you making it unnecessary complicated here.  You reverted the "quiet" arg without telling why. And now I make the proposal?  It doesn't open the "No actions avail." noop panel.

Also adding it to context menu is just not thought out.  For context menus you want `is_enabled` not `is_visible` to get it grey.  Then you need to implement `want_event` because otherwise we show actions for something the user actually not clicked on.  (Because we only evaluate the `sel` but the context opens on right click.)  Then you put everything in a sub menu, making it even more cumbersome to interact with, nobody asked for it, essentially for a context handler not even implemented correctly.

braver:
> You reverted the "quiet" arg without telling why.

I thought we covered that in the other super long PR, but I should've explained in the commit. We probably need to revise that again anyway though. Sorry if that made stuff harder for you!

> And now I make the proposal?

Well you seem to have a pretty good idea of how you want it to work and I think I get it and agree. I don't mind building it, I just don't want to go back and forth with you about it unnecessarily. Also I'm a lousy developer and I don't have a lot of time, so I'd risk making errors and not doing it correctly... so my thought here was that I should just get out of your way. If you want me to do it though, just say the word.

> Also adding it to context menu is just not thought out

I agree now that I tried it for a bit and gave it some time. I don't believe it's actually possible to show actions for *what you clicked on*, so we should just get rid of the context sub-menu and revert it to just "lint this view". I'll open a PR to do just that.

PR #1754, opened 2020-06-29
Quick action context menu - use clicked position

braver:
When triggering quick actions through the context menu, use the position you clicked on, not the selection. I was wrong, it *is* actually possible, just nobody uses this? (Or nobody uses context menus...).

I right-clicked the position at the arrow:
Screen Shot 2020-06-29 at 22 06 50

So, the selection is overruled if an `'event'` is passed, which contains x/y coords we can convert to a text point. 

Just as a proof of concept to get your thoughts here.

### DIFF
--- a/Context.sublime-menu
+++ b/Context.sublime-menu
@@ -1,24 +1,11 @@
 [
     {
-        "id": "sublimelinter",
-        "caption": "Linter",
-        "children": [
-            {
-                "id": "sublimelinter-lint",
-                "caption": "Lint This View",
-                "command": "sublime_linter_lint"
-            },
-            {
-                "id": "sublimelinter-highlights",
-                "caption": "Toggle Highlights",
-                "command": "sublime_linter_toggle_highlights"
-            },
-            {
-                "id": "sublimelinter-fix",
-                "caption": "Quick Action",
-                "command": "sublime_linter_quick_actions",
-                "args": { "prefer_panel": true }
-            }
-        ]
+        "caption": "Lint This View",
+        "command": "sublime_linter_lint"
+    },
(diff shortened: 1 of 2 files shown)


kaste:
Isn't that just great.  Really, 5 lines 100% better already.  Now I would check if point is in sel, and if it is use the selection instead.  Basically mimicking if you click on a selection, you act on the selection. In code, typing on mobile, if not sel.contains(point): sel = point given that at this point sel is set to the first selection.

kaste:
I think the context menu is very valid entry point.  Unfortunately, we hide it in a sub menu, and on my install "Linter" is the last entry in the context menu.  If I as a user decided to use my mouse for such things I would put it on the top of the menu.

I think about hiding the command if the view is not handled by SublimeLinter at all.  This could be a fast check, whereby checking if there are any actions available is rather slow.

braver:
Ok, try this. 

It now uses the selection you right-clicked on.

I moved the entries out of the sub-menu and to the top of the menu. I removed the "toggle highlights" entry because it's not what we're working on now and it was just a showcase of the benefit of the sub menu really. So now we have 1 entry, unless there are errors in which case we have 2. And we claim almost the top spot and that just makes it super easy to work with.

Is "Lint this view" actually something anyone ever uses? I only run linters manually when I changed a linter-config file. I really feel like making that one conditional too, for manual-only linters. Taking the top spot is cool, but it should be really useful.

kaste:
That's really awesome.

kaste:
That's a very unique implementation, and a "deep detail". 💘

braver:
🔥

PR #1755, opened 2020-06-30
Only show "Quick Action" in the context menu if there is any available

kaste:
~~Unbelievable ugly.~~

~~I looked at greying this "Quick Action" entry.  This gets really ugly code-wise.  Basically, we compute the actions in `is_enabled` or `is_visible` and pass it around.  But fun facts and fun bugs.  The order Sublime calls `enabled` or `visible` changes, on right clicks for the context menu `enabled` is called twice and *before* `visible`. (That's a double foul, as far as I can see, if `visible` returns `False` the entry does not appear, and asking `enabled` was a wasted computation because it only greys otherwise *visible* entries. Calling it even twice doesn't help here.)~~

~~Naturally you'd want to cache this computation, I'll invalidate the cache here "time-based" on the exact next tick.  Throughout SL we only cache by memoization (lru_caches), I would have to think and play with it if hashing the `view.sel()` plus `view.change_count` plus the uids in the error store for the file would be a unique cache key to proper invalidate.~~

~~Let's just look at if "greying" is that great.~~

Trivially, we only add an entry to the context menu if there are any actions available.  We never grey out.

### DIFF
--- a/quick_actions.py
+++ b/quick_actions.py
@@ -8,23 +8,23 @@ from .lint import util
 
 MYPY = False
 if MYPY:
-    from typing import Callable, List, TypedDict
+    from typing import Callable, List, Optional, TypedDict
 
     LintError = persist.LintError
+    QuickAction = quick_fix.QuickAction
     Event = TypedDict("Event", {"x": float, "y": float})
 
 
 class sublime_linter_quick_actions(sublime_plugin.TextCommand):
+    def want_event(self):
+        return True
+
     def is_visible(self, event=None, prefer_panel=False):
         # type: (Event, bool) -> bool
         if event:
-            filename = util.get_filename(self.view)
-            return len(persist.file_errors[filename]) > 0
+            return bool(self.available_actions(self.view, event))
         else:
             return len(self.view.sel()) == 1
-
-    def want_event(self):
-        return True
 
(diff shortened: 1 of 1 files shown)


braver:
Well, what is kind of inconsistent now is that Quick Actions is not there for some files, and there but greyed out in others. 

For main menu entries graying out is an important pattern. It ensures things are always in the same place, but can not necessarily always be used. 

For a context menu though, I already expect the menu to change based on what I right-clicked on (ie. the context). It's different all the time. It should only have things applicable to that context. Therefore, I don't think we should ever grey-out context menu entries. If it's not something you can use, don't show it.

kaste:
I ♥️ you.  That will make the code easier and the ux better.

braver:
Aww, I ❤️ you too man 🤣

kaste:
Here as in #1756 the code gets easy as soon as we don't "grey" anything.  Implementing both `is_enabled` and `is_visible` makes everything ugly because Sublime calls these callbacks in any order and even multiple times.

kaste:
Let's test this all together "in concerto".  Since we have it on "dev" we could even rebase "dev" if we see something so bad we feel a revert would be better than a fix.

PR #1756, opened 2020-06-30
Show "Lint this view" only if it makes sense

kaste:
Since we place our menu entry very much at the top of the context menu,
we want to show it only when it could be relevant for the user.

We assume that this is never the case when all linters for the view
actually run in background mode.  ~~We grey out the entry if some non-
background linters can't run at this point.  Usually only applicable
for linters which run against the real files on disk if the view is
currently dirty/not saved.~~ We never grey, we show an entry only 
if you can use it at this point in time.

### DIFF
--- a/goto_commands.py
+++ b/goto_commands.py
@@ -5,6 +5,7 @@ from Default import history_list
 from itertools import dropwhile, takewhile
 
 from .lint import persist, util
+from .lint.util import flash
 
 
 MYPY = False
@@ -91,9 +92,3 @@ def move_to(view, point):
     # type: (sublime.View, int) -> None
     history_list.get_jump_history_for_view(view).push_selection(view)
     view.run_command('_sublime_linter_move_cursor', {'point': point})
-
-
-def flash(view, msg):
-    # type: (sublime.View, str) -> None
-    window = view.window() or sublime.active_window()
-    window.status_message(msg)
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -220,6 +220,14 @@ class ViewSettings:
             self.view.id(), self.prefix.rstrip('.'))
 
 
+NOT_EXPANDABLE_SETTINGS = {
+    "lint_mode",
+    "selector",
+    "disable",
+    "filter_errors",
+}
+
+
 class LinterSettings:
     """
     Smallest possible dict-like container for linter settings to lazy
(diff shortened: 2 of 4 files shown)


kaste:
Not having this entry is nice esp. when it is placed at the top of the menu.  

The code complication because of the optimizations is a *must* here.  On my lousy computer we go from 60ms to open the context menu, down to 10ms in the worst case. T.i. we're now faster than `master` which had only naive functionality.

braver:
> Not having this entry is nice

Yes I agree. 

> We assume that this is never the case when all linters for the view actually run in background mode

Actually, I think the command is only useful if you use the "manual" mode. Linters that run on-save, e.g. on real files only, will run on-save, and triggering them to run on a (possibly dirty) buffer is weird. 

So, for a given view we need a list of manual-mode linters, and a check against the global `lint_mode` setting. That should be pretty performant. Accessing the settings that are computed for a view is cheap. 

Also, like the Quick Actions command I believe we should not bother with `is_enabled` and only implement `is_visible` for the context menu.

braver:
I just had a thought. I sometimes need to trigger a manual lint, because I changed a config file. So, if we’re doing is_visible, we probably need to make a proxy command for the context menu. The main command should always be available.

kaste:
"Not background" is the mode we want here.  I personally don't have any file-only linters but can switch to save mode if it gets too noisy.  For example, you switch the lint mode for mypy to save only, then you can write code and from time to time manually lint, you get an automatic lint on every save.  That's totally okay.  (I have a shortcut to switch the lint_mode.  I just never used the context menu.  ... But my context menu was cluttered and not very useful.  Since I removed the SM ads, and some other stuff with the tip you gave, I actually have a very nice and short ctx menu where you can actually find your stuff.)

Sure the command is always "enabled" and ready to use via key or palette.  I removed the greying already before initial push here, so the commit message was already outdated. We either have a usable entry to click on or none.

braver:
Ok, good points. I think it's worth the effort here, but we do need to reach a point where you're happy with the amount of code it requires. I'm also ok with leaving it dumb, I can always remove it from the context menu if I never use it (it's still available via the command palette anyway).

kaste:
If we're okay with the feature here in the PR, I can remove all the complexity.  At this point I only left it to make tweaking the UX easier.  (If we don't implement both `is_enabled` and `is_visible` we don't need caching.)

braver:
> If we're okay with the feature here in the PR

I'm totally happy with the UX here as it is. If I switch a linter to be in manual mode, it's there, and otherwise it's not and doesn't get in the way. Which is excellent.

kaste:
That's a trivial PR now.  10 lines if I had not added the basic status message flashing in `run`.

PR #1757, opened 2020-06-30
Do not offer to ignore flake8 codes which are hard errors

kaste:
Remove most of the flake8 toy real fixers.  But keep an easter egg.


braver:
👍

PR #1723, opened 2020-04-03
Remove unused `debug` and `debug_mode` functions

kaste:
Fixes #1714

We generally use `logging` in SL4.

All open source linter plugins (listed by Package Control) have been
converted at this time, so we hard deprecate without further notice.

PR #1758, opened 2020-07-13
Two removals

kaste:
- Remove `util.printf` probably forgotten in #1723

- Hard deprecate so called project settings; de-facto deprecated since 4.6.6, finally deprecated in 4.13.0


braver:
Excellent 👌

PR #1767, opened 2020-08-21
Allow for customization of Quick Fixes

TomasBarry:
This commit implements the capacity to configure the Quick Fixes
made available by SublimeLinter. This commit deals solely with the
ESLint Quick Fixes but lays the foundation for extending the
customization to the other linters that can apply the Quick Fix feature.

Taking ESLint as our example, it is personal preference whether an
ESLint violation is ignored with an `// eslint-disable-next-line
VIOLATION(S)` or an `// eslint-disable-line VIOLATION(S)` comment.
Currently, SublimeLinter only provides the functionality to use the
`// eslint-disable-next-line` comment. It would be preferable to be able
to configure this to use the type of inline Quick Fix that the
programmer prefers.

TODO - Actually detail the final SublimeLinter settings

@kaste (or another maintainer) - I am looking for a bit of advice as to what you think of this feature and whether I should pursue it from here and if I should, how you feel the schema settings should look. This is a bit of an "in theory" PR, which is why I've made it a draft, as I am looking for maintainer feedback, particularly on the schema settings, and how best to access those settings.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -110,6 +110,14 @@
                 }
             ],
 
+            // Specify what form of quick fix to implement if there are
+            // multiple options
+            "quick_fix_settings": {
+                // An ESLint can either have an eslint-disable-next-line or an
+                // eslint-disable-line comment
+                "eslint": "previous_line"
+            },
+
             // The current working dir the lint job will run in.
             "working_dir": "",
 
--- a/lint/quick_fix.py
+++ b/lint/quick_fix.py
@@ -5,6 +5,8 @@ import re
 
 import sublime
 
+import Settings
+
 from . import persist
 from .generic_text_command import replace_view_content, text_command
 flatten = chain.from_iterable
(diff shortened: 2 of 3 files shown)


TomasBarry:
Agree with you both.

What is the flexibility here in enabling specific extension of SublimeLinter to implement their own quick fixes functionality? Some linters have a `--fix` command that could possibly be used. Just commenting an ignore isn't really a fix.

kaste:
Well, *how* to construct good fixes is actually not part of the API here.  That's to say, we don't call a "--fix" command here, that would be a different feature.  To what I know, there is no linter that would fix a specific error per CLI request anyway.  A fix here is a `TextRange` or "TextEdit" object which can be applied to a view.  (Interestingly, I have now seen that the Microsoft language servers emit basically the same object for some Code Actions, so we're in a basic way compatible with LSP.) 

I think "eslint" actually emits fixes in its JSON output.  If so, these maybe can be read, stored and then made available within the provided UI.

kaste:
Well, 🎧, that PR and interaction just doesn't move in a reasonable speed.

PR #1793, opened 2021-05-17
Ensure to skip saved-file-only-linter for unnamed files

kaste:
`file_on_disk` linters should *not* run on newly created (empty) files.  We always just asked `is_dirty()` which is true for all unnamed, *non-empty* views so this PR resolves an obvious but edge case.  Surprising it went unnoticed for so long.


kaste:
Great, travis doesn't even run. 🚬

PR #1797, opened 2021-05-19
Add id to menu entry

### LINKED ISSUE #1796: allow overrides for main menu
Currently `SublimeLinter` doesn't allow overrides in the main menu. If an id were added to `SublimeLinter` in `Main.sublime-menu` i could add a submenu for my linter settings.

arcturus140:
this PR closes #1796

### DIFF
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -11,6 +11,7 @@
                 "children": [
                     {
                         "caption": "SublimeLinter",
+                        "id": "sublimelinter",
                         "children": [
                             {
                                 "caption": "Settings",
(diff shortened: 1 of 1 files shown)


kaste:
One side effect of adding an id is that Sublime will put these entries at the end of the list, for example at the end of the sub-menu.  I don't think that's a problem here as the order of the entries seems random anyway.  I would choose the `id` "sublimelinter-settings" as that's what LSP does.  

@braver?

arcturus140:
fyi i decided to use `sublime-commands` instead. The size of the menu under `Package - Settings` is getting out of hand. 

i can still provide changes as suggested if that is of use for somebody else.

PR #1799, opened 2021-05-21
Let LintMatch recognize end_line and end_col

### LINKED ISSUE #1584: `LintMatch` should support `end_line` and `end_col`
We have 'line' and 'col', and since we have the underscore naming philosophy here, we should introduce 'end_line' and 'end_col'. This is probably not for regex based linters, at least I don't know any linter yielding such info on the command line, but for JSON reporters. 

Note: If these keys are given, `reposition_match` will not be called at all since it is the sole purpose of that method to compute these.

kaste:
Fixes #1584 

Add support for `end_line` and `end_col` in `LintMatch`.  This is usually only useful for JSON linters, at least I don't know any compact line-based stdout format which outputs this information.  

This is basically extracted from the eslint plugin.  It works with the companion PR over there https://github.com/SublimeLinter/SublimeLinter-eslint/pull/309

Of course, generalizing it here, handling the edge cases and do the proper clamping/error handling, logging doubles the code from eslint. 

---

By looking at the code, we capture `end` in `LintError` which is unused as we nowadays use `region` throughout the app.  But computing it makes this PR a bit more unreadable.  A future PR should remove it.  (We still use `line` and `start` basically as a cached value to present in the error panel; so we don't have to call `view.rowcol` per render.)  

--- 

The last commit has a refactoring which is strictly speaking a "major" update in that we don't allow "holes" in the underlying `LintMatch` dict.  A hole is for example, a missing key `col` etc.  We now fill in these "holes" and set the value `None`.  This should be of no particular interest as the user probably already used the dotted notation `match.col` which already exchanged a missing key with `None` in such cases. The old tuple unpacking we see in old plugins did this as well.

Usually we don't want the doubled optionality we technically had: first, maybe the key is present, then maybe the value is None.  The dotted, object notation already opted in for the optionality on the value, not the key, so we finally follow.  

No tests enforced the old behavior as well.

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -171,6 +171,13 @@ class VirtualView:
         """Return the (full) line region including any trailing newline char."""
         return sublime.Region(*self.full_line(line))
 
+    def line_region(self, line):
+        # type: (int) -> sublime.Region
+        """Return the line region without the possible trailing newline char."""
+        r = self.full_line_region(line)
+        t = self.substr(r).rstrip('\n')
+        return sublime.Region(r.a, r.a + len(t))
+
     def select_line(self, line):
         # type: (int) -> str
         """Return code for the given line."""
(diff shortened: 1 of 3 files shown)


kaste:
This can be tested by installing the eslint PR; the exact behavior can be read out of the tests.  (How we clamp wrong values, what are the defaults for missing values.)

This *is* mostly test-driven developed.  

Examples:

- Given just a `line` and a `end_line`, select both lines completely.
- Given just a `line`, a `col`, and a `end_col`, select the range on `line`, t.i. assume `end_line==line`
- etc.

PR #1801, opened 2021-05-21
Log detailed information about the queued linting tasks

kaste:
If for some buffer we put "excessively many" tasks on the queue,
currently more than 4, log detailed information about which linter
uses how many tasks.  E.g.

```
SublimeLinter: backend.py:88          WARNING: 'index.html' puts in
total 6(!) tasks on the queue:  1x htmlhint, 1x htmltidy, 1x eslint,
1x annotations, 2x flow.
```

Compared to before:

```
SublimeLinter: backend.py:88          WARNING: 'index.html' puts in
total 6(!) tasks on the queue.
```


braver:
I don't have anything at hand test this, but agreed on the concept 👌🏻

PR #1803, opened 2021-05-21
Fix matching absolute Windows file paths in the panel

kaste:
Obviously, absolute Windows paths can contain a `:` for which our syntax pattern failed.  See added test.

### DIFF
--- a/panel/panel.sublime-syntax
+++ b/panel/panel.sublime-syntax
@@ -47,8 +47,9 @@ contexts:
 
   expect-filename:
     - include: pop-at-end
-    - match: ([^:]+)
-      scope: entity.name.filename.sublime_linter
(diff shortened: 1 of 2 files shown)


panel/panel.sublime-syntax
```diff
@@ -47,8 +47,9 @@ contexts:
 
   expect-filename:
     - include: pop-at-end
-    - match: ([^:]+)
-      scope: entity.name.filename.sublime_linter
+    - match: (.+):(?=\n)
```
braver, line 50: I'd lean towards keeping it a bit more specific, e.g.

```
((\S:)?[^:]+):
```

kaste: Done

PR #1805, opened 2021-05-23
Fix: Draw dangle and icon in EOF position

### LINKED ISSUE #1784: Does not highlight errors at EOF position
Hello,

It seems that this error is not displayed by SublimeLinter-php (there are 2 opening curly brackets) if there are empty lines below the closing } , while it would be shown in the console:

```php
<?php

if (!$query->execute()) {{
	$a = 1;
}


```
WechatIMG1412

If you remove the empty lines, it would be then properly displayed in Sublime text.

WechatIMG1413

kaste:
Fixes #1784

We always ensured regions of size 1 for all errors, because it looks
better to not have dangles all over the code.  We further recognize
empty regions which can happen when the user edits on errors, or deletes
or moves blocks of line, and remove these automatically.  That means
an empty region has a special meaning in SublimeLinter.

Unfortunately, Sublime will not draw squiggles for regions that go past
the EOF at all as #1784 has shown.

Let the linter produce empty regions for that exact edge case on EOF,
t.i. exactly the region `Region(view.size())`.  However, if only parts
of a file are linted, the virtual view inside the linter can "think" it
has an error on eof when it is actually on the end of the linted *part*
of the file only.  Check in the backend, and maybe undo what the linter
did.

Draw empty regions as dangles, t.i. with the flag
`DRAW_EMPTY_AS_OVERWRITE`, and ensure they're excluded from our dangles/
zombie protection.

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -622,12 +622,16 @@ def maybe_update_error_store(view):
         # XXX: Why keep the error in the store when we don't have
         # a region key for it in the store?
         key = uid_key_map.get(uid, None)
-        region = head(view.get_regions(key)) if key else None
+        if key is None:
+            new_errors.append(error)
+            continue
+
+        region = head(view.get_regions(key))
         if region is None or region == error['region']:
             new_errors.append(error)
             continue
-        else:
-            changed = True
+
+        changed = True
 
         if region.empty():
             # Either the user edited away our region (and the error)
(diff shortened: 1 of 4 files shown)


kaste:
E.g. 

image

This can be tested with a fake linter.  For example, create `User/fake_linter.py` with the following content:

```
from SublimeLinter.lint import Linter


class fake_php(Linter):
    cmd = None
    defaults = {
        'selector': 'source.php, text.html.basic'
    }
    regex = r"^(?P<filename>.*):(?P<line>\d+):(?P<col>\d+) (?P<message>.*)"

    def run(self, cmd, code):
        return "stdin:{}:1 The message".format(self.view.rowcol(self.view.size())[0] + 1)
```

PR #1815, opened 2021-06-05
Showcase: Use ST4 annotations aside highlighted regions

kaste:
For example

image

But usually it doesn't look right because you always run out of space. 

image

The implementation is raw and not to be merged.


braver:
The problem I have with these kinds of things (similar for “git lens”, but also some quick actions LSP provides) is indeed that I hardly ever have sufficient space on the side for anything. Well maybe when doing some python. I’d rather glance at the status bar for more info. 🤷🏼‍♂️ Cool though!

kaste:
New implementation:

Introduce a new style key `annotation` (by default falsy) which can be used to enable and format side annotations. E.g.

image

image


This way annotations can be enabled per linter, per error_type, or "code"/rule name.  Since the message is often too wordy and also not standardized the setting accepts placeholders like `{msg}` or `{code}`.   I actually run this patch for months.  Personally I only have it enabled for `codespell` and some other metrics.

braver:
Oh, that's smart. I can totally see how `{code}` would be useful there!

kaste:
The new champion is #1875

PR #1823, opened 2021-07-30
Paypal pools are ending soon

braver:
And I don't have a replacement for it. Any ideas?


kaste:
Github sponsors/funding is probably the future, but mildly complex to setup.  🤷

PR #1828, opened 2021-10-10
Add shellcheck ignore action

Narretz:
Hi,
I stumbled upon the experimental QuickAction API that was introduced last year.
I really like the idea, because it can really improve SublimeLinter's usefulness. While for some languages, LSP is now the better choice, some linters are so simple that Quick Actions are easier implemented without a whole language server.

So as an experiment, I added shellcheck ignore + go to docs quick action to https://github.com/SublimeLinter/SublimeLinter-shellcheck

I would like to see this as an official API, although I guess the project has gone a bit quiet. Here are some suggestions:

- as far as I can see, only the ignore actions can be embedded into the popup as quick fix buttons. But I think it would also be useful to embed "go to documentation" and possibly other actions
- the embedded quick fixes need the option to set a name for the action. At the moment, it's just "⌦" which doesn't really tell you what will happen
- if additional quick actions are available, the popup should have a button at the bottom (next to "Copy") that opens the quick action panel
- the current implementation seems to assume that all quick actions use TextCommand. It's possible to not use it of course, but if this API was going public, quick actions should not be required to use a TextCommand

### DIFF
--- a/lint/quick_fix.py
+++ b/lint/quick_fix.py
@@ -454,34 +454,6 @@ def fix_shellcheck_error(error, view):
         )
     )
 
-
-@quick_actions_for("shellcheck")
-def shellcheck_goto_documentation(errors, view):
-    # type: (List[LintError], Optional[sublime.View]) -> Iterator[QuickAction]
-    if view:
-        region = view.sel()[0]
-
-        for error in errors:
-            match = re.search(SHELLCHECK_CODE_PATTERN, error["msg"])
-
-            code = match.groups("code")[0];
-
-            yield QuickAction(
-                "shellcheck: Open documentation — {}".format(error['msg']),
-                # partial(eslint_ignore_block, errors, region),
-                partial(shellcheck_open_documentation_in_browser, code, region),
-                "",
-                solves=[]
-            )
-
-def shellcheck_open_documentation_in_browser(errorCode, region, view):
-    # type: (List[LintError], sublime.Region, sublime.View) -> Iterator[TextRange]
-    webbrowser.open_new_tab("https://github.com/koalaman/shellcheck/wiki/{}".format(errorCode))
-
-    # return generator without any text commands
-    return
-    yield
-
 def line_from_point(view, pt):
     # type: (sublime.View, int) -> TextRange
     line_region = view.line(pt)
--- a/lint/quick_fix.py
+++ b/lint/quick_fix.py
@@ -431,8 +431,6 @@ def fix_mypy_error(error, view):
         )
     )
 
-SHELLCHECK_CODE_PATTERN=r"\[(?P<code>SC\d+)\]$"
-
 @ignore_rules_inline("shellcheck")
 def fix_shellcheck_error(error, view):
     # type: (LintError, sublime.View) -> Iterator[TextRange]
--- a/lint/quick_fix.py
+++ b/lint/quick_fix.py
@@ -1,7 +1,6 @@
 from collections import defaultdict
 from functools import partial
 from itertools import chain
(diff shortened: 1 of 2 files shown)


kaste:
Sorry for the delay, I had my sabbatical.  (And am now faced with this PR against the rather non-trivial, dense and abstract quick fix code I wrote.)  As I read it, a quick fix is a function  `(error, view) -> TextRange`; such a `TextRange` can be seen as an edit instruction which can be applied to a view or a buffer.  T.i. the fixer is in some sense pure as it does not do the side effect of fixing but returns a description of it.

Showing documentation is in contrast a function `(error) -> url_string` if again we just want the description of the action, not the actual performance of `webbrowser.open`.  T.i. the documentation is rather static and does not depend on a `view` or actual buffer.  Most likely it is a static mapping from error code or rule name to an URL string.   

I have a feeling that the different signatures already indicate that the UI for both should not be mixed as it is a different thing.  We had a PR https://github.com/SublimeLinter/SublimeLinter/pull/1480 btw to show documentation links we could revive; iirc we did not merge this PR because for most linters there was no documentation available. 

I would recommend to split this PR.  The safe part is to add the quick inline fixer for shellcheck.  Please add tests in https://github.com/SublimeLinter/SublimeLinter/blob/master/tests/test_ignore_fixers.py as appropriate.

Narretz:
Thanks for the comments.

I can see that it makes more sense to keep the API clean, but documentation definitely has its place. I don't know if some linters have the docs in their library or maybe statically in the filesystem, but if you add an API for docs it should be flexible enough to handle this. Or if not, then a simple config option in the base linter that constructs a url would be enough.

I have removed the documentation command. I'll add the tests next week. Looks like I can adapt the stylelint checks.

But I wonder what the | (pipe) does for example in this line: https://github.com/SublimeLinter/SublimeLinter/blob/7c2f2922067bd794e9c094e03316402bfd981c0c/tests/test_ignore_fixers.py#L320

kaste:
The `|` simply denotes where the cursor is when running the command. 

We like the idea of having documentation available; it was basically a chicken-egg-problem.  #1480 might be still the way to go as it just expects a new `url` key on the `error` dictionary.  The presentation in the error popup looked good, for example to underline the rule name and make it clickable.  To have it *also* in the error menu ("quick panel" in Sublime Text speech) would be just the next step, a different presentation.

Narretz:
Ok, thanks.

How do I run the tests, though? I didn't find any info about this in the repository at all.

kaste:
Shit documentation we have atm. 🙄  The CI is actually fast.  But you can just install `UnitTesting` which is the testing framework for Sublime Text.  Then `Test Current Package` from the `ctrl+shift+p` menu (or: `Test current file` while looking at the test file) to run the tests.

Narretz:
I couldn't get the tests to run locally (no output), so I've pushed a speculative commit. It seems like you need to approve the runs individually, though.

kaste:
Yeah okay, it's red 🤪

Narretz:
ok, I had some small errors in the tests. Should work better now

kaste:
You're still red.  I think I found the right setting so that the tests don't need approval anymore.  🤞

Narretz:
Very nice. Just in time for me to fix all errors. :D

In hindsight, I should have taken a closer look at the code

kaste:
Excellent.  Unfortunate that we don't grab the `code` in the shellcheck plugin (so that you have to parse the `msg` here).  Maybe you can change the regex over there, and then we follow here.  

Other than that, I think I merge and release tomorrow.  ❤️

PR #1875, opened 2022-09-09
Even better than the real thing

kaste:
Annotations aren't good enough for the purpose.  We don't have enough white space on the right hand too show the actual message.  Multiple errors on the same line don't work without mouse interaction.  

So we try the real phantoms.  Usually these are annoying and special care must be taken so that the viewport doesnÄt shift around all the time.  From the annotations, I learned that it must be a per linter thing.  Then we can turn it off for e.g. the "annotations" plugin, we can turn it off for some whitespace rules, **and** we absolutely **must** turn it off for the syntax errors.  Because while writing code we all the time produce syntax errors.   Repeat after me: mute syntax errors.  

image

Here we even demote all warnings and infos etc.  As long as the view has *errors* we *only* show errors.  It would be useful if the plugins could ship with some defaults.  Because there is no general rule name for "syntax error" so we can't mute these magically.  

The phantoms completely fulfill the purpose of annotating the view so that the user knows what a squiggle means.  There is a 100% relationship between the squiggle and the explainer message.   The squiggle toggle becomes necessary to have at hand, for example 

```
   { "keys": ["ctrl+k", "ctrl+k"],
     "command": "sublime_linter_toggle_highlights"
    },
```

(as suggested in our help text) because copy, pasting code can lead to enormous, temporary errors, for example because nothing is imported yet and all variables have the wrong name.  Here `ctrl+k ctrl+k`is a must.  

E.g. 

- a syntax error

image

- a whitespace **warning**

image

- a short right hand side annotation

image


kaste:
Example settings:

image

kaste:
The last commit here c5cddc8 (Revert "Draw annotations of hidden squig..) controls if the right hand annotations toggle with the `ctrl+k ctrl+k` command.  Without the revert, only the sqiggles with the phantoms toggle, the icons on the left hand and the annotations on the right hand of the view stay.

PR #1884, opened 2022-12-15
(Minor) copyedits

jareha:
A couple minor copyedits for `SublimeLinter.sublime-settings`.

### DIFF
--- a/SublimeLinter.sublime-settings
+++ b/SublimeLinter.sublime-settings
@@ -3,7 +3,7 @@
     "debug": false,
 
     // When in the "background" lint mode, this value determines
-    // the minimum delay before a request is send to the linter
+    // the minimum delay before a request is sent to the linter
     "delay": 0.1,
 
     // Available gutter themes:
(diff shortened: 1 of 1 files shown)


SublimeLinter.sublime-settings
```diff
@@ -55,11 +55,11 @@
...
     "lint_mode": "background",
 
-    // Linter specific settings.
+    // Linter-specific settings.
     // More info: http://www.sublimelinter.com/en/stable/linter_settings.html
-    // Linter specific settings except for 'styles' can also be changed
+    // Linter-specific settings except for 'styles' can also be changed
     // in sublime-project settings.
-    // What settings are available is documented in the readme of the
+    // What settings are available are documented in the readme of the
```
kaste, line 62: really?  Isn't that here the short form of "It is documented ...".  Not "other possible settings are documented".

jareha: The “it” in this case is plural (“linter-specific settings”).

SublimeLinter.sublime-settings
```diff
@@ -3,7 +3,7 @@
     "debug": false,
 
     // When in the "background" lint mode, this value determines
-    // the minimum delay before a request is send to the linter
+    // the minimum delay before a request is sent to linter
```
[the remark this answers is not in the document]
jareha: Went ahead and backed that change out, as could go either way as “the” isn't always necessary.

SublimeLinter.sublime-settings
kaste: Ok, I'm not a native speaker.  I thought "Additional settings _are_ documented in the readme." but "Which settings are valid here _is_ documented in the readme." because it is not the subject.  I can transform the sentence to "What settings are allowed here *I* wrote down over there." or typically flip "*It* is documented in the readme what/which settings are further available".   LOL, but what do I know. :grin:

PR #1892, opened 2023-02-05
Serialize squiggle side-effects on the UI thread

### LINKED ISSUE #1891: SublimeLinter hardlocks Sublime Text with deadlock on rapid line content deletes with annotations
This was originally submitted as sublimehq/sublime_text#5846

### Description of the bug

When using Neo​Vintageous (I did not test Vintageous), doing rapid ex commands with deletes on lines with annotations hard locks Sublime Text. I have been able to do this crash across multiple computers in safe mode

Please take a look at my reproduction steps for details. There is probably an even more minimal reproduction, but this is as far as I got since I didn't have a quick way of adding a lot of annotations otherwise.

Turning off annotation prevents the lockup for me; using ctrl+shift+k to delete lines does not lock up, and moving lines does not lock up. Doing a chain of ex commands with some deletes and annotation is currently my most faithful recreation.

This happens across multiple languages and files.

### Steps to reproduce

1. Start ST in safe mode
2. Install Package Control
3. Install `Neo​Vintageous`, `Sublimelinter`, `Sublime​Linter-contrib-yamllint`
4. Add the enclosed following `SublimeLinter.sublime-settings`
6. Open the enclosed `alacritty.yal` if yamllint is working almost every line should have a warning annotation. See screenshot enclosed
7. quickly type `<esc>jDjD`
8. Enjoy your pinwheel of doom

[Lockup replication files.zip](https://github.com/sublimehq/sublime_text/files/10514062/Lockup.replication.files.zip)

### Expected behavior

Sublime Text deletes to the end of the line and does not lock up

### Actual behavior

Sublime Text locks up and never recovers, requiring a force quit. ST is hanging waiting for the plugin to respond. The SublimeLinter is hanging [here](https://github.com/SublimeLinter/SublimeLinter/blob/0b540fa1aae0b925d1ab6c2856dcb1506b7f5883/highlight_view.py#L622). Looks like a deadlock.

### Sublime Text build number

4143

### Operating system & version

macos 13.1

### Additional information

This lockup happens both on my 1.4 GHz Quad-Core Intel Core i5 MacBook and my M1 Max MacStudio

### OpenGL context information

```plain
Intel
OpenGL Context Information:
  GL API Version: 4.1 INTEL-20.2.48
  GLSL Version: 4.10
  Vendor: Intel Inc.
  Renderer: Intel(R) Iris(TM) Plus Graphics 645

M1 Max
OpenGL Context Information:
  GL API Version: 4.1 Metal - 83
  GLSL Version: 4.10
  Vendor: Apple
  Renderer: Apple M1 Max
```

Stack trace from plugin host here
https://github.com/sublimehq/sublime_text/issues/5846#issuecomment-1409673359

kaste:
Fixes #1891
Fixes sublimehq/sublime_text#5846

The idea of the code was to serialize the side-effects of drawing squiggles while on different threads.  The code wanted to protect the view effects (`[add|erase]_regions`) and our in-memory store of these regions (`CURRENTSTORE`).  Using a `threading.Lock` looked easy and convenient.

This didn't work out because the UI thread itself holds a lock on all views so that e.g. a `view.add_regions` waits for the UI thread to be available again.

Then we have the following scenario:

1. On the worker, we acquire our `StorageLock` because of e.g. new linter results.
2. The user edits and thus `on_modified` fires on the UI thread to `revalidate_regions`.  This implicitly locks the view.  We then try to acquire the `StorageLock`.  The intention was to just wait for (1) to complete, speed did not matter here.
3. On the worker we call `view.add_regions` which never returns because it awaits the (implicit) lock of the view.

In consequence we deadlock: (3) waits for (2), but (2) waits for (1).

The fix:

Remove the `Lock` and serialize all side-effects on the UI thread using tasks.  Use decorators to mark the essential functions.

- `@assert_on_ui_thread` marks the inner functions that used the `Lock` before.  It raises `RuntimeError`s if the functions are wrongly used on a different thread.

- `@ensure_on_ui_thread` lifts functions onto the UI thread if necessary.

Lifting the inner functions (the "edges") (e.g. `draw_view_region`) is too costly as we don't want to create possible hundred tasks on the UI for hundred squiggles to draw.  We also want them to be bundled, t.i. drawn at once.  Therefore the dev has to decide which upper function should be marked `ensure_on_ui_thread`.

I use a decorator here to lift the whole function because not lifting is an error.  T.i. I don't expect a legal call to e.g. `draw` from anything else than the UI thread.

### DIFF
--- a/highlight_view.py
+++ b/highlight_view.py
@@ -5,7 +5,6 @@ from itertools import chain
 from functools import partial
 import re
 import textwrap
-import threading
 import uuid
 
 import sublime
@@ -164,12 +163,8 @@ def highlight_linter_errors(views, filename, linter_name):
         squiggle_regions = ChainMap(
             {}, highlight_regions, hidden_highlight_regions  # type: ignore[arg-type]
         )  # type: Squiggles
-        draw(
-            view,
-            linter_name,
-            squiggle_regions,
-            gutter_regions,
-        )
+
+        draw(view, linter_name, squiggle_regions, gutter_regions)
         draw_phantoms(view)
 
 
@@ -185,6 +180,7 @@ def draw_phantoms(view):
     update_phantoms(view, phantoms)
 
 
+@util.ensure_on_ui_thread
 def update_phantoms(view, phantoms):
     with stable_viewport(view, phantoms):
         get_phantom_set(view).update(phantoms)
@@ -437,6 +433,7 @@ def undraw(view):
         erase_view_region(view, key)
 
 
+@util.ensure_on_ui_thread
 def draw(
     view,               # type: sublime.View
     linter_name,        # type: LinterName
@@ -585,7 +582,6 @@ class DemotePredicates:
 # --------------- ZOMBIE PROTECTION ---------------- #
 #    [¬º°]¬ [¬º°]¬  [¬º˚]¬  [¬º˙]* ─ ─ ─ ─ ─ ─ ─╦╤︻ #
 
-StorageLock = threading.Lock()
 # Just trying and catching `NameError` reuses the previous value or
 # "version" of this variable when hot-reloading
 try:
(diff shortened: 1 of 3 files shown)


kaste:
We need 🙏 and 🤞 here as I can't code-review myself, and it is a sophisticated bug after all.

PR #1899, opened 2023-04-25
Detect all Yarn projects, PnP or not

mataha:
In its current form, SublimeLinter avoids calling Node's binary when invoking a linter installed with npm, preferring to run executables while relying on the implementation detail of the `node_modules` linker, i.e. `node_modules/.bin`. However, a Yarn PnP project (*de facto* any project not relying on [Node Resolution Algorithm](https://nodejs.org/api/modules.html#modules_all_together)) doesn't play well with this strategy due to using a different install mode to manage its dependencies. The algorithm used to determine whether a project used the PnP linker (#1649) relied on, among other things, existence of Yarn-specific files including a `.pnp.js` mapper. Recent versions of Yarn (since 3+) don't generate files with that extension anymore, having moved towards `.pnp.cjs` in tandem with `.pnp.loader.mjs` should the project require PnP API compatible with ESM. That part could be fixed by globbing `.pnp.*js` in about two lines of code. Continuing with this approach, however, has shortcomings:

- [a project may use a different install strategy e.g. pnpm](https://yarnpkg.com/advanced/rulebook#user-scripts-shouldnt-hardcode-the-node_modulesbin-folder)
- [`yarn.lock` might not be generated](https://yarnpkg.com/getting-started/qa#should-lockfiles-be-committed-to-the-repository)
- [the layout of `node_modules` might not always be the same](https://yarnpkg.com/advanced/rulebook#modules-shouldnt-hardcode-node_modules-paths-to-access-other-modules)

This PR aims to provide a better solution by not relying on implementation modules of various package resolution methods, hoisting etc. instead always calling Yarn regardless of linker used, effectively deferring executable lookup to the package manager.

The proposed algorithm relies on the following markers in order to determine whether the project is a Yarn-based project:
- [`packageManager`](https://nodejs.org/api/packages.html#packagemanager) key in `package.json` containing `yarn`
- [`.yarn.lock`](https://yarnpkg.com/getting-started/qa#should-lockfiles-be-committed-to-the-repository)
- [`.yarnrc.yml`](https://yarnpkg.com/configuration/yarnrc) (Yarn >= 2)
- [`.yarnrc`](https://classic.yarnpkg.com/en/docs/yarnrc) (Yarn 1.x)
- `node_modules/.yarn-integrity` (veeeeery legacy, last resort)

See also: #1644

### DIFF
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -214,12 +214,11 @@ class NodeLinter(linter.Linter):
 
                     self.logger.warning(
                         "Skipping '{}' for now which is listed as a {} "
-                        "in {} but not installed.  Forgot to '{} install'?"
+                        "in {} but not installed.  Forgot to 'npm install'?"
                         .format(
                             npm_name,
                             'dependency' if is_dep else 'devDependency',
-                            manifest_file,
-                            'npm'
+                            manifest_file
                         )
                     )
                     self.notify_failure()
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -47,6 +47,20 @@ def smart_paths_upwards(start_dir):
     )
 
 
+def is_yarn_project(path, manifest):
+    # type: (str, Dict[str, Any]) -> bool
+    package_manager = manifest.get('packageManager')
+    if isinstance(package_manager, str):
+        name = package_manager.split('@')[0]
+        return name == 'yarn'
+
+    yarn_files = ['yarn.lock', '.yarnrc.yml', '.yarnrc']
+    if [file for file in yarn_files if os.path.exists(os.path.join(path, file))]:
+        return True
+
+    return os.path.exists(os.path.join(path, 'node_modules', '.yarn-integrity'))
+
+
 # `read_json_file` is maybe used by plugins. Check `eslint` and
 # `xo` for example.
 def read_json_file(path):
@@ -173,18 +187,15 @@ class NodeLinter(linter.Linter):
                 if is_dep or is_dev_dep:
                     self.context['project_root'] = path
 
-                    # Perhaps this is a Yarn PnP project?
-                    yarn_lock_exists = os.path.exists(os.path.join(path, 'yarn.lock'))
-                    if yarn_lock_exists and (
-                        os.path.exists(os.path.join(path, '.pnp.js'))
-                        or bool(manifest.get('installConfig', {}).get('pnp'))
-                    ):
+                    # Perhaps this is a Yarn project?
+                    if is_yarn_project(path, manifest):
+                        # https://yarnpkg.com/advanced/rulebook#user-scripts-shouldnt-hardcode-the-node_modulesbin-folder
                         yarn_binary = shutil.which('yarn')
                         if yarn_binary:
                             return [yarn_binary, 'run', '--silent', npm_name]
 
                         self.logger.warning(
-                            "This seems like a Yarn PnP project. However, finding "
+                            "This seems like a Yarn project. However, finding "
                             "a Yarn executable failed. Make sure to install Yarn first."
                         )
                         self.notify_failure()
@@ -199,7 +210,6 @@ class NodeLinter(linter.Linter):
                         if executable:
                             return executable
 
-                    package_lock_exists = os.path.exists(os.path.join(path, 'package-lock.json'))
                     self.logger.warning(
                         "Skipping '{}' for now which is listed as a {} "
                         "in {} but not installed.  Forgot to '{} install'?"
@@ -207,7 +217,7 @@ class NodeLinter(linter.Linter):
                             npm_name,
                             'dependency' if is_dep else 'devDependency',
                             manifest_file,
-                            'yarn' if yarn_lock_exists and not package_lock_exists else 'npm'
+                            'npm'
                         )
                     )
                     self.notify_failure()
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -57,7 +57,7 @@ def is_yarn_project(path, manifest):
         return name == 'yarn' or name == '@yarnpkg/berry'
 
     yarn_files = ['yarn.lock', '.yarnrc.yml', '.yarnrc']
-    if [file for file in yarn_files if os.path.exists(os.path.join(path, file))]:
+    if any(os.path.exists(os.path.join(path, file)) for file in yarn_files):
         return True
 
     return os.path.exists(os.path.join(path, 'node_modules', '.yarn-integrity'))
--- a/lint/base_linter/node_linter.py
+++ b/lint/base_linter/node_linter.py
@@ -51,8 +51,10 @@ def is_yarn_project(path, manifest):
     # type: (str, Dict[str, Any]) -> bool
     package_manager = manifest.get('packageManager')
     if isinstance(package_manager, str):
-        name = package_manager.split('@')[0]
-        return name == 'yarn'
+        # When this field was being adopted and hadn't been standardised yet
+        # package managers could have scoped names, i.e. `@<scope>/<package>`
+        name = package_manager.rsplit('@', 1)[0]
+        return name == 'yarn' or name == '@yarnpkg/berry'
 
     yarn_files = ['yarn.lock', '.yarnrc.yml', '.yarnrc']
     if [file for file in yarn_files if os.path.exists(os.path.join(path, file))]:
--- a/tests/test_node_linter.py
+++ b/tests/test_node_linter.py
@@ -177,27 +177,17 @@ class TestNodeLinters(DeferrableTestCase):
         verify(linter).notify_failure()
 
     @p.expand([
-        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', False, False, 'npm'),
-        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', False, True, 'npm'),
-        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', False, False, 'npm'),
-        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', False, True, 'npm'),
-        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', True, False, 'yarn'),
-        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', True, True, 'npm'),
-        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', True, False, 'yarn'),
-        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', True, True, 'npm'),
+        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', 'npm'),
+        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', 'npm'),
     ])
     def test_uninstalled_local_dependency(
         self,
         ROOT_DIR,
         CONTENT,
         DEPENDENCY_TYPE,
-        HAS_YARN_LOCK,
-        HAS_PACKAGE_LOCK,
         EXPECTED_PACKAGE_MANAGER,
     ):
         PRESENT_PACKAGE_FILE = os.path.join(ROOT_DIR, 'package.json')
-        PACKAGE_LOCK_FILE = os.path.join(ROOT_DIR, 'package-lock.json')
-        YARN_LOCK_FILE = os.path.join(ROOT_DIR, 'yarn.lock')
 
         when(self.view).file_name().thenReturn('/p/a/f.js')
         linter = make_fake_linter(self.view)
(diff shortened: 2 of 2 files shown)


kaste:
To summarize what we currently have:  `yarn.lock` is mandatory, and then either a `.pnp.js` file or an entry in _package.json_  (under the key: `installConfig`) and we will assume a yarn pnp project.  

> - [`yarn.lock` might not be generated](https://yarnpkg.com/getting-started/qa#should-lockfiles-be-committed-to-the-repository)

You say "might not be generated", but you link to "should it be committed to the git repo?" which is a completely different thing.  Is there another reference telling the _yarn.lock_ file is not generated anymore?

I think the yarn configuration files (e.g. `.yarnrc.yml_) are poor markers as they can be placed outside of the project.  That's what I read at least. (Are they even merged if multiple files lay around, so users might have some configuration in their home directory?) 

You state:

> Continuing with this approach, however, has shortcomings:
>
> - [a project may use a different install strategy e.g. pnpm](https://yarnpkg.com/advanced/rulebook#user-scripts-shouldnt-hardcode-the-node_modulesbin-folder)

We don't do this, do we?

> - [`yarn.lock` might not be generated](https://yarnpkg.com/getting-started/qa#should-lockfiles-be-committed-to-the-repository)

See above. 

> - [the layout of `node_modules` might not always be the same](https://yarnpkg.com/advanced/rulebook#modules-shouldnt-hardcode-node_modules-paths-to-access-other-modules)

We don't do this as well.  We just call out to `yarn run`.  

You propose we should watch out for a `packageManager` key in _package.json_ but this is experimental:

image

(E.g. _Corepack_ must be used to enable it.)

Overall, I'm confused here about this PR and about yarn in general as they now promote "Zero-Installs" modes.  To look out for _.pnp.cjs_ is the only thing really clearly documented.

mataha:
Firstly, I'd like to preface my comment by mentioning the driving force behind this PR: frustration. SublimeLinter hasn't been working correctly with my Yarn projects for more than a year, which caused me to disable it entirely and rely on CLI tools instead.

> So when you're deleting old tests, I think "really?"; was that broken? would the your new approach would have worked before too?

Rather than 'broken', those paths weren't necessary anymore with the new approach.

> To summarize what we currently have: `yarn.lock` is mandatory, and then either a `.pnp.js` file or an entry in _package.json_ (under the key: `installConfig`) and we will assume a yarn pnp project.

Correct.

> > * [`yarn.lock` might not be generated](https://yarnpkg.com/getting-started/qa#should-lockfiles-be-committed-to-the-repository)
> 
> You say "might not be generated", but you link to "should it be committed to the git repo?" which is a completely different thing. Is there another reference telling the _yarn.lock_ file is not generated anymore?

My apologies, I linked one thing while thinking about the other - while some library authors insist that lockfiles do not belong in repositories, I meant that it is possible for a lockfile to exist, though under a different name (specified under the [`lockfileFilename`](https://yarnpkg.com/advanced/rulebook#user-scripts-shouldnt-hardcode-the-node_modulesbin-folder) key in `.yarnrc.yml` or via an environment variable).

> I think the yarn configuration files (e.g. `.yarnrc.yml`) are poor markers as they can be placed outside of the project. That's what I read at least. (Are they even merged if multiple files lay around, so users might have some configuration in their home directory?)

That file is integral to the project (as it contains internal settings e.g. linker used and the path to a local binary) and is automatically generated for new ones (`.yarnrc.yml` for Yarn >= 2, `.yarnrc` otherwise) in the root of the repository. To be fair there is an option for checking for PnP here, but would require parsing YAML.

That said, there can be a global `.yarnrc.yml` (in that case the settings are replaced, not merged).

> You state:
> 
> > Continuing with this approach, however, has shortcomings:
> > 
> > * [a project may use a different install strategy e.g. pnpm](https://yarnpkg.com/advanced/rulebook#user-scripts-shouldnt-hardcode-the-node_modulesbin-folder)
> 
> We don't do this, do we?

There are code paths that look for an executable in `node_modules/.bin` later on.

> > * [`yarn.lock` might not be generated](https://yarnpkg.com/getting-started/qa#should-lockfiles-be-committed-to-the-repository)
> 
> See above.

See my earlier point on lockfiles with different names. Fresh projects check this mark as well.

> > * [the layout of `node_modules` might not always be the same](https://yarnpkg.com/advanced/rulebook#modules-shouldnt-hardcode-node_modules-paths-to-access-other-modules)
> 
> We don't do this as well. We just call out to `yarn run`.

If the project is PnP. What if there's a global cache or an offline mirror set?

> You propose we should watch out for a `packageManager` key in _package.json_ but this is experimental:
> 
> image
> 
> (E.g. _Corepack_ must be used to enable it.)

Having this key in `package.json` is good practice and is not relevant to using or even turning on Corepack. In fact, you don't need Corepack at all if you're running Yarn through a wrapper (this PR doesn't touch that though). And it's not like the new approach won't work when that key is not present - we will just fall back to looking for marker files.

> Overall, I'm confused here about this PR and about yarn in general as they now promote "Zero-Installs" modes. To look out for _.pnp.cjs_ is the only thing really clearly documented.

Zero-Installs is unrelated to this PR - it's a workflow, not an installation strategy. It doesn't have to use PnP.

To summarize - calling Yarn straight-up when a project exhibits signs of being Yarn-based should cover a lot more use-cases compared to the previous approach (only-if-PnP) and be less error-prone.

kaste:
> To summarize - calling Yarn straight-up when a project exhibits signs of being Yarn-based should cover a lot more use-cases compared to the previous approach (only-if-PnP) and be less error-prone.

The current (pre-your-PR) implementation reads like that we *must* detect the PnP strategy as otherwise `yarn run the_thing` would not work.  Can yarn 1.0 or 2.0 just `run` anything?  That's basically my concern about backwards compatibility.

mataha:
[Yes](https://classic.yarnpkg.com/en/docs/cli/run#toc-yarn-run-script) and [yes](https://yarnpkg.com/cli/run#details).

lint/base_linter/node_linter.py
```diff
@@ -47,6 +47,22 @@ def smart_paths_upwards(start_dir):
...
+    # type: (str, Dict[str, Any]) -> bool
+    package_manager = manifest.get('packageManager')
+    if isinstance(package_manager, str):
+        if package_manager.startswith('@'):
+            package_manager = package_manager[1:]
+        name = package_manager.split('@', 1)[0]
+        return name == 'yarn'
+
+    yarn_files = ['yarn.lock', '.yarnrc.yml', '.yarnrc']
+    if [file for file in yarn_files if os.path.exists(os.path.join(path, file))]:
```
kaste, line 60: Various ways to make this efficiently stop on the first match. Typically `if any(os.path.exists(...) for file in yarnfiles)`.

lint/base_linter/node_linter.py
```diff
@@ -47,6 +47,22 @@ def smart_paths_upwards(start_dir):
     )
 
 
+def is_yarn_project(path, manifest):
+    # type: (str, Dict[str, Any]) -> bool
+    package_manager = manifest.get('packageManager')
+    if isinstance(package_manager, str):
+        if package_manager.startswith('@'):
+            package_manager = package_manager[1:]
+        name = package_manager.split('@', 1)[0]
```
kaste, line 56: So the expected string is `@yarn@3.5` or `@yarn` etc.?  Make sure to add a test for it.

lint/base_linter/node_linter.py
```diff
@@ -199,15 +212,14 @@ def find_local_executable(self, start_dir, npm_name):
                         if executable:
                             return executable
 
-                    package_lock_exists = os.path.exists(os.path.join(path, 'package-lock.json'))
                     self.logger.warning(
                         "Skipping '{}' for now which is listed as a {} "
                         "in {} but not installed.  Forgot to '{} install'?"
```
kaste, line 217:
```suggestion
                        "in {} but not installed.  Forgot to 'npm install'?"
```

lint/base_linter/node_linter.py
```diff
@@ -199,15 +212,14 @@ def find_local_executable(self, start_dir, npm_name):
...
                         .format(
                             npm_name,
                             'dependency' if is_dep else 'devDependency',
                             manifest_file,
-                            'yarn' if yarn_lock_exists and not package_lock_exists else 'npm'
+                            'npm'
```
kaste, line 222:
```suggestion

```

kaste: Looks like this part is static now.

tests/test_node_linter.py
```diff
@@ -177,27 +177,17 @@ def test_uninstalled_local_dependency_cant_read_package_json(self, ROOT_DIR):
...
-        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', False, False, 'npm'),
-        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', False, True, 'npm'),
-        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', False, False, 'npm'),
-        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', False, True, 'npm'),
-        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', True, False, 'yarn'),
-        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', True, True, 'npm'),
-        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', True, False, 'yarn'),
-        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', True, True, 'npm'),
+        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency', 'npm'),
+        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency', 'npm'),
```
kaste, line 181:
```suggestion
        ('/p', {'dependencies': {'mylinter': '0.2'}}, 'dependency'),
        ('/p/a', {'devDependencies': {'mylinter': '0.2'}}, 'devDependency'),
```

tests/test_node_linter.py
```diff
@@ -177,27 +177,17 @@ def test_uninstalled_local_dependency_cant_read_package_json(self, ROOT_DIR):
...
     ])
     def test_uninstalled_local_dependency(
         self,
         ROOT_DIR,
         CONTENT,
         DEPENDENCY_TYPE,
-        HAS_YARN_LOCK,
-        HAS_PACKAGE_LOCK,
         EXPECTED_PACKAGE_MANAGER,
```
kaste, line 188:
```suggestion

```

tests/test_node_linter.py
kaste: Again, looks like this is static now.

tests/test_node_linter.py
kaste: same

lint/base_linter/node_linter.py
```diff
@@ -47,6 +47,22 @@ def smart_paths_upwards(start_dir):
     )
 
 
+def is_yarn_project(path, manifest):
+    # type: (str, Dict[str, Any]) -> bool
+    package_manager = manifest.get('packageManager')
+    if isinstance(package_manager, str):
```
kaste, line 53: Is the intention to check for `None`s or can we have numbers and booleans here?

kaste (changes requested):
This looks like a fabulous PR.

lint/base_linter/node_linter.py
mataha: `None`s.

kaste: Okay.  Typically `if package_manager:` or `if pm is not None` but since this is user input your defensive approach makes totally sense.

lint/base_linter/node_linter.py
mataha: Done.

lint/base_linter/node_linter.py
mataha: Fixed.

lint/base_linter/node_linter.py
mataha: Fixed.

tests/test_node_linter.py
mataha: Fixed.

lint/base_linter/node_linter.py
mataha: Now uses `any` with a generator expression.

lint/base_linter/node_linter.py
```diff
@@ -47,6 +47,23 @@ def smart_paths_upwards(start_dir):
...
+        # When this field was being adopted and hadn't been standardised yet
+        # package managers could have scoped names, i.e. `@<scope>/<package>`
+        tokens = package_manager.split('@')
+        name = '@' + tokens[1] if package_manager.startswith('@') else tokens[0]
```
kaste, line 57: That looks rather complicated but I also don't exactly know if we're expecting any variations here.  E.g. this 

```suggestion
        name = package_manager.rsplit('@', 1)[0]
```

should be the same.  

But the really static `return package_manager.startswith("yarn") or package_manager.startswith("@yarnpkg/berry")` makes the tests green and naively readable.

But that's a nit.

kaste (approved):
That's *really* a fantastic PR here, I will pull later.

lint/base_linter/node_linter.py
mataha: In this case it is the same, thanks!

braver:
🚀

PR #1907, opened 2023-06-01
Fix type errors in `ruby_linter.py`

kaste:
After that we're mypy-green.


kaste:
LOL, I just merged without waiting for the checks.  Double-click.

PR #1917, opened 2023-08-14
Add `VirtualView.rowcol` method to calculate (row, col) from offset

harrymander:


kaste:
Very well done!  👏

kaste:
I just released this as 4.23.0

PR #1949, opened 2024-05-19
Prepare 4.25.0 release

kaste:
We add `typing_extensions` as a dependency to download it now and have the users prepared for when we actually depend on it.  (We already use it in *master*. )


rgpublic:
Hmm. I get this when I try to install SublimeLinter on a new machine:

```
Package Control: The dependency 'typing_extensions' is not currently installed; installing...
Package Control: The dependency 'typing_extensions' is not available
Package Control: The dependency 'typing_extensions' could not be installed or updated

```

It seems to work on an existing machine. What could be the reason for this?

kaste:
Same Sublime Text version?

rgpublic:
Yes, DEV 4178. We've solved it now by coping the package over from a working machine. It works without any problems but the new installation via Package Control fails.

kaste:
Strange.  I'll look into this later.  Which OS, platform are you using?

rgpublic:
Linux / Manjaro

kaste:
I only can reproduce this with compiled extensions where Package Control might not have the an exact version for the specific platform.  E.g. a python extension that is not available for Apple Silikon (yet) but only for x86.  But `typing_extensions` is not compiled but a python only extension and just ships (https://pypi.org/project/typing-extensions/#files) one file *typing_extensions-4.12.2-py3-none-any.whl* for *any* platform although I don't know what the *none* portion in the file name stands for.  

Maybe enabling the verbose mode in Package Control would have been the smarter advice.  Maybe it would have printed a detailed complain.

kaste:
FWIW.  I had the same issue now.  On a fresh Sublime install, after just installing Package Control.  At that point Package Control was outdated already, although just installed, and couldn't fetch the new dependencies.  I had to restart Sublime to get Package Control working/updating, and then to install SublimeLinter again.

flyskywhy:
And also need

    sudo cp /usr/lib/python3/dist-packages/typing_extensions.py /opt/sublime_text/Lib/python38/

kaste:
@flyskywhy That should not be necessary as we declare that in https://github.com/SublimeLinter/SublimeLinter/blob/master/dependencies.json

There must be a reason why the automatic install fails.

PR #1957, opened 2024-08-28
Add new setting enable_cells=False

kaste:
SublimeLinter always linted parts of a view, t.i. it did not *only*
operate on whole files.  E.g. it could lint code blocks inside markdown
documents.  There was no way to opt-out of this behavior, except maybe
to use the `excludes` setting to exclude all `*.md` files.

Let's call these sub-blocks "cells", as in: notebook cells.  Normal
linters do not understand a cell structure of a document just out of
their box, and in that *normal* case linting cells is practically
useless.

Let's introduce a new setting `enable_cells` which defaults to `False`
(which makes it a breaking change).  Plugin authors can opt-in in their
plugin, so can users on a case-by-case basis.

E.g. you could still let `flake8` run on python blocks in your docs if
you want to but then probably configure per-file overrides, e.g. to only
report whitespace errors or to exclude any unused-var warnings, etc.

### DIFF
--- a/docs/linter_settings.rst
+++ b/docs/linter_settings.rst
@@ -46,6 +46,21 @@ For both, Python and Node, SublimeLinter has sophisticated ways to find *locally
 
 When the `disable_if_not_dependency` setting is set to `true`, SublimeLinter will not attempt to use globally installed binaries if a local installation cannot be found. Instead, it will skip linting such projects altogether.
 
+enable_cells
+------------
+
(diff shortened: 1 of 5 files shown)


braver:
That's smart. It hardly ever worked properly anyway (ie. I always have issues when eslint tries to lint a `<script>` block, at the very least indentation is off across the entire thing). Good work man 👍🏻

PR #1972, opened 2025-03-26
Humanize column in the panel

kaste:
We literally used `rowcol(error.region.a)` to compute the line and start used in the panel.  This is ok, but if the user uses real tabs, the column we show does not match the visual column the user sees in the editor.

Actually, Sublime Text shows the humanized/visual column in the status bar, and that mismatch makes it even stranger.  Once noticed you cannot unsee it.

Fix it by reading the "tab_size" and adjusting the `rowcol`.

Note that "tab_size" belongs to the "context" of the lint request, just like all the other properties of the view.

Rel SublimeLinter/SublimeLinter-phpcs#57


kaste:
@braver  👀

braver (approved):
Yeah, that works for me 👍🏻 The change is pretty elegant too!

braver:
@kaste I think there is a problem with this. If I use the keyboard shortcuts / commands to jump to the previous or next result, the cursor lands in the wrong position.

braver:
Although… hmm, not consistently. Ugh this entire business breaks my brain 😅

kaste:
Ah dang, likely if you use the built-in next_result command, the one driven by the output regex.  I never use them, they are bound to f4 and confusing, because they're like a shared resource: all output panels and the find result panel populate the regex, and the user never knows what's currently in this "places" memory.  Every view just overwrites this location list.   And if the panel isn't open the list isn't even updated.  
On the other hand, they can jump to different files too.  
What are the feature we actually need here?  Should we also make a Sublime Text upstream feature/bug report.  Is it a bug that Sublime Text doesn't handle this situation?

braver:
I think that’s indeed exactly the thing. I’ll need more than a few minutes to wrap my head around and whether to put blame with ST or something else.

kaste:
Maybe we should not populate the location list.  The double click can be supported without it since ST gained contextual mousemaps.  We can still support "next_result"  by deferring to our own "goto"-command.  Is the jumping to other files useful?  Should this work when the panel is closed?  There is probably a good reason vim has multiple of these location lists, and linter results don't mess up grep results iirc.   If we lint here, we do that in an ongoing-process, in the background, so there is no real task or "run"-result that imo justifies overwriting the location list all the time.  We also shuffle the output panel results while navigating between files.  (That's why going to the next file using next_result feels so random I believe).  

I think the "next_result" feature is after all: a) one key-binding to rule them all, b) jumps between files.

But i) it gets confusing because it is a shared resource, ii) the lint results are not ordered like ... e.g. compiler errors.  Hence jumping between files is not reasonable but somewhat random.
(vim has multiple lists but then of course mnemo-shortcuts to switch between them; it is really not just two bindings, next and previous result.)

Our "goto" could learn to jump to a next file.  Likely that would be just the next view, t.i. in tab-order.  

Clearly, overwriting the location list always, even if the panel is closed, would be insane.  So at max we can update the list if the panel is open.  But then the location list is still loaded if we close the panel, although the actual linter errors come and go "on_modified".  Which makes the jumping ugly because we jump to outdated positions all the time.  

> Maybe we should not populate the location list.

braver:
> Is the jumping to other files useful? Should this work when the panel is closed?

I use that all the time. Indeed it's a shared resource, but I like having the same keybinding to go through any set of results, whether search results, lint results, build results, etc. So that's why we have an output panel that complies with what ST does. 

I also confirmed that pre humanization, next_result jumps to the correct spot, and with humanization it does not. Even though the column numbers seem wrong:

Screenshot 2025-04-08 at 12 14 28

Of course search results are by line only, not column:

Screenshot 2025-04-08 at 12 16 06

I don't have a build system at hand that reports columns, to check what the output looks like there. I don't immediately see an open issue for that on ST... I'll check in on Discord, see what the people have to say about it.

braver:
Oh, of course LSP has a diagnostics panel. And they also don't humanize and have the "wrong" column (this is via intelephense, not phpcs), but it does work as expected:

Screenshot 2025-04-08 at 12 26 16

braver:
Well regardless of what we think is correct, this PR needs to be reverted to restore functionality and make our panel behave like all other panels again. The panel clearly has an implicit expectation that tabs are not converted to spaces in the column position, since that's what all tools do and otherwise all plugins and build systems would need to do the conversion we're doing here (and especially in build systems that's not even possible in most cases).

kaste:
Of course jumping works before the change here.  😕  We just trading two bugs.  *But*  I can also fix it

koen-at-peppered:
> Note by the way that if you use "goto anything" in my example, you should put in 97:3 and not 97:13.

I'm starting to think that the thing to fix is the status bar column indicator more than anything else.

kaste:
Oh, it comes with a nice picture.  Matches the humanize topic here.

braver:
Yeah, I guess I just blew my cover 😂 🕵🏻

PR #1990, opened 2025-10-30
Fix save events cancelling modified events in "on_modified" mode

### LINKED ISSUE #346: Not clearing most recent fixed error
Often when I fix an error it does not disappear from the gutter until I trigger a new error.

### LINKED ISSUE #1985: Linting randomly breaks with 4.28.0
Plugin gets stuck highlighting non-errors in code, need to close/re-open file as a workaround.

kaste:
Fixes #1985
Fixes https://github.com/SublimeLinter/SublimeLinter-eslint/issues/346

### DIFF
--- a/lint/linter.py
+++ b/lint/linter.py
@@ -1115,7 +1115,11 @@ class Linter(metaclass=LinterMeta):
             return True
 
         effective_lint_mode = get_effective_lint_mode(settings)
-        if real_file_linter and "on_modified" in effective_lint_mode:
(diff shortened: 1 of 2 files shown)


kaste:
This is likely correct.  If we want to de-dupe "on_modified, on_save" successions where the view's `change_count` did not change we need to make that not so deep down in the code, but likely either where we `hit` or debounced where we prepare the `lint` (both in `backend.py`).

